PR #12, opened 2018-04-23
Add hook to create readme

gabel0287:


terraform_docs.sh
```diff
@@ -0,0 +1,28 @@
...
+  fi
+
+  let "index+=1"
+done
+
+for path_uniq in $(echo "${paths[*]}" | tr ' ' '\n' | sort -u); do
+  path_uniq="${path_uniq//__REPLACED__SPACE__/ }"
+
+  pushd "$path_uniq" > /dev/null
+  terraform-docs md ./ > README.md
```
antonbabenko: This will overwrite the whole `README.md` which is not what everyone wants.

I think we should make some things configurable via [pre-commit arguments](https://pre-commit.com/#passing-arguments-to-hooks):
1. Filename to create (by default, `README.md`)
2. Beginning and end markers where terraform-docs content should be placed between (by default, invisible blocks like `<!-- BEGINNING OF TERRAFORM-DOCS HOOK -->` and `<!-- END OF TERRAFORM-DOCS HOOK -->`). If markers were not found - replace the whole file.

What do you think?

gabel0287: Great idea!  I'll update the PR.

PR #27, opened 2018-12-14
Add new hook for running terraform-docs with main.tf doc replace

### LINKED ISSUE #17: hook fails if comment tags not found
I would expect a tool like this to operate in a very standard mode, not trying to do anything like inserting between comments. This is your design decision to make but I'd much prefer a more standard execution of the tool like:

```
terraform-docs md ./ | cat -s | ghead -n -1 > README.md
```

This approach is markdown linter compliant and supports the standard of putting your documentation in the `main.tf`

screen shot 2018-05-17 at 12 48 57 am

RothAndrew:
Adds a new hook called `terraform_docs_replace` which uses the documentation found at the top of `main.tf` to recreate README.md.

Fix #17


setup.py
```diff
@@ -0,0 +1,34 @@
...
+from setuptools import setup
+
+
+setup(
+    name='pre-commit-terraform',
+    description='Pre-commit hooks for Terraform',
+    url='https://github.com/antonbabenko/pre-commit-terraform',
+    version_format='{tag}+{gitsha}',
+
+    author='Andrew Roth',
```
RothAndrew: I put my name since I wrote the python package, but I'm fine with it changing to yours if desired.

antonbabenko: It is fine to have your name as a package owner, just change `description` to `Pre-commit hooks for terraform_docs_replace`.

.pre-commit-hooks.yaml
```diff
@@ -23,6 +23,14 @@
   files: (\.tf)$
   exclude: \.terraform\/.*$
 
+- id: terraform_docs_replace
+  name: Generate documentation for Terraform modules
```
antonbabenko: `name: Terraform docs (overwrite README.md)`

.pre-commit-hooks.yaml
```diff
@@ -23,6 +23,14 @@
...
+  language: python
+  entry: terraform_docs_replace
+  files: (\.tf)$
+  exclude: \.terraform\/.*$
+  description: Generates README.md files for Terraform modules
```
antonbabenko: `description: Overwrite content of README.md with terraform-docs`

setup.py
RothAndrew: done

RothAndrew:
> Make filename README.md customizable via args.

Sure that sounds reasonable.

> Where main.tf comes from? Where it is used? Why main.tf?

main.tf is the standard entrypoint for a Terraform module. `terraform-docs` looks for documentation at the top of main.tf to add to the markdown document it creates. I'm not explicitly telling it to do that.

> Move setup.py and package files inside a folder (eg, terraform_docs_replace) to not mix root of the repo.

According to [the docs](https://pre-commit.com/#python) setup.py has to stay at the root in order to get picked up by pip.

> sed command can be different between systems. Since this is already in Python I would recommend to use native Python methods and replace sed.

Makes sense

.pre-commit-hooks.yaml
RothAndrew: done

.pre-commit-hooks.yaml
RothAndrew: done

antonbabenko:
Looks good. :)

Could you also add a note to README for developers (not users) something like "It is possible to support any and all future pre-commit hooks written in Python. All you have to do is add a line to the console_scripts array in entry_points."

RothAndrew:
Don't merge this yet. I'm not done with it.

RothAndrew:
Actually, maybe I am. @antonbabenko if you're fine with leaving the call to `sed` in for now this can be merged. I'll open another PR to change the `sed` call to native Python.

README.md
```diff
@@ -62,13 +62,13 @@ Check the [source file](https://github.com/antonbabenko/pre-commit-terraform/blo
 
 1. `terraform_docs` and `terraform_docs_without_aggregate_type_defaults` will insert/update documentation generated by [terraform-docs](https://github.com/segmentio/terraform-docs) between markers - `<!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->` and `<!-- END OF PRE-COMMIT-TERRAFORM DOCS HOOK -->` if they are present in `README.md`. Make sure that `terraform-docs` is installed.
 
-1. `terraform_docs_replace` replaces the entire README.md rather than doing string replacement between markers. Put your additional documentation at the top of your `main.tf` for it to be pulled in.
+1. `terraform_docs_replace` replaces the entire README.md rather than doing string replacement between markers. Put your additional documentation at the top of your `main.tf` for it to be pulled in. The optional `--dest` argument lets your change the name of the file that gets created/modified
```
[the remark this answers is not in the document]
RothAndrew: :+1:

RothAndrew:
Ready to merge. I'm done making changes unless there are more requested.

robinbowes:
> main.tf is the standard entrypoint for a Terraform module. terraform-docs looks for documentation at the top of main.tf to add to the markdown document it creates. I'm not explicitly telling it to do that.

This is not clear - could you update the docs to explain this, @RothAndrew ?

PR #34, opened 2019-02-21
Add exit code for 'terraform validate' so pre-commit check fails

josiahhalme:
Exit code for failed 'terraform validate -check-variables=true', so it fails pre-commit check correctly.
[https://github.com/antonbabenko/pre-commit-terraform/issues/22](https://github.com/antonbabenko/pre-commit-terraform/issues/22)

PR #35, opened 2019-02-28
fix check for errors at the end on validate w/ variables

supertylerc:
`-n` checks to see if a variable is defined.  The way the script is currently written, the `terraform_validate_with_variables` check will always fail because the last thing it does is check to see if `error` is defined and does an `exit 1` if so.  This change makes it so that we check to see `error` is not `0` instead.

### DIFF
--- a/terraform_validate_with_variables.sh
+++ b/terraform_validate_with_variables.sh
@@ -27,6 +27,6 @@ for path_uniq in $(echo "${paths[*]}" | tr ' ' '\n' | sort -u); do
   popd > /dev/null
 done
 
-if [[ -n "${error}" ]] ; then
+if [[ "${error}" -ne 0 ]] ; then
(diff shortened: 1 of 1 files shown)


supertylerc:
This looks like it was inadvertently introduced in #34, which was to fix the exact opposite problem: a passing run that should've failed.

An alternative to this might be to completely remove the line that initializes `error = 0` and revert to checking to see if `error` is defined.

antonbabenko:
I agree, your explanation and this PR makes sense.

@josiahhalme what do you think? Please give it a try and let us know if there is any problem with this.

antonbabenko:
[v1.11.0](https://github.com/antonbabenko/pre-commit-terraform/releases/v1.11.0) has been released.

@josiahhalme Please try this one if you can.

PR #38, opened 2019-03-18
Use more idiomatic bash code

robinbowes:
I've tweaked the `terraform_validate_with_variables.sh` script to be a little more robust and use more idiomatic bash code which should handle files with spaces in their name more elegantly.

A similar approach could be applied to `terraform_validate_no_variables.sh`. In fact, both scripts could probably be merged into one, with a command-line option to choose whether to test with or without variables.

### DIFF
--- a/terraform_validate.sh
+++ b/terraform_validate.sh
@@ -8,9 +8,17 @@ ERROR=0
 
 # main code entrypoint
 main() {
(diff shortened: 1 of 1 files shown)


robinbowes:
Conflict fixed, plus a few enhancements.

PR #42, opened 2019-05-31
Allow env overide of terraform command

dpetzold:
Hi,

I have terraform v0.12 installed as terraform2. This PR allows me override the terraform command the hook uses. Everything should work the same as before.

Derrick


antonbabenko:
I don't think that this is needed.

Use [tfenv](https://github.com/tfutils/tfenv) or similar tools to switch Terraform version.

PR #55, opened 2019-07-16
Support mktemp on macosx

rgevaert:


antonbabenko:
AFAIK, due to the difference in implementations of `mktemp` on different platforms, it is better to not use `-t` and `--tmpdir`. This has been changed already - https://github.com/antonbabenko/pre-commit-terraform/commit/10854fcfa2202bd677446b2d298a0ea6119941fb#diff-de43b10a1d85aff5249c2beb34d032e8R33

Am I missing something? Is the proposed change required for this hook to work on your machine? If so, what is the error?

alexs77:
mktemp <template> DOES work on macOS.

```bash
⋊> ~/t/S/misc on master ⨯ which mktemp
/usr/bin/mktemp
⋊> ~/t/S/misc on master ⨯ ls -la /usr/bin/mktemp
-rwxr-xr-x  1 root  wheel  18800  4 Mai 09:03 /usr/bin/mktemp
⋊> ~/t/S/misc on master ⨯ strings /usr/bin/mktemp
mktemp
dqt:u
TMPDIR
/tmp/
%s%s.XXXXXXXX
%s/%s.XXXXXXXX
cannot generate template
mkdtemp failed on %s
mkstemp failed on %s
usage: mktemp [-d] [-q] [-t prefix] [-u] template ...
       mktemp [-d] [-q] [-u] -t prefix
@(#)PROGRAM:mktemp  PROJECT:shell_cmds-203
⋊> ~/t/S/misc on master ⨯ mktemp /tmp/foo-XXXXXX
/tmp/foo-iObPoG
```

rgevaert:
yes, indeed, I can't remember why it didn't work 🤔

PR #63, opened 2019-09-05
Minor refactor of shell scripts, python and readme

guidodobboletta:
- Removed `tfvars` from `validate` as it will only work on directories which have `.tf` files
  - Directories with only `.tfvars` will make it fail (`Error: No Terraform configuration files found in directory`)
- Removed `.tfvars` from `tflint`
  - `Error: Failed to load 'test/only.tfvars': File is not a target of Terraform`
- Updated readme to latest release
- Modified `-a` to `--all-files` as I think the added clarity will be better for people not familiar with `pre-commit`
- Applied [black](https://github.com/psf/black) to python file
- Modified `dir`, `procArgs`
- Initiated `procargs` with single element so it looks better
- Removed the need to use the `__REPLACED__SPACE__` hack
- General shellcheck changes on bash scripts
- Run [shfmt](https://github.com/mvdan/sh) on bash scripts


guidodobboletta:
@antonbabenko I added a couple of optional arguments to hooks that I think would be very useful for some people

PR #70, opened 2019-10-11
update rev in README.md

davegallant:
Updating the rev in the README.

In order for `terraform_tflint` to work, the rev must be at least `v1.19.0`.


antonbabenko:
Thank you, this looks good. This is always a moving target :)

PR #74, opened 2019-10-31
[WIP] Run TFLint from repository root

njunxuan:
The current TFLint hook changes directory before running TFLint. This causes TFLint to ignore any `.tflint.hcl` config file that may be present in the repository root.

Is there some reason for changing directories before running TFLint? Thanks

Edit: marked this as WIP, as I noticed that TFLint will error occasionally as it is unable to parse sub-modules properly when not running from the module directory.

### DIFF
--- a/terraform_tflint.sh
+++ b/terraform_tflint.sh
@@ -17,7 +17,5 @@ done
 for path_uniq in $(echo "${paths[*]}" | tr ' ' '\n' | sort -u); do
   path_uniq="${path_uniq//__REPLACED__SPACE__/ }"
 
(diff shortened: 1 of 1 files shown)


antonbabenko:
[v1.21.0](https://github.com/antonbabenko/pre-commit-terraform/releases/v1.21.0) of this module has been just released.

It is now possible to specify `--config` as argument like this:
```
    - id: terraform_tflint
      args: ['--deep --config ~/my_tflint_config']
```

See this for more info - https://github.com/terraform-linters/tflint/blob/master/docs/guides/config.md

Issue #71, opened 2019-10-14
TFLint --deep argument


benmontour-wf:
I'd like to see the tflint pre-commit script have --deep be an optional argument rather than the default. I believe --deep only works with AWS and causes issues if you attempt to run it for GCP or Azure infrastructure.

antonbabenko:
[v1.21.0](https://github.com/antonbabenko/pre-commit-terraform/releases/v1.21.0) of this module has been just released.

It is now possible to specify it like this:
```
    - id: terraform_tflint
      args: ['--deep']
```

PR #53, opened 2019-06-19
Added support for TFLint with --deep parameter

costingalan:
Added support for TFLint (https://github.com/wata727/tflint).

This commit addresses #37 

Signed-off-by: Costin Galan <costin.galan@mambu.com>

PR #75, opened 2019-11-15
Allow arbitrary TFlint args (instead of --deep by default)

choppedpork:
This changes the behaviour of the TFlint hook from enforcing `--deep` to allowing arbitrary args to be specified (including, but not limited to, `--deep`) in the hook configuration.

Addresses #71 and a few comments in #53.

I'm in a similar boat to the other commenters - using this hook for non-AWS Terraform code, in my case to enforce some terraform rules rather than AWS rules (commented variables, commented outputs, pinned versions etc).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 <a name="unreleased"></a>
 ## [Unreleased]
 
-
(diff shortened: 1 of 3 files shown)


antonbabenko:
Thank you, @choppedpork, for putting this all together!

[v1.21.0](https://github.com/antonbabenko/pre-commit-terraform/releases/v1.21.0) of this module has been just released.

PR #77, opened 2019-11-26
tflint: iterate over files

robsonpeixoto:
It will avoid this bug:
```
/Users/robinho/.cache/pre-commit/repoxfm3uiv2/terraform_tflint.sh: line 42: pushd: modules/prometheus/lb.tf modules/prometheus/ec2.tf modules/prometheus/variables.tf modules/prometheus: No such file or directory
```

PR #78, opened 2019-11-29
terraform_tflint: Fix files array assignation

maxbrunet:
Files were concatenated into a single space-separated string

How `args` are passed to the `tflint` function might have to be improved too.

Here's `shellcheck` output:

```bash
$ shellcheck terraform_tflint.sh                                                                                                                                       [12:11:58]

In terraform_tflint.sh line 36:
    let "index+=1"
    ^-----------^ SC2219: Instead of 'let expr', prefer (( expr )) .


In terraform_tflint.sh line 43:
    tflint $args
           ^---^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean:
    tflint "$args"


In terraform_tflint.sh line 534:
[[ $BASH_SOURCE != "$0" ]] || main "$@"
   ^----------^ SC2128: Expanding an array without an index only gives the first element.

For more information:
  https://www.shellcheck.net/wiki/SC2128 -- Expanding an array without an ind...
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
  https://www.shellcheck.net/wiki/SC2219 -- Instead of 'let expr', prefer (( ...
```

### DIFF
--- a/terraform_tflint.sh
+++ b/terraform_tflint.sh
@@ -7,7 +7,7 @@ main() {
   eval "set -- $argv"
 
   declare args
-  declare files
+  declare -a files
 
   for argv; do
     case $argv in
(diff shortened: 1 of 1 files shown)


maxbrunet:
Closing in favor of #77

Issue #90, opened 2020-02-01
Failure on terraform 0.11.x


robinbowes:
The terraform_docs plugin is failing on terraform 0.11.x.

```
$ pre-commit run -a terraform_docs
Terraform docs...........................................................Failed
- hook id: terraform_docs
- exit code: 1
```

It's caused by a combination of `set -e` and this line:

```
hack_terraform_docs=$(terraform version | head -1 | grep -c 0.12)
```

This exits with error code 1 on terraform 0.11.x, and the `set -e` causes the script to exit.

Working on a fix...

robinbowes:
Fix in #91

PR #91, opened 2020-02-01
Fix for terraform 0.11.x

robinbowes:
Fix for #90 

Also includes a couple of bash lint fixes.


antonbabenko:
Terraform 0.11 is already history, so closing this one.

PS: It is probably already working if use the latest release.

PR #98, opened 2020-03-26
fix: Allow passing multiple args to terraform-docs

sergei-ivanov:
The `args` variable was mistakenly double-quoted in one of the recent commits, which made it impossible to pass multiple arguments to `terraform-docs`.
With this change it is possible to configure the hook like this:

```yaml
    hooks:
      - id: terraform_docs
        args:
          - '--args=document --indent 3'
```

### DIFF
--- a/terraform_docs.sh
+++ b/terraform_docs.sh
@@ -95,7 +95,8 @@ terraform_docs() {
     fi
 
     if [[ "$terraform_docs_awk_file" == "0" ]]; then
-      terraform-docs md "$args" ./ > "$tmp_file"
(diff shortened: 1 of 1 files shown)


eytanhanig:
I can confirm that your fix works for me.

@antonbabenko Any idea when we can expect to see this merged in?

antonbabenko:
Thanks @sergei-ivanov for the work and @eytanhanig for the verification and confirmation!

[v1.28.0](https://github.com/antonbabenko/pre-commit-terraform/releases/v1.28.0) has been just released.

PR #105, opened 2020-04-29
docs: Added coreutils as requirements in README.md

gchappell99:
Since the change to make terraform_validate recursive on April 4th, coreutils is required on macOS because there is no `realpath` available by default


antonbabenko:
Thanks @gchappell99 for the addition!

PR #106, opened 2020-04-29
docs: Fixed the docs to use the latest config syntax

snolan-uturn:
Added the top-level repos map that was introduced in pre-commit 1.0.0.


rquadling:
PHPStorm was complaining about this missing. I thought it was me. Glad it wasn't.

PR #114, opened 2020-06-07
feat: Add support for passing args and environment vars to terraform_validate

paullschock:
I set out to allow support for passing environment variables and arguments to `terraform_validate` in order to address: https://github.com/pre-commit/pre-commit/issues/758#issuecomment-608113130

I noticed passing multiple arguments to the `terraform_tflint` hook did not work as documented for two reasons:
1. documentation examples missing required '--' prefix for args
1. for loop in main would overwrite $args, thus passing only the last specified "args" item

I resolved this issue for `terraform_tflint` and carried this pattern into `terraform_validate`

Added/corrected readme documentation for two affected hooks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,42 @@ if they are present in `README.md`.
           - '--args=--enable-rule=terraform_documented_variables'
     ```
 
+## Notes about terraform_validate hooks
+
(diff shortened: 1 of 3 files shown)


paullschock:
Closing in favor of https://github.com/antonbabenko/pre-commit-terraform/pull/125 as the implementation appears to be superior.

PR #122, opened 2020-06-30
Honor excludes; don't run against entire directories (closes #121)

### LINKED ISSUE #121: terraform_fmt.sh doesn't respect excludes
`pre-commit` has the ability to exclude files either globally or per hook.

The `terraform_fmt.sh` accepts a list of files to format.  However, it ends up simply running `terraform fmt` on the dirname of all the files.  This means if a file which is explicitly excluded in the `.pre-commit-config.yaml` file is sitting in the same directory as a file which is not excluded, it'll end up getting formatted.

The `terraform_fmt.sh` script should stop trying to be clever and simply run terraform fmt against the files explicitly handed to it and nothing more.

ericfrederich:
Closes #121

### DIFF
--- a/terraform_fmt.sh
+++ b/terraform_fmt.sh
@@ -1,34 +1,9 @@
 #!/usr/bin/env bash
 set -e
 
-declare -a paths
-declare -a tfvars_files
-
-index=0
-
 for file_with_path in "$@"; do
   file_with_path="${file_with_path// /__REPLACED__SPACE__}"
-
-  paths[index]=$(dirname "$file_with_path")
-
-  if [[ "$file_with_path" == *".tfvars" ]]; then
-    tfvars_files+=("$file_with_path")
-  fi
-
-  let "index+=1"
-done
-
-for path_uniq in $(echo "${paths[*]}" | tr ' ' '\n' | sort -u); do
-  path_uniq="${path_uniq//__REPLACED__SPACE__/ }"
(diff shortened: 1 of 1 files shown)


bruce-szalwinski:
https://github.com/antonbabenko/pre-commit-terraform/blob/2e40ade8d0bdff4eea5fb64e55d6f40a71768426/CHANGELOG.md#v120---2017-06-08 the `exclude` option is documented way back in version 1.2.0.  Is that no longer working?  Was it working previously?

I ran into this while upgrading a module mono repo to 0.12.  My client is using the 0.11 version of the module but the 0.12 version of the module comes along for the ride.

When trying to commit my changes on a repo that uses the module, I get:

```bash
Error running fmt: In .terraform/modules/26b44daf48f2995f5f7159f55b3e58ba/0.12/datadog.tf: At 5:25: Unknown token: 5:25 IDENT var.monitor_name
```

I have my pre-commit configured as:

```yaml
  rev: v1.31.0
  hooks:
    - id: terraform_fmt
      exclude: \.terraform\/.*$
```

MaxymVlasov:
Works fine with the latest version

```yaml
repos:
  rev: v1.62.3
  hooks:
  - id: terraform_fmt
    exclude: environment\/.*$
```

`\.terraform\/.*$` excluded globally from Aug 2019

PR #118, opened 2020-06-15
fix: terraform_validate: accept --var-file option

robinbowes:
This change adds the `--var-file` option to `terraform_validate`.

I wrote it without realising `terraform_tflint` already supported command-line arguments and used an alternate algorithm, ie. not the re-implementation of `getopts` embedded in `terraform_tflint`.

I also missed that @paullschock had submitted #114 recently, which does something similar.

Finally, I've also just noticed that the `--var-file` option is not present in `0.12.x` (we're still on 0.11.x at the moment).

Fixes #116

PR #124, opened 2020-07-06
refactor: move getopt function into separate file

robinbowes:
This change refactors the pure bash getopt function into a separate file so it may be loaded and used by other scripts.

It also makes a few tweaks to keep shellcheck happy.

I will be submitting a further PR shortly to use the getopt function in terraform_validate, replacing both #114 and #118.

### DIFF
--- a/lib_getopt
+++ b/lib_getopt
@@ -1,0 +1,494 @@
+#!/bin/bash
+
+getopt() {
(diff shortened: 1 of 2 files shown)


robinbowes:
@paullschock + @antonbabenko  - please review.

antonbabenko:
Closing this one, because it has been already implemented in #125

PR #127, opened 2020-07-16
Fix to follow HTTP redirection when downloading latest terraform-docs

ryanellis:
The curl command to download the latest version of `terraform-docs` is failing due to a HTTP/301 redirection.
Adding `-L` will follow the redirection.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ brew install pre-commit gawk terraform-docs tflint tfsec coreutils
 ```bash
 sudo apt install python3-pip gawk &&\
 pip3 install pre-commit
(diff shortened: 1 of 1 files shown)


nicholasmhughes:
Should we just point to the correct location instead?

Old URL on line 27:
```
https://api.github.com/repos/segmentio/terraform-docs/releases/latest
```
Potential new URL on line 27:
```
https://api.github.com/repos/terraform-docs/terraform-docs/releases/latest
```

Not sure if the URL is a moving target.

ryanellis:
@nicholasmhughes - agreed, the latter doesn't redirect :-)

antonbabenko:
Thanks for the update, this has been just fixed by #130

Issue #128, opened 2020-07-20
terraform_docs hangs on Windows


matiaszilli:
When I execute `terraform_docs` precommit in Git Bash, it hangs.

The environment is Windows 10.

$ bash --version
GNU bash, version 4.4.23(1)-release (x86_64-pc-msys)

fbuinosquy:
same problem

MaxymVlasov:
Fix was merged

PR #129, opened 2020-07-20
fix: terraform_docs compatible with Git bash on Windows

matiaszilli:
Adds support for terraform_docs running on Git Bash over Windows environment.

It fixs #128

### DIFF
--- a/terraform_docs.sh
+++ b/terraform_docs.sh
@@ -26,7 +26,7 @@ main() {
   done
 
   local hack_terraform_docs
-  hack_terraform_docs=$(terraform version | head -1 | grep -c 0.12) || true
+  hack_terraform_docs=$(terraform version | sed -n 1p | grep -c 0.12) || true
 
   if [[ ! $(command -v terraform-docs) ]]; then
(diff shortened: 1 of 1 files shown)


fbuinosquy:
Please merge.

matiaszilli:
Please @antonbabenko can you merge it?. Thanks.

antonbabenko:
Good one! Thanks for the nudge, too :)

[v1.42.0](https://github.com/antonbabenko/pre-commit-terraform/releases/v1.42.0) has been just released.

robinbowes:
Is `awk` available in git bash on Windows?

If so, this will do the trick:
```
terraform version | awk 'NR > 1 { exit } ; /0.12/ { ++count } ; END { printf "%d\n", count }'
```
This awk program consists of three elements:

- `NR > 1 { exit }` - exit after the first line
- `/0.12/ { ++count }` - if the line matches the pattern, increment the `count` variable
- `END { printf "%d\n", count }` - when we exit (either at the end of the file, or if "exit" is called), print the value of `count` as a number

PR #130, opened 2020-07-25
docs: Update terraform-docs link pointing to new organization

khos2ow:
We've moved `terraform-docs` to its own organization at https://github.com/terraform-docs/

This PR actually supersedes #127


antonbabenko:
Thanks @khos2ow for the PR and great job on terraform-docs we all love and use (at least we should)!

PR #135, opened 2020-08-21
feature: have option for terraform_tfsec.sh to only run in relevant modified directories

nkazarian-spokeo:
This will allow fine-grain control in `.pre-commit-config.yml` for those who implement the hook to specify `tfsec` to run in only certain directories like so:

```
- id: terraform_tfsec
  files: ^dir1/
```

Maybe someone can point me to the rationale behind the comment stating that "`$files` should be ignored since `tfsec` will run recursively anyway" since I was able to get a working example with these changes. Perhaps there was a time when `tfsec` _only_ ran in a recursive fashion no matter what?

@jon-proietti-nutrien would like your input on this, as I noticed that you are the only contributor to this file on this repository.

Thanks.

**EDIT**: actually I believe this is more of an enhancement, as leaving out `files:` key will have `tfsec` run recursively, but now we have the option to filter it down since there's no `--include` type flag that you can pass to `tfsec` itself (they just offer `--exclude-dir`). I changed the title to appropriately reflect my intent.


nkazarian-spokeo:
@antonbabenko can you look at this again? I have resolved conflicts resulting from recent PR that involved changes to this file and back-ported my changes on top of it.

antonbabenko:
@nkazarian-spokeo Thanks for this PR and final rebase!

[v1.36.0](https://github.com/antonbabenko/pre-commit-terraform/releases/v1.36.0) has been just released.

nkazarian-spokeo:
Awesome!!! Appreciate it @antonbabenko. Thanks for looking.

PR #141, opened 2020-09-03
fix: Correctly handle arrays in terraform_docs.sh

sergei-ivanov:
@antonbabenko A serious regression has been introduced in v1.34.0 in the process of factoring out `getopt`. Starting with that version, `terraform_docs` pre-commit hook would only process one terraform module, ignoring any other modules that are part of the commit.

This PR attempts to fix the regression. I have tested it on Ubuntu Focal (GNU bash, version 5.0.17), but I would appreciate if anyone could also test it on Bash 4.x and on MacOS.

Please note that `args` are still passed around in the script as a single argument (just as it was), and I have also successfully verified the correct operation with multiple `--args=` in the pre-commit YAML configuration (fixed earlier in #98).

Side note: other hooks may also be affected in a similar way, but I haven't tested them because we are not using them.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 - repo: git://github.com/pre-commit/pre-commit-hooks
-  rev: v2.5.0
+  rev: v3.2.0
   hooks:
     - id: check-yaml
     - id: end-of-file-fixer
@@ -9,7 +9,7 @@ repos:
     - id: check-merge-conflict
     - id: check-executables-have-shebangs
 - repo: git://github.com/jumanjihouse/pre-commit-hooks
-  rev: 2.0.0
+  rev: 2.1.4
   hooks:
     - id: shfmt
       args: ['-l', '-i', '2', '-ci', '-sr', '-w']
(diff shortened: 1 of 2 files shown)


antonbabenko:
Thanks a lot, @sergei-ivanov !

I have verified it by running on [terraform-aws-security-group](https://github.com/terraform-aws-modules/terraform-aws-security-group) on Mac and it works as expected.

@robinbowes Could you please verify the code and maybe update other hooks also?

antonbabenko:
[v1.38.0](https://github.com/antonbabenko/pre-commit-terraform/releases/v1.38.0) has been just released.

robinbowes:
> @robinbowes Could you please verify the code and maybe update other hooks also?

Looks good to me. I shall investigate the other hooks.

robinbowes:
> [v1.38.0](https://github.com/antonbabenko/pre-commit-terraform/releases/v1.38.0) has been just released.

Did you push it? I don't see it on github?

antonbabenko:
Pushed now.

robinbowes:
@antonbabenko https://github.com/antonbabenko/pre-commit-terraform/pull/142 - needs regression testing before merging, I have requested that @sergei-ivanov takes a look.

PR #142, opened 2020-09-08
fix: Cleanup internal argument handling

robinbowes:
ARGS and FILES are global arrays - no need to pass them as function arguments

### DIFF
--- a/terraform_docs.sh
+++ b/terraform_docs.sh
@@ -4,7 +4,7 @@ set -eo pipefail
 main() {
   initialize_
   parse_cmdline_ "$@"
-  terraform_docs_ "${ARGS[*]}" "${FILES[@]}"
+  terraform_docs_
 }
 
 initialize_() {
@@ -47,10 +47,6 @@ parse_cmdline_() {
 }
 
 terraform_docs_() {
-  local -r args="$1"
-  shift
-  local -a -r files=("$@")
-
   local hack_terraform_docs
   hack_terraform_docs=$(terraform version | head -1 | grep -c 0.12) || true
 
@@ -64,7 +60,7 @@ terraform_docs_() {
 
   if [[ -z "$is_old_terraform_docs" ]]; then # Using terraform-docs 0.8+ (preferred)
 
-    terraform_docs "0" "$args" "${files[@]}"
+    terraform_docs "0"
 
   elif [[ "$hack_terraform_docs" == "1" ]]; then # Using awk script because terraform-docs is older than 0.8 and terraform 0.12 is used
 
@@ -76,28 +72,25 @@ terraform_docs_() {
     local tmp_file_awk
     tmp_file_awk=$(mktemp "${TMPDIR:-/tmp}/terraform-docs-XXXXXXXXXX")
     terraform_docs_awk "$tmp_file_awk"
-    terraform_docs "$tmp_file_awk" "$args" "${files[@]}"
+    terraform_docs "$tmp_file_awk"
     rm -f "$tmp_file_awk"
 
   else # Using terraform 0.11 and no awk script is needed for that
 
-    terraform_docs "0" "$args" "${files[@]}"
+    terraform_docs "0"
 
   fi
 }
 
 terraform_docs() {
-  local -r terraform_docs_awk_file="$1"
-  local -r args="$2"
-  shift 2
-  local -a -r files=("$@")
+  local -r terraform_docs_awk_file="$1" ; shift
 
   declare -a paths
   declare -a tfvars_files
 
   local index=0
   local file_with_path
-  for file_with_path in "${files[@]}"; do
+  for file_with_path in "${FILES[@]}"; do
     file_with_path="${file_with_path// /__REPLACED__SPACE__}"
 
     paths[index]=$(dirname "$file_with_path")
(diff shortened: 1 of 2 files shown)


robinbowes:
@sergei-ivanov Can you please take a look at this. I've cleaned up the argument handling in terraform_docs so we simply use the global arrays `ARGS` and `FILES` rather than trying to pass them around internally.

terraform_docs.sh
```diff
@@ -125,18 +118,18 @@ terraform_docs() {
 
     if [[ "$terraform_docs_awk_file" == "0" ]]; then
       # shellcheck disable=SC2086
-      terraform-docs md $args ./ > "$tmp_file"
+      terraform-docs md "${ARGS[*]}" ./ > "$tmp_file"
```
sergei-ivanov, line 121: `ARGS` needs to be unquoted to support multiple arguments like in #98 , otherwise `terraform-docs` errors out.
```suggestion
      terraform-docs md ${ARGS[*]} ./ > "$tmp_file"
```

terraform_docs.sh
```diff
@@ -76,28 +72,25 @@ terraform_docs_() {
...
 
   fi
 }
 
 terraform_docs() {
-  local -r terraform_docs_awk_file="$1"
-  local -r args="$2"
-  shift 2
-  local -a -r files=("$@")
+  local -r terraform_docs_awk_file="$1" ; shift
```
sergei-ivanov, line 86: There's not really much point in `shift` anymore, now that we are using the globals.
```suggestion
  local -r terraform_docs_awk_file="$1"
```

terraform_docs.sh
```diff
@@ -125,18 +118,18 @@ terraform_docs() {
...
-      mv "$tmp_file_docs" "$tmp_file_docs.tf"
       local tmp_file_docs_tf
+      tmp_file_docs=$(mktemp "${TMPDIR:-/tmp}/terraform-docs-XXXXXXXXXX")
       tmp_file_docs_tf="$tmp_file_docs.tf"
+      mv "$tmp_file_docs" "$tmp_file_docs_tf"
 
       awk -f "$terraform_docs_awk_file" ./*.tf > "$tmp_file_docs_tf"
       # shellcheck disable=SC2086
-      terraform-docs md $args "$tmp_file_docs_tf" > "$tmp_file"
+      terraform-docs md "${ARGS[*]}" "$tmp_file_docs_tf" > "$tmp_file"
```
sergei-ivanov, line 132: Same here:
```suggestion
      terraform-docs md ${ARGS[*]} "$tmp_file_docs_tf" > "$tmp_file"
```

terraform_tfsec.sh
```diff
@@ -4,25 +4,22 @@ set -eo pipefail
...
-
-    let "index+=1"
+    (( index+=1 ))
   done
 
   for path_uniq in $(echo "${paths[*]}" | tr ' ' '\n' | sort -u); do
     path_uniq="${path_uniq//__REPLACED__SPACE__/ }"
     pushd "$path_uniq" > /dev/null
-    tfsec $ARGS
+    tfsec "${ARGS[@]}"
```
sergei-ivanov, line 22: Same here probably:
```suggestion
    tfsec ${ARGS[*]}
```

terraform_docs.sh
robinbowes: > There's not really much point in `shift` anymore, now that we are using the globals.

This is more of a safety net - it will throw an error if there is no arg.

terraform_docs.sh
robinbowes: OK

terraform_docs.sh
robinbowes: OK

terraform_tfsec.sh
robinbowes: OK

robinbowes: Actually, I think using the quoted `@` form is the correct approach here, and in `terraform_docs`.

Can you re-check with 3a68667 and let me know what you think?

robinbowes: FYI, it seems work OK for me:
```
 $ pre-commit try-repo -a ~/code/pre-commit-terraform terraform_docs
===============================================================================
Using config:
===============================================================================
repos:
-   repo: ../../../../pre-commit-terraform
    rev: 3a68667efc0f95f9c01db4a79aff33075697fc2e
    hooks:
    -   id: terraform_docs
===============================================================================
[INFO] Initializing environment for ../../../../pre-commit-terraform.
- hook id: terraform_docs
```

sergei-ivanov: It does not work with args specified like this:

```yaml
    hooks:
      - id: terraform_docs
        args:
          - '--args=document --indent 3'
```

...which is expected. But it does work when args are split into individual tokens:

```yaml
    hooks:
      - id: terraform_docs
        args:
          - '--args=document'
          - '--args=--indent'
          - '--args=3'
```

I am not sure whether we would want to enforce passing the args one by one, or be more lenient like in the 1st example.

robinbowes: @antonbabenko - any thoughts?

antonbabenko: I am almost sure that `--args` should be without spaces, so users should pass multiple items as a list which I believe works.

robinbowes: I think that's a cleaner solution than trying to cope with eg. `'--args=document --indent 3'`

sergei-ivanov: Both options are fine by me. If we settle on option 2, I think we need a section in the documentation/readme with a usage example for additional arguments (feel free to use the example above). Simply because it is not obvious at all at the moment.

robinbowes: Agreed - the docs need updating.

Is your example correct, though? Wouldn't it be this:
```
hooks:
      - id: terraform_docs
        args:
          - 'document'
          - '--indent=3'
```
I think the other examples in the README may be wrong too.

sergei-ivanov: The one above is correct and tested. One needs to prepend each argument with `--args=` prefix. I guess that is because we want to distinguish the arguments to terraform-docs from the list of files passed in by pre-commit.

So the arguments should be passed like this:
```yaml
      - id: terraform_docs
        args:
          - '--args=document'
          - '--args=--indent'
          - '--args=3'
```

robinbowes: OK, so I think we're saying we should use one arg per line ie. option 2 but that we should make sure we update the documentation to be clearer.

jpatallah:
When is the slated to be merged in? Our tfsec exclusion parameter isn't being read properly and this fix is very important!

MaxymVlasov:
Nope, we should pass them to functions, because it explicitly mentions which vars go to function.

PR #151, opened 2020-09-24
fix: Fix regex considering terraform-docs v0.10.0 old

evanstoddard23:
Got this error using `v1.42.0` on terraform_docs hook:

```sh
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x13fdbe6]
```

Looks like the new `v0.10.0` terraform-docs is going the awk script route instead of "Using terraform-docs 0.8+ (preferred)" route because the regex matches when it shouldn't.

### DIFF
--- a/terraform_docs.sh
+++ b/terraform_docs.sh
@@ -60,7 +60,7 @@ terraform_docs_() {
   fi
 
   local is_old_terraform_docs
-  is_old_terraform_docs=$(terraform-docs version | grep -o "v0.[1-7]" | tail -1) || true
+  is_old_terraform_docs=$(terraform-docs version | grep -o "v0.[1-7]\." | tail -1) || true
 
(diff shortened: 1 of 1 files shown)


antonbabenko:
Thanks for such a timely fix!

[v1.43.0](https://github.com/antonbabenko/pre-commit-terraform/releases/v1.43.0) has been just released.

khos2ow:
Although this is a complete valid fix but the error was not completely related to the regex, it was a regression which was fixed in [v0.10.1](https://github.com/terraform-docs/terraform-docs/releases/tag/v0.10.1). I'd suggest to double check on your side for the new version too.

antonbabenko:
@khos2ow thanks for an update!

PR #152, opened 2020-10-02
Fix Terraform Validate failing because AWS_DEFAULT_REGION isn't set

Makeshift:
`terraform validate` will fail to run if `AWS_DEFAULT_REGION` isn't set, as of 0.13.

Setting it to `us-east-1` shouldn't matter, as Terraform does no provider checks anyway.

### DIFF
--- a/terraform_validate.sh
+++ b/terraform_validate.sh
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 set -eo pipefail
 
+export AWS_DEFAULT_REGION=us-east-1
+
 main() {
   initialize_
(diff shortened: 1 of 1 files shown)


ahmad-hamade:
@Makeshift I solved this issue by passing the region name as an args to the hook:

```
    rev: v1.43.0
    hooks:
      - id: terraform_validate
        args: ['--envs=AWS_DEFAULT_REGION="eu-west-1"']
```

You would still need to run `terraform init` before running the hook.

antonbabenko:
Fixed in #152

PR #156, opened 2020-11-01
fix: Correct deprecated parameter to terraform-docs

tolidano:
The --sort-inputs-by-required was removed here: https://github.com/terraform-docs/terraform-docs/commit/b6a6ad1bbf77e99fa920dee784b9936170b35477#diff-2fb3c316208521762f7701f5a16834dc239636c3b6c8ec403d02688e5f7da53eL23 after being marked for deprecation in favor of the --sort-by-required flag instead. and noted for deprecation here: https://github.com/terraform-docs/terraform-docs/releases/tag/v0.8.0

This PR switches to the new supported flag which has been available for 6 months now. 

Caveat: Existing terraform-docs users using 0.9.1 or older will need to upgrade. If relevant messaging is necessary, that can be added.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 - repo: git://github.com/pre-commit/pre-commit-hooks
-  rev: v3.2.0
+  rev: v3.3.0
   hooks:
     - id: check-yaml
     - id: end-of-file-fixer
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ if they are present in `README.md`.
     ```yaml
     hooks:
       - id: terraform_docs_replace
-        args: ['--with-aggregate-type-defaults', '--sort-by-required', '--dest=TEST.md']
+        args: ['--sort-by-required', '--dest=TEST.md']
     ```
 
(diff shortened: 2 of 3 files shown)


tolidano:
Alternatively, we can run `terraform-docs --version | cut -f4 -d' '` in a command first and by adding semver (v2.10.2 from June to support py2.7) or semantic-version (v2.8.5 from April) to the setup.py, check what parameter to use based on the version returned (<=0.9.1, use the old --sort-inputs-by-required, otherwise use --sort-by-required)

antonbabenko:
Thanks, @tolidano !

I think the provided solution is fine and there is no need to introduce more dependencies to `semver` and similar.

Could you please also update README and replace `--sort-inputs-by-required` there also?

tolidano:
I also noted that terraform-docs v0.10.0+ will be required in addition to updating the flag value in README.

tolidano:
Bumping this one

tolidano:
Bumping @antonbabenko

antonbabenko:
Thanks @tolidano !

[v1.45.0](https://github.com/antonbabenko/pre-commit-terraform/releases/v1.45.0) has been just released.

PR #158, opened 2020-11-02
feat: Make terraform_validate to run init if necessary

### LINKED ISSUE #157: terraform-validate fails if modules are not initialized
The hook `terraform-validate` will fail if the terraform modules are not initialized with `terraform init -backend=false`.

Any reason for not adding this missing step as part of the validation hook script like in `gruntwork-io/pre-commit` hooks?

https://github.com/gruntwork-io/pre-commit/blob/master/hooks/terraform-validate.sh#L12

antonbabenko:
Resolves #157

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 - repo: git://github.com/pre-commit/pre-commit-hooks
-  rev: v3.2.0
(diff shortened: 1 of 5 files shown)


antonbabenko:
[v1.44.0](https://github.com/antonbabenko/pre-commit-terraform/releases/tag/v1.44.0) has been just released.

PR #173, opened 2021-02-23
feat: Add new hook for `terraform providers lock` operation

sergei-ivanov:
After upgrade to Terraform 0.14 we faced a range of issues with dependency locking, and the only reasonable solution seemed to be a pre-commit hook that would force updating all hashes. It's a slow operation, but better than sporadic failures because of a missing hash.


MaxymVlasov:
Hi @sergei-ivanov 
Thank you for your contribution!
One question: do you still need to use this hook with the latest TF version tr this problem was gone?

antonbabenko (commented):
How will this feature behave if there are several directories with terraform files? Will it initialize in all of the directories and create a lock file in each? I don't think it is the right approach.

sergei-ivanov:
> do you still need to use this hook with the latest TF version or this problem was gone?

TL;DR Yes we still need it (for the lack of better options).

We are finally migrating to Terraform 1.0 with a big bang change scheduled for the next week. Unfortunately, Terraform team has not addressed the problems with the lock files, and dealing with them is still a royal PITA. I opened an issue https://github.com/hashicorp/terraform/issues/27811 that would have significantly alleviated the pain, but it has received zero attention so far.

> How will this feature behave if there are several directories with terraform files? Will it initialize in all of the directories and create a lock file in each? I don't think it is the right approach.

If you look at the hook configuration, the hook only triggers on changes in the lock files. No lock files will be created, only the existing ones will be updated. It will still need to run `terraform init` though, in the same way as `terraform_validate` hook works.
https://github.com/antonbabenko/pre-commit-terraform/blob/9ef1864e45e5daf00b9cd85dd1fceef85df7ca1e/.pre-commit-hooks.yaml#L51

antonbabenko:
@sergei-ivanov Thanks for the explanation. I missed the important part when asking the question :)

LGTM. Let's merge this one and hope for some reaction/improvement in https://github.com/hashicorp/terraform/issues/27811

antonbabenko:
@sergei-ivanov Could you please fix the issues to make status checks green - https://github.com/antonbabenko/pre-commit-terraform/runs/3769054913 ?

antonbabenko (approved):

sergei-ivanov:
OK, it's ready, I have fixed all linting issues.

antonbabenko:
Great! 🚀 

[v1.52.0](https://github.com/antonbabenko/pre-commit-terraform/releases/v1.52.0) has been just released.

PR #175, opened 2021-02-25
docs: Updated README regarding installation steps

mavogel:
updates readme regarding install of
- macOS
- ubuntu


antonbabenko:
Thanks for the update!

Issue #28, opened 2018-12-14
The hook terraform_docs_replace uses sed


RothAndrew:
The `terraform_docs_replace` hook uses `sed`, rather than taking care of the cleaning up of the document using native Python. Removing the dependency on `sed` will make this hook more platform agnostic.

antonbabenko:
Fixed in #176, [v1.47.0](https://github.com/antonbabenko/pre-commit-terraform/releases/v1.47.0) has been just released.

PR #176, opened 2021-02-25
fix: remove sed postprocessing from the terraform_docs_replace hook to fix compatibility with terraform-docs 0.11.0+

choppedpork:
Since updating to terraform-docs 0.11.0+ (also affects 0.11.1 and 0.11.2) all my README.md's are now missing the last line!

After a little bit of digging I've noticed that, since version 0.11.0, `terraform-docs` no longer adds an extra empty newline at the end. This PR amends the sed pattern to produce the same output in <0.11.0 as in 0.11.0+:

0.10.0:
```
$ terraform-docs --sort-by-required md ./ | gsed -e '$ d' -e 'N;/^\n$/D;P;D' | tail -n 3
|------|-------------|
| ids | List of IDs of the created instances |
| private\_ips | The private IP addresses assigned to the instances |

$ terraform-docs --sort-by-required md ./ | gsed -e '${/^$/d;}' -e 'N;/^\n$/D;P;D' | tail -n 3
|------|-------------|
| ids | List of IDs of the created instances |
| private\_ips | The private IP addresses assigned to the instances |
```

0.11.0:
```
$ terraform-docs --sort-by-required md ./ | gsed -e '$ d' -e 'N;/^\n$/D;P;D' | tail -n 3
| Name | Description |
|------|-------------|
| ids | List of IDs of the created instances |

$ terraform-docs --sort-by-required md ./ | gsed -e '${/^$/d;}' -e 'N;/^\n$/D;P;D' | tail -n 3
|------|-------------|
| ids | List of IDs of the created instances |
| private\_ips | The private IP addresses assigned to the instances |
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 <a name="unreleased"></a>
 ## [Unreleased]
 
+- fix: Removes sed post-processing from the `terraform_docs_replace` hook which was causing the last line to be missing when using `terraform-docs` 0.11.0+. Note: for older versions this change will result in an extra newline at the end of the file (making the pre-commit hook and `terraform-docs` output identical).
 
 
 <a name="v1.46.0"></a>
@@ -123,7 +124,7 @@ All notable changes to this project will be documented in this file.
 
 - fix: Change terraform_validate hook functionality for subdirectories with terraform files ([#100](https://github.com/antonbabenko/pre-commit-terraform/issues/100))
 
-### 
+###
 
 configuration for the appropriate working directory.
 
--- a/pre_commit_hooks/terraform_docs_replace.py
+++ b/pre_commit_hooks/terraform_docs_replace.py
@@ -29,7 +29,7 @@ def main(argv=None):
 
     dirs = []
     for filename in args.filenames:
-        if (os.path.realpath(filename) not in dirs and \
+        if (os.path.realpath(filename) not in dirs and
                 (filename.endswith(".tf") or filename.endswith(".tfvars"))):
             dirs.append(os.path.dirname(filename))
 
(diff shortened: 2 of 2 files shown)


antonbabenko:
I don't think we should modify the behavior and response of the tools we use in this project (eg, `terraform-docs`). It is going to be a moving target if we apply this PR and then `terraform-docs` updates.

choppedpork:
I'm not sure I follow @antonbabenko - at the moment (without this PR) the behaviour between pre-0.11 and later is different as well as incorrect. The sed pattern has been here since the hook was added in 2018. There's no documentation around _why_ it was added but my assumption is that it was trying to remove the last line (which until 0.11.0 was an extra empty line as well as a trailing newline). Since 0.11.0 an empty line is no longer being added and so the automatically generated documentation is incomplete. My suggested change _preserves the old behaviour_ for older versions while making sure that the resulting README is not missing a line (in the example I gave above - and every terraform module repository I'm currently working on at work - it's missing the last `output`).

I do agree with you that the underlying tool output should not be modified. I've created a pipeline to verify the pre-commit hooks have been applied by committers in my company and it took me a while to find why the output generated using `terraform-docs` was different to the output of the pre-commit hook (ie. the sed expressions found in this hook). I would be happy to re-raise a PR to _remove_ the sed command altogether if you think that would be a better solution, with this fix I have decided to err on the side of caution and preserve the existing behaviour for versions pre-0.11 rather than changing it.

antonbabenko:
I apologize for misunderstanding the root of the PR and closing it so fast (sometimes I do this very fast) :)

I think it will be better if we just remove this `sed` altogether. Please open a PR for that.

choppedpork:
Thanks!

PR updated, tested and rebased. I've amended the PR title as well as added a tentative changelog entry in order for the extra newline when using pre-0.11.0 versions to not surprise anyone.

Following the changed approach this will now also supersede #31 and resolve #28.

antonbabenko:
Thanks, @choppedpork !

[v1.47.0](https://github.com/antonbabenko/pre-commit-terraform/releases/v1.47.0) has been just released.

PS: Currently we autogenerate CHANGELOG from the PR messages.

PR #182, opened 2021-03-12
docs: Added checkov install

mavogel:
- Adds `checkov` to the install section as it was missing. 
- Are you generally interested in having a `Dockerfile` in the repo with the install instruction, so we can verify them in the future?

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@
 ##### MacOS
 
 ```bash
-brew install pre-commit gawk terraform-docs tflint tfsec coreutils
+brew install pre-commit gawk terraform-docs tflint tfsec coreutils checkov
 ```
 
 ##### Ubuntu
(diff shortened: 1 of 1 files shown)


antonbabenko:
I don't think it makes a lot of sense to have `Dockerfile` in this repo because we don't require docker for this to work.
A large part of the users does not use docker with pre-commit AFAIK (some do, but not the majority).

mavogel:
I too often messed up with `python` on my machine etc. so every time there is python involved I try to wrap it into a `Dockerfile`. Well I just build it locally and run all your tools for a repository as follows
```sh
docker run -v $(pwd):/lint -w /lint pre-commit pre-commit run -a
```
Sure entrypoint can be improved. I can file a PR and you can let me know your thoughts. A good side effect can be that the install stuff can be tested if it works :)

PR #185, opened 2021-03-24
chore: Fix typo on command in readme

serkef:
Command 
```bash
python3.7 -m pip install -U checkovpython3.7 -m pip install -U checkov
```
is probably a typo with one command being pasted twice.


antonbabenko:
Good catch, @serkef !

PR #192, opened 2021-04-06
fix: terragrunt validate for module folders (#191)

### LINKED ISSUE #191: terragrunt_validate - fails for folders without terragrunt.hcl
Suppose a multi-tier project (dev,stage, prod, etc) where not every folder contains a terragrunt.hcl. Non-module folders will only contain hcl files with some "local" definitions which get inherited from the module folders, e.g.

```
base.hcl
- tier1
  tier.hcl
  module1
    terragrunt.hcl
- tier2
  tier.hcl
  module1
    terragrunt.hcl
```

In this setup, terragrunt_validate will fail in folders where it can't find a terragrunt.hcl.

Proposed solution:
Only run `terragrunt validate` in folders, which contain a terragrunt.hcl

mhennecke:
Only include folders with a `terragrunt.hcl` file (default for `terragrunt validate`) in the terragrunt_validate pre-commit hook

Fixes #191

### DIFF
--- a/terragrunt_validate.sh
+++ b/terragrunt_validate.sh
@@ -17,7 +17,9 @@ done
 for path_uniq in $(echo "${paths[*]}" | tr ' ' '\n' | sort -u); do
   path_uniq="${path_uniq//__REPLACED__SPACE__/ }"
 
-  pushd "$path_uniq" > /dev/null
-  terragrunt validate
-  popd > /dev/null
+  if [[ -f "$path_uniq/terragrunt.hcl" ]]; then
+    pushd "$path_uniq" > /dev/null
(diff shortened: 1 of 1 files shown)


MaxymVlasov:
This PR does not provide a fix that can be accepted.

Firstly, `terragrunt_validate` hook run only in folders with `.hcl` files
https://github.com/antonbabenko/pre-commit-terraform/blob/ac9299cf5747910d46edd12f311a3fd024f1cb90/.pre-commit-hooks.yaml#L80-L86

Secondly, folks can create many `.hcl` files in one dir and/or name them different than `terragrunt.hcl`.

PR #193, opened 2021-04-20
fix: Fix and pin versions in Dockerfile

mavogel:
- use curl to obtain `tfsec` and use `latest` version as well and hence remove go as I got an error at
```sh
env GO111MODULE=on go get -u github.com/tfsec/tfsec/cmd/tfsec@${TFSEC_VERSION}
```
- use latest terraform version `0.15.0` by default
- reduce the docker image size by ~50%

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,9 +2,9 @@ FROM ubuntu:18.04
 
 ARG PRE_COMMIT_VERSION="2.11.1"
 ARG TERRAFORM_VERSION="0.15.0"
-ARG TFSEC_VERSION="latest" 
-ARG TERRAFORM_DOCS_VERSION="latest"
-ARG TFLINT_VERSION="latest"
+ARG TFSEC_VERSION="v0.39.21" 
+ARG TERRAFORM_DOCS_VERSION="v0.12.0"
+ARG TFLINT_VERSION="v0.27.0"
 ARG CHECKOV_VERSION="1.0.838"
 
 # Install general dependencies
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,42 +1,31 @@
 FROM ubuntu:18.04
 
 ARG PRE_COMMIT_VERSION="2.11.1"
-ARG GOLANG_VERSION="1.16"
-ARG TERRAFORM_VERSION="0.14.8"
-ARG TFSEC_VERSION="v0.39.6"
+ARG TERRAFORM_VERSION="0.15.0"
+ARG TFSEC_VERSION="latest" 
 ARG TERRAFORM_DOCS_VERSION="latest"
```
antonbabenko, line 6: Could you also update `TERRAFORM_DOCS_VERSION` to use 0.12.0 since this is what we use in GH Actions and in all terraform-aws-modules.

Ref: https://github.com/terraform-aws-modules/terraform-aws-lambda/blob/master/.github/workflows/pre-commit.yml#L97

mavogel: Sure. Then I decided to pin all versions to be consistent. WDYT?

antonbabenko: Yes, we only had some inconsistencies with `terraform-docs` but pinning everything sounds good.

antonbabenko:
Thank you!

[v1.49.0](https://github.com/antonbabenko/pre-commit-terraform/releases/v1.49.0) has been just released.

Issue #188, opened 2021-03-26
Allow to specify the version of terraform-docs for the hook to use


antonbabenko:
Currently, `terraform_docs` (and similar) hooks will use the version of `terraform-docs` installed on the system.

With each major release of the `terraform-docs` the output format changes and users need to regenerate the docs.

We need to be able to pin a specific version in the hook so that we have more control over `README.md`.

## Proposal:
- Use `args` to define a specific version of docker image - https://terraform-docs.io/user-guide/installation/#docker

## Sample:
```yaml
repos:
  - repo: git://github.com/antonbabenko/pre-commit-terraform
    rev: v1.48.0
    hooks:
      - id: terraform_fmt
      - id: terraform_docs
        args: ['--version=0.12']
```

antonbabenko:
@bryantbiggs I just want to link this one with the PR (eg, https://github.com/terraform-aws-modules/terraform-aws-alb/pull/190 ).

antonbabenko:
Yes, this is a good option. If it would be implemented in `pre-commit-terraform` maybe we can also take care of installing the required version (not sure if it is a good idea though).

khos2ow:
In my opinion, since using `.terraform-docs.yml` as a single source for configuration becomes more popular it makes sense to delegate these kind of configurations and restrictions to terraform-docs itself. But on the other hand installing the required version as part of the hook is gonna be very useful I think.

MaxymVlasov:
Can be possible to implement by https://github.com/antonbabenko/pre-commit-terraform/issues/418#issuecomment-2040209926

PR #194, opened 2021-04-22
feat: Add an option `--tf-version` to terraform_docs, terraform_fmt, and terraform_validate

svvitale:
Enables the use of a version parameter which is in turn used by `tfenv` to dynamically switch the version of terraform being used at runtime.

This feature is intended for users who may have more than one version of Terraform in use. Running `terraform fmt` can produce wildly different results depending on the version of Terraform being used, and by allowing the version to be set in the hook configuration, pre-commit can switch to the proper version, run the hook, and then switch back to the previously configured version to restore the original system state.

Please let me know if you have any questions at all. I did modify `terraform_fmt.sh` pretty extensively to look like the other shell scripts since it now needs to accept a parameter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 <a name="unreleased"></a>
 ## [Unreleased]
 
-
+- feat: Add option for using `tfenv` to switch terraform versions in terraform_docs, terraform_fmt, terraform_validate
 
 <a name="v1.49.0"></a>
 ## [v1.49.0] - 2021-04-20
@@ -145,7 +145,7 @@ All notable changes to this project will be documented in this file.
 
 - fix: Change terraform_validate hook functionality for subdirectories with terraform files ([#100](https://github.com/antonbabenko/pre-commit-terraform/issues/100))
 
-### 
+###
 
 configuration for the appropriate working directory.
 
--- a/README.md
+++ b/README.md
@@ -12,13 +12,14 @@
 * [`TFSec`](https://github.com/liamg/tfsec) required for `terraform_tfsec` hook.
 * [`coreutils`](https://formulae.brew.sh/formula/coreutils) required for `terraform_validate` hook on macOS (due to use of `realpath`).
 * [`checkov`](https://github.com/bridgecrewio/checkov) required for `checkov` hook.
+* [`tfenv`](https://github.com/tfutils/tfenv) if you wish to specify a desired version of terraform for the `terraform_fmt`, `terraform_validate`, or `terraform_docs` hooks.
 
 or build and use the Docker image locally as mentioned below in the `Run` section.
 
 ##### MacOS
 
 ```bash
-brew install pre-commit gawk terraform-docs tflint tfsec coreutils checkov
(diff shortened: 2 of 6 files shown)


antonbabenko:
Hi @svvitale !

Thanks for the proposal. I am not sure we need to parameterize each hook separately but instead...

I think we can have another hook named something like `use_terraform_version` which will configure the Terraform version based on:
1. `tfenv` from `.terraform-version` or via `args`.
2. `tfswitch` (https://tfswitch.warrensbox.com/ - I use this one myself and like it more). We don't need to support it right away but good to plan in advance.

Sample config:
```yaml
hooks:
  - id: use_terraform_version
    args: ['--tfenv --version=min-required']
    # args: ['--tfswitch --version=latest']
    # args: ['--tfenv --version=0.12.18']
    # args: ['--tfswitch --version="^0.12"']
    # args: ['--tfswitch --version="^0.12"']
```

The final PR will be much smaller.

What do you think?

svvitale:
Hi @antonbabenko! Thanks so much for the quick reply, I'm glad to hear that you would get some value out of this feature as well.

`tfswitch` looks really slick, I may actually move our usage to that instead.

I didn't realize that hooks were executed in sequential order, although that makes sense now that you highlighted it.  I think this is a MUCH better approach that isolates it from current and future hook code. My only concern is that I'd like to ensure that the previously configured terraform version is restored after the hooks complete. I suppose that can be an additional (and optional) hook that users can place at the end of their hooks list.

This approach would also not be compatible with the [fail_fast](https://pre-commit.com/#top_level-fail_fast) option, but that could be called out in the docs.

Let me know if you agree with the "restore" option and I'll get to work on the rest. Thanks again for taking the time to look through this.

antonbabenko:
Perfect!

There are some cases like the one you described (fail_fast) but even more often I run `pre-commit run -a terraform_fmt` which will have to be prefixed with `use_terraform_version`.

I agree it will be fine as long as we have `use_terraform_version` and `restore_terraform_version` at the end.

MaxymVlasov:
Relates #188

MaxymVlasov:
Close as too outdated. Simpler to create a new branch with specified changes than try to resolve conflicts

BartusZak:
Is there any possibility to stick with `terraform 1.5.7` for `terraform_validate`?
@MaxymVlasov @antonbabenko

MaxymVlasov:
Sure. Install `terraform` 1.5.7 locally and use it for validation proposes.

BartusZak:
@MaxymVlasov does docker image use latest terraform version?

Im using GitLab CI/CD Pipeline and wondering if I should use a decent version or simply install required terraform version before triggering `pre-commit run`.
e.g:

```yaml
precommit:
  image: ghcr.io/antonbabenko/pre-commit-terraform:latest
  script:
    - <----------- INSTALL TF here?
    - pre-commit run --all-files
```

MaxymVlasov:
You can build your own docker image with any tool version, including terraform

Check the `Docker` section at https://github.com/antonbabenko/pre-commit-terraform?tab=readme-ov-file#1-install-dependencies

PR #210, opened 2021-07-19
chore: Update tfsec install to include new URL and fix grep 

margaritakop:
This change is to:
1. reflect the recent acquisition of tfsec by aquasec that resulted in a url change for the installation source
2. update the grep command that now returns two results, hence the fix to select 1 (-m 1)


andormarkus:
Hi @margaritakop 

Please can you update the Dockerfile as well?

Thanks,
Andor

margaritakop:
Hi @andormarkus, See the command in the dockerfile updated too.

MaxymVlasov:
Wow, cool grep option. I will replace the recently merged `| head -n 1` on `-m 1`. 

Thank you for participate!

PR #228, opened 2021-09-17
fix: trigger terraform-docs on changes in lock files

sergei-ivanov:

- [ ] This PR fixes a bug.

The recent versions of `terraform-docs` extract provider version from terraform lock files. Starting from version 0.15.0, this behaviour can also be controlled by `--lockfile` command line flag. Since the new behaviour is active by default, add lock files to the pattern that triggers pre-commit hook, so that README files are updated on any changes in lock files.

Tested locally by overriding file pattern in `.pre-commit-config.yaml` inside the client project.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -12,7 +12,7 @@
   require_serial: true
   entry: terraform_docs.sh
   language: script
-  files: (\.tf)$
+  files: (\.tf|\.terraform\.lock\.hcl)$
   exclude: \.terraform\/.*$
 
 - id: terraform_docs_without_aggregate_type_defaults


antonbabenko:
Thanks for the PR.

I think it makes sense to treat lockfile a bit differently and maybe we need to make a separate hook that will watch for changes in lockfile like you are proposing, but the default behavior should be as it is now.

What do you think?

sergei-ivanov:
I don't think there is any point in creating a separate hook, it will only end up in a proliferation of the hooks for no good reason. I can always override the file pattern locally, but I'll need to do it across multiple projects. The functionality is part of `terraform-docs` now, and I felt that the existing hook should reflect that. It's fully backwards compatible (older versions of `terraform-docs` won't produce any changes on a change in lock file). And the latest version of `terraform-docs` will still pull version information from the lock file by default, on any change in `*.tf` files. So this change only makes it more consistent: the hook will get triggered when either of the inputs changes.

sergei-ivanov:
Background: I had a local change in the repo, a few modules had lock files added, and some modules had .tf file changes too. And after applying pre-commit I noticed that provider version was updated in the generated docs only for the modules with .tf changes. That looked inconsistent to me, and that's why I decided to update the pattern.

antonbabenko:
I agree, here we go.

[v1.51.0](https://github.com/antonbabenko/pre-commit-terraform/releases/v1.51.0) has been just released.

PR #229, opened 2021-09-17
fix: remove dead code from terraform-docs script

sergei-ivanov:

- [x] This PR fixes a bug.

`terraform-docs` does not care about tfvars, the variable is unused inside the script
and probably was originally copy-pasted from another hook.

Tested locally against a client project.

### DIFF
--- a/terraform_docs.sh
+++ b/terraform_docs.sh
@@ -93,7 +93,6 @@ terraform_docs() {
   local -a -r files=("$@")
 
   declare -a paths
-  declare -a tfvars_files
 
   local index=0
   local file_with_path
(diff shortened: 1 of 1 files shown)


sergei-ivanov:
Not sure why the test fails, I think it does not have `shfmt` executable available in the PATH.
I have re-run `shfmt` locally against the branch and it did not pick up any changes.

MaxymVlasov (commented):
Hi
Thank you for contribution!

I will fix pre-commit in a week.
About PR - LGTM, but need confirmation from @antonbabenko , maybe some undocumented logic used across tfvars or some not yet implemented feature.

antonbabenko:
I don't think this code is needed. Maybe it was needed in previous versions but now it is definitely dead, so I will merge it.

Thanks, @sergei-ivanov !

PR #283, opened 2021-11-16
fix: typo in arg name for terraform-docs

mj3c:

- [x] This PR fixes a bug.

As someone who often does not copy commands/args but types them out, I encountered an issue when trying to add the argument `--add-to-existing-file` for `terraform_docs`, as it has a typo in the name currently. This PR updates `exiting` to `existing` in all places.

This, however, is a breaking change for people that are using this argument with its current name. In order to make it non-breaking, we can check for `--add-to-exiting-file|--add-to-existing-file)` in the `case` block maybe?

Tested using this:

```yaml
repos:
  - repo: git://github.com/mj3c/pre-commit-terraform
    rev: master
    hooks:
      - id: terraform_docs
        args:
```

### DIFF
--- a/terraform_docs.sh
+++ b/terraform_docs.sh
@@ -132,7 +132,7 @@ terraform_docs() {
       --path-to-file)
         text_file=$value
         ;;
-      --add-to-existing-file)
+      --add-to-existing-file|--add-to-exiting-file) # Typo left for compatibility. Will removed in 2.0.0.
         add_to_existing=$value
         ;;
       --create-file-if-not-exist)
--- a/terraform_docs.sh
+++ b/terraform_docs.sh
@@ -119,7 +119,7 @@ terraform_docs() {
   # Get hook settings
   #
   local text_file="README.md"
-  local add_to_exiting=false
+  local add_to_existing=false
   local create_if_not_exist=false
 
   configs=($hook_config)
@@ -132,8 +132,8 @@ terraform_docs() {
       --path-to-file)
         text_file=$value
         ;;
-      --add-to-exiting-file)
-        add_to_exiting=$value
+      --add-to-existing-file)
+        add_to_existing=$value
         ;;
       --create-file-if-not-exist)
         create_if_not_exist=$value
@@ -171,10 +171,10 @@ terraform_docs() {
     [[ ! -f "$text_file" ]] && popd > /dev/null && continue
 
     #
-    # If `--add-to-exiting-file=true` set, check is in file exist "hook markers",
+    # If `--add-to-existing-file=true` set, check is in file exist "hook markers",
     # and if not - append "hook markers" to the end of file.
     #
-    if $add_to_exiting; then
+    if $add_to_existing; then
       HAVE_MARKER=$(grep -o '<!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->' "$text_file" || exit 0)
 
       if [ ! "$HAVE_MARKER" ]; then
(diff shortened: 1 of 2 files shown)


yermulnik:
@mj3c Yay, nice catching this weird typo! Thanks.

> In order to make it non-breaking, we can check for --add-to-exiting-file|--add-to-existing-file) in the case block maybe?

I'd suggest using bash built-in functionality to improve your suggestion though, like this:
```
    # Get `extglob` shell opt status
    EXTGLOB_STATUS=$(shopt -p extglob)
    # Turn `extglob` on
    shopt -s extglob
    case $key in
      --path-to-file)
        text_file=$value
        ;;
      --add-to-exi?(s)ting-file) # https://github.com/antonbabenko/pre-commit-terraform/pull/283
        add_to_existing=$value
        ;;
      --create-file-if-not-exist)
        create_if_not_exist=$value
        ;;
    esac
    # Revert `extglob` status if it was unset before the `case` block
    [[ $EXTGLOB_STATUS == "shopt -u extglob" ]] && shopt -u extglob
```
@mj3c @MaxymVlasov @antonbabenko Does it look sensible?

ps: I'm all fine with `--add-to-exiting-file|--add-to-existing-file)` though since it _affects_ only one single line of code comparing with what I suggest :stuck_out_tongue:

terraform_docs.sh
```diff
@@ -132,8 +132,8 @@ terraform_docs() {
       --path-to-file)
         text_file=$value
         ;;
-      --add-to-exiting-file)
-        add_to_exiting=$value
+      --add-to-existing-file)
```
MaxymVlasov, line 135:
```suggestion
      --add-to-existing-file|--add-to-exiting-file) # Typo left for compatibility. Will removed in 2.0.0.
```

MaxymVlasov (approved):

MaxymVlasov:
@mj3c yeah, my bad, thank you!

@yermulnik built-in functionality is good, but I prefer to make code more explicit and more suitable for future removal

terraform_docs.sh
```diff
@@ -132,8 +132,8 @@ terraform_docs() {
       --path-to-file)
         text_file=$value
         ;;
-      --add-to-exiting-file)
-        add_to_exiting=$value
+      --add-to-existing-file|--add-to-exiting-file) # Typo left for compatibility. Will removed in 2.0.0.
```
MaxymVlasov, line 135:
```suggestion
      --add-to-existing-file|--add-to-exiting-file) # Typo left for compatibility. Will be removed in 2.0.0.
```

terraform_docs.sh
```diff
@@ -132,8 +132,8 @@ terraform_docs() {
       --path-to-file)
         text_file=$value
         ;;
-      --add-to-exiting-file)
-        add_to_exiting=$value
+      --add-to-existing-file|--add-to-exiting-file) # Typo left for compatibility. Will be removed in 2.0.0.
```
MaxymVlasov, line 135: @antonbabenko just said that we don't need such greater backward compatibility, so...
 
```suggestion
      --add-to-existing-file)
```

antonbabenko:
[v1.57.0](https://github.com/antonbabenko/pre-commit-terraform/releases/v1.57.0) has been just released.

terraform_docs.sh
yermulnik: > @antonbabenko just said that we don't need such greater backward compatibility, so...

IMHO, it's not about **we** or **need**. It's about being polite and careful about consumers by adhering to `semver`. As such we could have had this "quirk" removed within future `2.0.0` release (alongside announcing this). IMHO.

antonbabenko: There are 3 users (including Maxym) of this argument (GitHub code search) :) In general, I agree that we need to keep compatibility.

PR #289, opened 2021-11-30
fix: analyse all folders with tflint and don't stop on first execution

bmbferreira:

- [x] This PR fixes a bug.

In a monorepo tflint is currently stopping after analysing the first module. This is because of the `set -e` that causes the script to exit. This change disables it temporarily for executing `tflint`, catches the exit codes from the `tflint` executions and then uses the exit code for the function. It also removes the `pushd`/`popd` since it is not necessary because tflint can just receive the path as an argument.

Create a mono repo with two or more modules in different folders, with code that makes tflint report something. Notice that before this change, it just analyses one of the modules. With this change it will keep analysing the modules and report the output for all of them.

### DIFF
--- a/terraform_tflint.sh
+++ b/terraform_tflint.sh
@@ -58,20 +58,24 @@ tflint_() {
 
     ((index += 1))
   done
-
+  set +e
+  tflint_final_exit_code=0
   for path_uniq in $(echo "${paths[*]}" | tr ' ' '\n' | sort -u); do
     path_uniq="${path_uniq//__REPLACED__SPACE__/ }"
-    pushd "$path_uniq" > /dev/null
 
     # Print checked PATH **only** if TFLint have any messages
     # shellcheck disable=SC2091 # Suppress error output
-    $(tflint "${ARGS[@]}" 2>&1) 2> /dev/null || {
+    $(tflint "${ARGS[@]}" $path_uniq 2>&1) 2> /dev/null || {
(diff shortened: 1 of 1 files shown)


MaxymVlasov (approved):
Checked, it works fine.
LGTM

antonbabenko:
@bmbferreira Please fix the failing CI.

yermulnik (approved):
LGTM too 👍🏻

yermulnik:
@antonbabenko 
> @bmbferreira Please fix the failing CI.

The error doesn't seem to be related to this PR: https://github.com/antonbabenko/pre-commit-terraform/runs/4461845751?check_suite_focus=true#step:8:77

antonbabenko:
@yermulnik Magic "пендаль" fixed the build.

yermulnik:
Hahaha, great example of mastering a kick up the arse mate 👍🏻

bmbferreira:
Thank you for merging the PR 🙇‍♂️ 
Really happy that I was able to do a small contribution for this useful repo! 😄  Thank you for your work 🙌

maxbrunet:
> It also removes the `pushd`/`popd` since it is not necessary because tflint can just receive the path as an argument.

This is needed, some `tflint` flags like `--module` needs the workspace directory to be initialized. And like `terraform`, `tflint` picks up the `.terraform` directory from the current working directory

PR #301, opened 2021-12-13
feat: Add terraform_validate support to init modules and providers if needed

### LINKED ISSUE #224: feat: Add argument like `force-init` to do init before terraform validate
Add an argument to `terraform_validate` to run `terraform init` before running `terraform validate`.

_Originally posted by @eytanhanig in https://github.com/antonbabenko/pre-commit-terraform/issues/203#issuecomment-918791847_

Notes fo contributors:
* [`--hook-config` support example](https://github.com/antonbabenko/pre-commit-terraform/blob/master/terraform_docs.sh#L32-L46)
* [K/V that passed to `--hook-config` support example](https://github.com/antonbabenko/pre-commit-terraform/blob/master/terraform_docs.sh#L118-L142)

---

Additionally to `force-init` option, we can use `terraform providers` and run `terraform init` if necessary, automatically. See this comment for more details - https://github.com/antonbabenko/pre-commit-terraform/issues/299#issuecomment-992323031

jgrumboe:

- [x] This PR fixes a bug.

This PR adds logic to `terraform_validate` to detect if module initialization is needed and run it. It's also possible now to force an initialization and/or provider upgrade by just adding the documented `init-args` to the hook call.

The new logic works as follows:
`terraform providers` is called on every run and used to detect if all needed modules are initialized correctly. If it succeeds it will pass `-get=false` to the `terraform init` call, which prevents modules to be re-downloaded. On the other hand, if `terraform providers` fails, this means modules or providers are not initialized and `-get=true`will be passed to `terraform init` - a full initialization will happen.

Since `terraform init` will be called every time, the already existing `--init-args` feature will give us to additional benefits:
- `--init-args=-get=true` will result in a forced initialization on every run
- `--init-args=-upgrade` will allow to upgrade providers in addition to the modules if needed

Fixes #224 

1. You need to have terraform code which references modules and providers.
2. Delete `.terraform` folder if present
3. Change a module version in your code to some other existing version and stage the commit
4. run `pre-commit run terraform_validate` and you see that it will initialize the `.terraform` folder newly and validate your code.
5. Every re-run of `pre-commit run terraform_validate` will be faster, as long as no module version changes happen
6. Again change a module version in your code to some other existing version and stage the commit
7. run `pre-commit run terraform_validate` and you see that it will reinitialize for the new module

Repeat the above steps with changing provider version and setting the `--init-args=-upgrade` argument. You will notice that only providers or both are initialized if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -534,22 +534,21 @@ Example:
         - --init-args=-lockfile=readonly
     ```
 
-4. It may happen that Terraform working directory (`.terraform`) already exists but not in the best condition (eg, not initialized modules, wrong version of Terraform, etc.). To solve this problem, you can find and delete all `.terraform` directories in your repository:
+4. `terraform_validate` hook identifies automatically, if all needed modules are initialized and will reinitialize them if needed before running the `terraform validate` command. If you want to force a module initialization on every run, pass the following argument to its `terraform init`:
 
-    ```bash
-    echo "
-    function rm_terraform {
-        find . -name ".terraform*" -print0 | xargs -0 rm -r
-    }
-    " >>~/.bashrc
-
-    # Reload shell and use `rm_terraform` command in the repo root
+    ```yaml
+    - id: terraform_validate
+      args:
+        - --init-args=-get=true
     ```
 
-   `terraform_validate` hook will try to reinitialize them before running the `terraform validate` command.
+5. Pass the following argument, if you want `terraform_validate` hook to also upgrade all specified providers:
 
-    **Warning:** If you use Terraform workspaces, DO NOT use this workaround ([details](https://github.com/antonbabenko/pre-commit-terraform/issues/203#issuecomment-918791847)). Wait to [`force-init`](https://github.com/antonbabenko/pre-commit-terraform/issues/224) option implementation.
-
+    ```yaml
+    - id: terraform_validate
+      args:
+        - --init-args=-upgrade
+    ```
 
 ## Authors
 
--- a/README.md
+++ b/README.md
@@ -542,7 +542,7 @@ Example:
         - --init-args=-get=true
     ```
 
-5. Pass the following argument, if you want `terraform_validate` hook to also upgrade all specified providers:
+5. Pass the following argument, if you want `terraform_validate` hook to also upgrade all specified providers and (if `-get=true` is specified) to (re-)install latest versions of module(s) allowed within configured constraints:
 
     ```yaml
     - id: terraform_validate
--- a/README.md
+++ b/README.md
@@ -543,13 +543,17 @@ Example:
         - --init-args=-get=true
     ```
 
-5. Pass the following argument, if you want `terraform_validate` hook to also upgrade all specified providers and (if `-get=true` is specified) to (re-)install latest versions of module(s) allowed within configured constraints:
+5. If you want `terraform_validate` hook upgrade all specified providers and install the latest versions of modules allowed within configured constraints, set:
 
     ```yaml
+```suggestion
     - id: terraform_validate
       args:
         - --init-args=-upgrade
+        - --init-args=-get=true
     ```
+
+    For all possible `terraform init` args, check [official docs](https://www.terraform.io/docs/cli/commands/init.html).
 
 ### terrascan
 
--- a/terraform_validate.sh
+++ b/terraform_validate.sh
@@ -90,19 +90,20 @@ terraform_validate_() {
 
       pushd "$(realpath "$path_uniq")" > /dev/null
 
-      if [[ ! -d .terraform ]]; then
-        set +e
-        init_output=$(terraform init -backend=false "${INIT_ARGS[@]}" 2>&1)
-        init_code=$?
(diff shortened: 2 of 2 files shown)


antonbabenko:
Looks pretty good. Let's see what @yermulnik or @MaxymVlasov have to say after they test it locally.

README.md
```diff
@@ -534,22 +534,21 @@ Example:
...
+    ```yaml
+    - id: terraform_validate
+      args:
+        - --init-args=-get=true
     ```
 
-   `terraform_validate` hook will try to reinitialize them before running the `terraform validate` command.
-
-    **Warning:** If you use Terraform workspaces, DO NOT use this workaround ([details](https://github.com/antonbabenko/pre-commit-terraform/issues/203#issuecomment-918791847)). Wait to [`force-init`](https://github.com/antonbabenko/pre-commit-terraform/issues/224) option implementation.
+5. Pass the following argument, if you want `terraform_validate` hook to also upgrade all specified providers:
```
yermulnik, line 545:
```suggestion
5. Pass the following argument, if you want `terraform_validate` hook to also upgrade all specified providers and (if `-get=true` is specified) to (re-)install latest versions of module(s) allowed within configured constraints:
```

yermulnik (approved):
Nice catch on using `terraform providers` to validate modules installation.
Code looks good to me.

README.md
jgrumboe: @yermulnik  Am I allowed to commit your suggestion directly from GitHub UI or do you want to have PRs with 1 commit and force-pushed?

yermulnik: I'm all fine with committing suggestions from UI (which is what they are for) if @antonbabenko as owner has no objections.

antonbabenko: Looks good to me. AFAIK the PR owner (@jgrumboe) should be able to commit suggestions.

README.md
```diff
@@ -534,22 +534,21 @@ Example:
...
+    ```yaml
+    - id: terraform_validate
+      args:
+        - --init-args=-get=true
     ```
 
-   `terraform_validate` hook will try to reinitialize them before running the `terraform validate` command.
-
-    **Warning:** If you use Terraform workspaces, DO NOT use this workaround ([details](https://github.com/antonbabenko/pre-commit-terraform/issues/203#issuecomment-918791847)). Wait to [`force-init`](https://github.com/antonbabenko/pre-commit-terraform/issues/224) option implementation.
+5. Pass the following argument, if you want `terraform_validate` hook to also upgrade all specified providers and (if `-get=true` is specified) to (re-)install latest versions of module(s) allowed within configured constraints:
```
MaxymVlasov, line 545:
```suggestion
5. If you want `terraform_validate` hook upgrade all specified providers and install the latest versions of modules allowed within configured constraints, set:
```

README.md
```diff
@@ -534,22 +534,21 @@ Example:
...
 
+    ```yaml
+    - id: terraform_validate
+      args:
+        - --init-args=-upgrade
+    ```
```
MaxymVlasov, line 551:
```suggestion
```suggestion
    - id: terraform_validate
      args:
        - --init-args=-upgrade
        - --init-args=-get=true
    ```

    For all possible `terraform init` args, check [official docs](https://www.terraform.io/docs/cli/commands/init.html).
```

MaxymVlasov (changes requested):

terraform_validate.sh
```diff
@@ -90,19 +90,20 @@ terraform_validate_() {
 
       pushd "$(realpath "$path_uniq")" > /dev/null
 
-      if [[ ! -d .terraform ]]; then
```
MaxymVlasov, line 93: You should return that check and add check that checks that any `--init-args` provided.


Because, time execution increased up to 80-90% for:

```yaml
repos:
- repo: git@github.com:jgrumboe/pre-commit-terraform.git
  rev: '4086982'
  hooks:
 - id: terraform_validate
 ```

compare with:

```yaml
repos:
  rev: v1.62.0
  hooks:
 - id: terraform_validate
```

MaxymVlasov: No need to run `terraform init` each time if it is not specified by arguments.
Or need to explicitly provide `force-init: true` etc. for that.

For now, it just looks like no needed speed violation.

jgrumboe: Thanks for your input.
I will try to come up with a better solution, which is fast as now when no modules/providers version changes were made in the code and still eliminates the need for manual deletion of the `.terraform` folder in case modules/providers version changes are made.

MaxymVlasov: Hi @jgrumboe 
Any updates? Do you need any help?

jgrumboe: Hi @MaxymVlasov 
Sorry, this one slipped out of my mind. I will have a look next week.
Thanks for the reminder.

MaxymVlasov:
Close as too outdated. Simpler to create a new branch with specified changes than try to resolve conflicts

PR #305, opened 2021-12-19
fix: Check all directories with changes and pass all args in terrascan hook

carlosbustillordguez:
Put an `x` into the box if that apply:

- [ ] This PR introduces breaking change.
- [x] This PR fixes a bug.
- [ ] This PR adds new functionality.
- [ ] This PR enhances existing functionality.

### Description of your changes

This PR introduces the following changes:

- Configure `terrascan` hook to run only with Terraform config files.
- Pass to the `terrascan` command all supplied arguments.
- Allow inspecting each directory with TF config files for the current commit and fetch the final `terrascan` exit code.
- Add hook usage notes and an example for `terrascan`.

### How has this code been tested

The current implementation of `terrascan.sh` only adds to the `terrascan` command the first argument of the `ARGS` array. Try to use the following configuration to reproduce the problem (the `--policy-type azure` is never passed as an argument to `terrascan`):

```yaml
repos:
- repo: https://github.com/antonbabenko/pre-commit-terraform
  rev: v1.62.1 # Get the latest from: https://github.com/antonbabenko/pre-commit-terraform/releases
  hooks:
    - id: terraform_validate
    - id: terraform_fmt
      args:
        - --args=-no-color
        - --args=-diff
        - --args=-write=false
    # Detect compliance and security violations across IaC to mitigate risk before provisioning cloud native infrastructure.
    - id: terrascan
      args:
        - --args=--non-recursive # avoids scan errors on subdirectories without Terraform config files
        - --args=--policy-type azure
    # Security scanner for your Terraform code.
    - id: terraform_tfsec
    # A Pluggable Terraform Linter
    - id: terraform_tflint
      args:
        - --args=--config=__GIT_WORKING_DIR__/.tflint.hcl
        - --args=--module

```

`terrascan` is executed over all committed files, which produce errors on subdirectories without Terraform config files:

```bash
$ git commit -m "test1"
Terraform validate.......................................................Passed
Terraform fmt............................................................Passed
terrascan................................................................Failed
- hook id: terrascan
- exit code: 1

2021-12-19T23:09:08.531+0100    error   cli/run.go:132  scan run failed{error 26 0  1 error occurred:
        * directory '/home/carlos/devel/azure/terraform/azure-kubernetes-services-iac/modules/cert-manager/templates' has no terraform config files

}

Terraform validate with tfsec............................................Passed
Terraform validate with tflint...........................................Passed
```

Also, only the first directory of the `FILES` array is only analyzed with `terrascan`.

PR #307, opened 2021-12-21
fix: Speedup `terrascan` hook up to x3 times in big repos

MaxymVlasov:

- [ ] This PR fixes a bug.

During https://github.com/antonbabenko/pre-commit-terraform/pull/305#pullrequestreview-837797221 found that tag v1.62.1 with `require_serial: true` took 9s when without taking took 20s.

The root cause is same to previous `require_serial: true`  fixes - `pre-commit` run hook multiply times in the same folder/

So, let's not wait to PR merge, and implement this fix now.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -108,3 +108,4 @@
   description: Runs terrascan on Terraform templates.
   language: script
   entry: terrascan.sh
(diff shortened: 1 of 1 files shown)


antonbabenko:
This PR is included in version 1.62.2 :tada:

PR #155, opened 2020-10-21
feat: simplify args inputs for `terraform_docs`

1oglop1:
This should fix #154.
I removed `getopt` and instead simply passed args array in the same order as specified via `.pre-commit-config.yaml` file.

example usage:

```
  - repo: https://github.com/antonbabenko/pre-commit-terraform
    rev: <version>
    hooks:
      - id: terraform_docs
        verbose: true
        name: Terraform docs
        description: Inserts input and output documentation into README.md (using terraform-docs).
        require_serial: true
        entry: _ignore/pre-commit-terraform/terraform_docs.sh
        language: script
        files: (\.tf)$
        exclude: \.terraform\/.*$
        args:
          - --header-from=doc.tf
          - --hide-all
          - md
```

PR #310, opened 2021-12-23
chore: Refactor all hooks

MaxymVlasov:

- [ ] This PR fixes a bug.

Make hooks the as same as possible 
The next big thing will be to move the `common::` functions to separate file, and import them to all hooks.
Then, most of the hooks will look like this:

```bash
#!/usr/bin/env bash

. common.sh

function main {
  common::initialize
  common::parse_cmdline "$@"
  common::per_dir_hook "${ARGS[*]}" "${FILES[@]}"
}


function per_dir_hook_unique_part {
  # common logic located in common::per_dir_hook
  local -r args="$1"
  local -r dir_path="$2"

  # pass the arguments to hook
  # shellcheck disable=SC2068 # hook fails when quoting is used ("$arg[@]")
  terrascan scan -i terraform ${args[@]}

  local exit_code=$?
  return $exit_code
}

[[ ${BASH_SOURCE[0]} != "$0" ]] || main "$@"
```

```yaml
repos:
  rev: a243f3f314de907b5c964d7adcb9e9fa128b7ac9
  hooks:
# all hooks
```

---

Merge will affect:
* https://github.com/antonbabenko/pre-commit-terraform/pull/155 (introduce small, but breaking changes)
* https://github.com/antonbabenko/pre-commit-terraform/pull/194 (already have big conflicts)

### DIFF
--- a/terraform_docs.sh
+++ b/terraform_docs.sh
@@ -370,4 +370,4 @@ EOF
 
 }
 
-[[ ${BASH_SOURCE[0]} != "$0" ]] || main "$@"
+[ "${BASH_SOURCE[0]}" != "$0" ] || main "$@"
--- a/terraform_docs.sh
+++ b/terraform_docs.sh
@@ -23,7 +23,6 @@ function common::parse_cmdline {
   # common global arrays.
   # Populated via `common::parse_cmdline` and can be used inside hooks' functions
   declare -g -a ARGS=() FILES=() HOOK_CONFIG=()
-  declare -g -r HOOK_CONFIG_SEPARATOR='%%SEPARATOR%%'
 
   local argv
   argv=$(getopt -o a:,h: --long args:,hook-config: -- "$@") || return
@@ -38,7 +37,7 @@ function common::parse_cmdline {
         ;;
       -h | --hook-config)
         shift
-        HOOK_CONFIG+=("${1}${HOOK_CONFIG_SEPARATOR}")
+        HOOK_CONFIG+=("$1;")
         shift
         ;;
       --)
@@ -57,7 +56,7 @@ function terraform_docs_ {
   local -a -r files=("$@")
 
   # Get hook settings
-  IFS="$HOOK_CONFIG_SEPARATOR" read -r -a configs <<< "$hook_config"
+  IFS=";" read -r -a configs <<< "$hook_config"
 
   local hack_terraform_docs
   hack_terraform_docs=$(terraform version | sed -n 1p | grep -c 0.12) || true
--- a/terraform_docs.sh
+++ b/terraform_docs.sh
@@ -140,11 +140,11 @@ function terraform_docs {
     esac
   done
 
-  local path_uniq
-  for path_uniq in $(echo "${paths[*]}" | tr ' ' '\n' | sort -u); do
-    path_uniq="${path_uniq//__REPLACED__SPACE__/ }"
+  local dir_path
+  for dir_path in $(echo "${paths[*]}" | tr ' ' '\n' | sort -u); do
+    dir_path="${dir_path//__REPLACED__SPACE__/ }"
 
-    pushd "$path_uniq" > /dev/null
+    pushd "$dir_path" > /dev/null
 
     #
     # Create file if it not exist and `--create-if-not-exist=true` provided
--- a/terraform_docs.sh
+++ b/terraform_docs.sh
@@ -23,6 +23,7 @@ function common::parse_cmdline {
   # common global arrays.
   # Populated via `common::parse_cmdline` and can be used inside hooks' functions
   declare -g -a ARGS=() FILES=() HOOK_CONFIG=()
+  declare -g -r HOOK_CONFIG_SEPARATOR='%%SEPARATOR%%'
 
   local argv
   argv=$(getopt -o a:,h: --long args:,hook-config: -- "$@") || return
@@ -37,7 +38,7 @@ function common::parse_cmdline {
         ;;
       -h | --hook-config)
         shift
-        HOOK_CONFIG+=("$1;")
+        HOOK_CONFIG+=("${1}${HOOK_CONFIG_SEPARATOR}")
         shift
         ;;
       --)
@@ -56,7 +57,7 @@ function terraform_docs_ {
   local -a -r files=("$@")
 
   # Get hook settings
-  IFS=";" read -r -a configs <<< "$hook_config"
+  IFS="$HOOK_CONFIG_SEPARATOR" read -r -a configs <<< "$hook_config"
 
   local hack_terraform_docs
   hack_terraform_docs=$(terraform version | sed -n 1p | grep -c 0.12) || true
--- a/terraform_docs.sh
+++ b/terraform_docs.sh
@@ -1,35 +1,32 @@
 #!/usr/bin/env bash
 set -eo pipefail
 
-main() {
-  initialize_
-  parse_cmdline_ "$@"
+function main {
+  common::initialize
+  common::parse_cmdline "$@"
   # Support for setting relative PATH to .terraform-docs.yml config.
   ARGS=${ARGS[*]/--config=/--config=$(pwd)\/}
   terraform_docs_ "${HOOK_CONFIG[*]}" "$ARGS" "${FILES[@]}"
 }
 
-initialize_() {
+function common::initialize {
+  local SCRIPT_DIR
   # get directory containing this script
-  local dir
-  local source
-  source="${BASH_SOURCE[0]}"
-  while [[ -L $source ]]; do # resolve $source until the file is no longer a symlink
-    dir="$(cd -P "$(dirname "$source")" > /dev/null && pwd)"
-    source="$(readlink "$source")"
-    # if $source was a relative symlink, we need to resolve it relative to the path where the symlink file was located
-    [[ $source != /* ]] && source="$dir/$source"
-  done
-  _SCRIPT_DIR="$(dirname "$source")"
+  SCRIPT_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
 
   # source getopt function
   # shellcheck source=lib_getopt
-  . "$_SCRIPT_DIR/lib_getopt"
+  . "$SCRIPT_DIR/lib_getopt"
 }
 
-parse_cmdline_() {
-  declare argv
-  argv=$(getopt -o a: --long args:,hook-config: -- "$@") || return
+# common global arrays.
+# Populated in `parse_cmdline` and can used in hooks functions
+declare -a ARGS=()
+declare -a HOOK_CONFIG=()
+declare -a FILES=()
+function common::parse_cmdline {
+  local argv
+  argv=$(getopt -o a:,h: --long args:,hook-config: -- "$@") || return
   eval "set -- $argv"
 
   for argv; do
@@ -39,9 +36,9 @@ parse_cmdline_() {
         ARGS+=("$1")
         shift
         ;;
-      --hook-config)
+      -h | --hook-config)
         shift
-        HOOK_CONFIG+=("$1")
+        HOOK_CONFIG+=("$1;")
         shift
         ;;
       --)
@@ -53,11 +50,14 @@ parse_cmdline_() {
   done
 }
 
-terraform_docs_() {
+function terraform_docs_ {
   local -r hook_config="$1"
   local -r args="$2"
   shift 2
   local -a -r files=("$@")
+
+  # Get hook settings
+  IFS=";" read -r -a configs <<< "$hook_config"
 
   local hack_terraform_docs
   hack_terraform_docs=$(terraform version | sed -n 1p | grep -c 0.12) || true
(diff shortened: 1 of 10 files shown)


terraform_fmt.sh
```diff
@@ -46,44 +48,77 @@ parse_cmdline_() {
...
+  shopt -qo errexit && ERREXIT_IS_SET=true
+  set +e
+  local final_exit_code=0
 
-    (
-      cd "$path_uniq"
-      terraform fmt "${ARGS[@]}"
-    )
+  # run hook for each path
+  for dir_path in $(echo "${dir_paths[*]}" | tr ' ' '\n' | sort -u); do
```
yermulnik, line 77: This will play bad if any `dir_path` has a space in it
```suggestion
  declare -A dir_paths_tmp
  for item in "${dir_paths[@]}"; do dir_paths_tmp["$item"]+=1; done
  dir_paths=("${!dir_paths_tmp[@]}")
  unset dir_paths_tmp
  for dir_path in "${dir_paths[@]}"; do
```

yermulnik: This means we also probably can get rid of `__REPLACED__SPACE__` from wherever it comes from.

terraform_fmt.sh
```diff
@@ -46,44 +48,77 @@ parse_cmdline_() {
...
+  local -r args="$1"
+  local -r dir_path="$2"
+
+  # pass the arguments to terrascan
+  # shellcheck disable=SC2068 # terrascan fails when quoting is used ("$arg" vs $arg)
+  terraform fmt ${args[@]}
+
+  # return exit code to common::per_dir_hook
+  local exit_code=$?
+  return $exit_code
```
yermulnik, line 121: Why do we need to assign `$?` to intermediate var `exit_code`?
Ain't going to work just like `return $?`

terraform_providers_lock.sh
```diff
@@ -2,33 +2,53 @@
...
+function common::colorify {
+  # Colors. Provided as first string to first arg of function.
+  # shellcheck disable=SC2034
+  local -r red="$(tput setaf 1)"
+  # shellcheck disable=SC2034
+  local -r green="$(tput setaf 2)"
+  # shellcheck disable=SC2034
+  local -r yellow="$(tput setaf 3)"
+  # Color reset
+  local -r RESET="$(tput sgr0)"
```
yermulnik, line 20: Why using `tput` (which btw behaves differently on different BSD systems) and no just escape-codes?
```suggestion
  local -r red="\e[0m\e[31m"
  # shellcheck disable=SC2034
  local -r green="\e[0m\e[32m"
  # shellcheck disable=SC2034
  local -r yellow="\e[0m\e[33m"
  # Color reset
  local -r RESET="\e[0m"
```

terraform_providers_lock.sh
```diff
@@ -2,33 +2,53 @@
...
+
+  # Params start #
+  local COLOR="${!1}"
+  local -r TEXT=$2
+  # Params end #
+
+  if [ "$PRE_COMMIT_COLOR" = "never" ]; then
```
yermulnik, line 27: Double square quotes are used most of the time across this file, so to unify approaches used:
```suggestion
  if [[ "$PRE_COMMIT_COLOR" == "never" ]]; then
```

terraform_providers_lock.sh
```diff
@@ -47,42 +72,74 @@ parse_cmdline_() {
...
-# global arrays
-declare -a ARGS
-declare -a FILES
+function per_dir_hook_unique_part {
+  # common logic located in common::per_dir_hook
+  local -r args="$1"
+  local -r dir_path="$2"
+
+  if [[ ! -d ".terraform" ]]; then
+    set +e
```
yermulnik, line 124:
```suggestion
    # preserve errexit status
    shopt -qo errexit && ERREXIT_IS_SET=true
```

terraform_providers_lock.sh
```diff
@@ -47,42 +72,74 @@ parse_cmdline_() {
...
+    init_output=$(terraform init -backend=false 2>&1)
+    init_code=$?
+    set -e
```
yermulnik, line 127:
```suggestion
    # restore errexit if it was set before the block
    [[ $ERREXIT_IS_SET ]] && set -e
```

terraform_providers_lock.sh
```diff
@@ -47,42 +72,74 @@ parse_cmdline_() {
...
+
+    if [[ $init_code != 0 ]]; then
```
yermulnik, line 129: `!=` is a string comparison op
```suggestion
    if [[ $init_code -ne 0 ]]; then
```

terraform_providers_lock.sh
```diff
@@ -47,42 +72,74 @@ parse_cmdline_() {
...
+      common::colorify "red" "Init before validation failed: $dir_path"
+      common::colorify "red" "$init_output"
+      exit 1
```
yermulnik, line 132: Might it worth to return exit code of `terraform init` rather then a somewhat synthetic exit code?
```suggestion
      exit $init_code
```

yermulnik (commented):
Great work Max. I added some comments (and didn't duplicate them for each script where they are relevant for the same code bits).

terraform_fmt.sh
```diff
@@ -46,44 +48,77 @@ parse_cmdline_() {
...
 
-    paths[index]=$(dirname "$file_with_path")
-
+    dir_paths[index]=$(dirname "$file_with_path")
+    # TODO Unique part
     if [[ "$file_with_path" == *".tfvars" ]]; then
       tfvars_files+=("$file_with_path")
     fi
-
+    #? End for unique part
```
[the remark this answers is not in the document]
MaxymVlasov: Unique part comparatively with `common::per_dir_hook` function

terraform_fmt.sh
```diff
@@ -46,44 +48,77 @@ parse_cmdline_() {
...
     ((index += 1))
```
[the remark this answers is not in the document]
MaxymVlasov: If it works the same, then there is no difference. I just reused the most common pattern funded it hooks.

terraform_fmt.sh
MaxymVlasov: I think it's just more explicit and understandable.

terraform_providers_lock.sh
MaxymVlasov: It's set in parent function already, so I just remove them

terraform_docs.sh
```diff
@@ -1,35 +1,32 @@
...
-  declare argv
-  argv=$(getopt -o a: --long args:,hook-config: -- "$@") || return
+# common global arrays.
+# Populated in `parse_cmdline` and can used in hooks functions
+declare -a ARGS=()
+declare -a HOOK_CONFIG=()
+declare -a FILES=()
+function common::parse_cmdline {
+  local argv
+  argv=$(getopt -o a:,h: --long args:,hook-config: -- "$@") || return
```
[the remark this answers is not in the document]
MaxymVlasov: This is the reason why it was chosen - nobody will specify `-h` in `args:` because it is useless, so we free to reserve it for our needs

terraform_docs.sh
```diff
@@ -39,9 +36,9 @@ parse_cmdline_() {
         ARGS+=("$1")
         shift
         ;;
-      --hook-config)
+      -h | --hook-config)
         shift
-        HOOK_CONFIG+=("$1")
+        HOOK_CONFIG+=("$1;")
```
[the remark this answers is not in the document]
MaxymVlasov: I reverted changes in [`f3dd6de` (#310)](https://github.com/antonbabenko/pre-commit-terraform/pull/310/commits/f3dd6deac411ee9144a22752db7a53923685fbce) during

```
/home/vm/.cache/pre-commit/repovz9rff8d/terraform_docs.sh: line 165: : No such file or directory
```


You can test it for yourself:

```yaml
repos:
  rev: f3dd6deac411ee9144a22752db7a53923685fbce
  # rev: v1.62.3
  hooks:
    - id: terraform_docs
      args:
      - --hook-config=--path-to-file=README.md        # Valid UNIX path. I.e. ../TFDOC.md or docs/README.md etc.
```

terraform_fmt.sh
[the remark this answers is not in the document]
MaxymVlasov: Here is reason why `((index += 1))`:

```
/tmp/tmp1bjg3zq6/repoh85kwmsh/terraform_docs.sh: line 110: paths[index]: bad array subscript
```

terraform_fmt.sh
MaxymVlasov: `dir_path` will have no spaces in their path, during

```bash
    file_with_path="${file_with_path// /__REPLACED__SPACE__}"

    dir_paths[index]=$(dirname "$file_with_path")
```

Also, I didn't understand what the code in Suggested change does. 
I will not replace things that looks pretty straight forward with things that need tons of comments

terraform_providers_lock.sh
MaxymVlasov: Hmm, are you sure that `tput` will not provide the same colors for macOS and Linux?
I'm not a fan of escape sequences.

terraform_docs.sh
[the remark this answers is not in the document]
yermulnik: Oops, my bad =( I've made a mistake as I missed that `IFS` may consist of only one single char =(
...so I took another look into the use case of `IFS` and it seems to me that it is needless.
Why did you decide to use custom `IFS`?

terraform_fmt.sh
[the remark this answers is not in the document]
yermulnik: Aha, I see, Bash is tricky sometimes 😠 It seems to treat vars as strings if not defined otherwise. So the fix is simple like this `local -i index=0` (the `-i` attr).

terraform_fmt.sh
yermulnik: > Also, I didn't understand what the code in Suggested change does.

It replaces external deps like `tr` and `sort` with bash builtin functionality.

>  I will not replace things that looks pretty straight forward with things that need tons of comments

Makes sense. It was literally a _suggestion_. I'm not pushing any suggestion I propose. 
If peer engineer says what you said, this means the suggestion should not be accepted!
Thanks for being mindful and straight-thinking 👍🏻

terraform_providers_lock.sh
yermulnik: I don't have MacOS, though on FreeBSD `tput {setaf,sgr0}` doesn't do what it's expected to comparing with Ubuntu, whereas escape codes work as intended on both.

terraform_providers_lock.sh
yermulnik: What I meant is to preserve/restore it every time we set it on-demand.
Imagine for example that we need to run 1st half of the script with `-e` and the other half with `+e` though somewhere in the middle of 2nd half we need to explicitly set `+e` without bearing in mind whether this code gets executed: whether in the 1st half of the script or in the other half.
Pretty over-cautiousness, though is reliable and predictable.

terraform_docs.sh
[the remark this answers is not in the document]
yermulnik: > Why did you decide to use custom `IFS`?

I'm asking because it seems to introduce an extra whitespace in array items.
So the code:
```
  echo "arr => \"${configs[@]}\""
  for item in "${configs[@]}"; do echo "item => \"$item\""; done
```
outputs this:
```
arr => "--path-to-file=README.md  --add-to-existing-file=true  --create-file-if-not-exist=true"
item => "--path-to-file=README.md"
item => " --add-to-existing-file=true"
item => " --create-file-if-not-exist=true"
```
and we can see an extra whitespace a the very beginning of array items starting with 2n item.
Removing `IFS="$HOOK_CONFIG_SEPARATOR"` from `read …` command, eliminates this.

yermulnik (approved):
LGTM. None of still ongoing discussions seem to be of a crucial matter so can be resolved (if needed) via another PR(s).

terraform_docs.sh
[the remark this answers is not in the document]
MaxymVlasov: It can be:

```
--path-to-file README.md  --add-to-existing-file true  --create-file-if-not-exist true"
```

It needs to support cases when space is part of the string:

```yaml
  hooks:
   - id: terraform_docs
      args:
      - --hook-config=--path-to-file README.md
      - --hook-config=--add-to-existing-file true
      - --hook-config=--create-file-if-not-exist true
    - id: infracost_breakdown
      args:
        - --args=--path=./environment/qa
        - --hook-config=.totalHourlyCost > 0.1
        - --hook-config=.totalHourlyCost|tonumber > 1
        - -h .projects[].diff.totalMonthlyCost|tonumber!=10000
        - -h .currency   >=   "USD"
```

terraform_providers_lock.sh
MaxymVlasov: Well, if we support FreeBSD, we totally should start supporting Windows :D 

OK, I will change it to escape codes

yermulnik: This is not literally about supporting different OS'es, but about compatibility and portability across different *NIX implementations, including e.g. difference in POSIX and GNU tools and their derivatives.

terraform_docs.sh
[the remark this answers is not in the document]
yermulnik: Argh, got you now.

antonbabenko (approved):

PR #312, opened 2022-01-03
fix(terragrunt_validate): Only run terragrunt_validate in valid direc…

### LINKED ISSUE #191: terragrunt_validate - fails for folders without terragrunt.hcl

thetimbecker:
…tories (#191)

The command fails if there is no terragrunt.hcl file in the directory.

- [x] This PR fixes a bug.

Fixes #191 

Used this `files` pattern locally in my `.pre-commit-config.yaml` file and verified it no longer fails and still runs in desired directories. Using `(/terragrunt\.hcl)$` instead of `(terragrunt\.hcl)$` so files with a `terragrunt` suffix aren't matched, like `root_terragrunt.hcl`.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -82,7 +82,7 @@
   description: Validates all Terragrunt configuration files.
   entry: terragrunt_validate.sh
   language: script
-  files: (\.hcl)$
+  files: (/terragrunt\.hcl)$
   exclude: \.terraform\/.*$
 
 - id: terraform_tfsec


.pre-commit-hooks.yaml
```diff
@@ -82,7 +82,7 @@
   description: Validates all Terragrunt configuration files.
   entry: terragrunt_validate.sh
   language: script
-  files: (\.hcl)$
+  files: (/terragrunt\.hcl)$
```
yermulnik, line 85: Irrespective to the goal of the PR: parentheses are redundant here (as there's no grouping), forward slash needs to be escaped (as it is a special char in regex), and possibly files w/o path need to be taken into account (?):
```suggestion
  files: (^|\/)terragrunt\.hcl$
```

MaxymVlasov:
https://github.com/antonbabenko/pre-commit-terraform/issues/191#issuecomment-1004316276

Issue #256, opened 2021-10-22
Add hook for autoformat of terraform code snippets embedded in the docs


antonbabenko:
There is a utility called [`terrafmt`](https://github.com/katbyte/terrafmt) that can be used to check and fix the formating of Terraform snippets embedded into usage sections in README.

### How could pre-commit-terraform help solve your problem?

It would be great to have native support as a pre-commit hook.

rahulmlokurte:
@MaxymVlasov @antonbabenko: Can I pick the `terrafmt` hook? I have tried it in my system. Looks like, `terrafmt` does not extract terraform files and does not do inline code formatting from the README.md file if, we start the Three backquotes after 4 spaces in the README.md file. It only works, if we start the Three backquotes at column 0.  

Ex:
1. Works: 

![image](https://user-images.githubusercontent.com/81438015/147782611-126f337a-2915-46d9-af45-73ecac790e0f.png)

```
terrafmt blocks README.md

####### B1 @ #587
resource "aws_lambda_function" "pass" {
  function_name = "test-env"
  role          = ""
  runtime       = "python3.8"

  environment {
    variables = {
      AWS_DEFAULT_REGION = "us-west-2"
    }
  }
}
```


2. Does not work

![image](https://user-images.githubusercontent.com/81438015/147782471-6a7626a6-2fae-4936-b131-06aa3d631683.png)

```
terrafmt blocks README.md

```

The README.md in the repository has all the inline code starting at 4 spaces. For ex: `Hooks usage notes and examples` , apart from `checkov`, all other sections defines the inline code starting at 4 spaces.

antonbabenko:
Sounds like this is the problem in `terrafmt` because markdown format allows having an offset.

PR #313, opened 2022-01-04
feat: Add terrafmt  hook

rahulmlokurte:

- [ ] This PR fixes a bug.

Adding Terrafmt to format terraform embedded in the document. Referencing [#256](https://github.com/antonbabenko/pre-commit-terraform/issues/256)

1. Install terrafmt
2. Add to .pre-commit-config.yaml:

```yaml
- repo: https://github.com/rahulmlokurte/pre-commit-terraform
  rev: 90bd2a5727d2bbd5d5a27ed5390073f11c359a0c
  hooks:
    - id: terrafmt
      args:
        - --args=fmt
 ```
3. Run

```sh
git add -A
pre-commit run
```

Example


README.md
```diff
@@ -568,6 +570,24 @@ Example:
 3. Use `--skip-rules="ruleID1,ruleID2"` parameter to skip one or more rules globally while scanning (e.g.: `--args=--skip-rules="ruleID1,ruleID2"`).
 4. Use the syntax `#ts:skip=RuleID optional_comment` inside a resource to skip the rule for that resource.
 
+### terrafmt
+
+1. `terrafmt` supports custom arguments so you can pass [supported flags](https://github.com/katbyte/terrafmt). Eg:
```
yermulnik, line 575: The URL should point to a respective section in README: `https://github.com/katbyte/terrafmt#usage`

README.md
```diff
@@ -568,6 +570,24 @@ Example:
...
+  role          = ""
+  runtime       = "python3.8"
+
+  environment {
+    variables = {
+      AWS_DEFAULT_REGION = "us-west-2"
+    }
+  }
+}
+```
```
yermulnik, line 589: This block should comprise an example of how to pass custom args to the hook, rather than an example of some irrelevant AWS Lambda resource definition.

terrafmt.sh
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bash
```
yermulnik, line 1: It might worth to pick up what @MaxymVlasov works on in https://github.com/antonbabenko/pre-commit-terraform/pull/310 to modernize and standardize common shell coding conventions and approaches across hook scripts in this project.

terrafmt.sh
```diff
@@ -0,0 +1,59 @@
...
+        ;;
+    esac
+  done
+}
+
+terrafmt_() {
+  find . | grep -E "README.md" | sort | while read -r f; do
+    echo "Processing $f"
+    echo "terrafmt ${ARGS[*]} $f"
+    terrafmt "${ARGS[@]}" $f
```
yermulnik, line 52: To avoid unexpectedness with e.g. files with whitespaces in name, one should enclose args into double-quotes:
```suggestion
    terrafmt "${ARGS[@]}" "$f"
```

terrafmt.sh
```diff
@@ -0,0 +1,59 @@
...
+        break
+        ;;
+    esac
+  done
+}
+
+terrafmt_() {
+  find . | grep -E "README.md" | sort | while read -r f; do
+    echo "Processing $f"
+    echo "terrafmt ${ARGS[*]} $f"
```
yermulnik, line 51: I think it's superfluous to let user know what's happening **twice**. I'd rather put something like this, which should be quite enough for the user to have a clue of what's happening at this particular moment in time:
```suggestion
    echo "terrafmt: $f"
```

terrafmt.sh
```diff
@@ -0,0 +1,59 @@
...
+      --)
+        shift
+        break
+        ;;
+    esac
+  done
+}
+
+terrafmt_() {
+  find . | grep -E "README.md" | sort | while read -r f; do
```
yermulnik, line 49: 1. `find` has an option to lookup files with particular names, hence `grep` is extraneous and redundant here (as a side note, `-E` means `grep` should interpret pattern as an extended regex, which is redundant in this code construction).
2. Hook config is configured in this PR to pass **all** markdown files to the hook (`files: \.md$`), so why would you need to pick only `README.md` files then?

yermulnik (changes requested):

.pre-commit-config.yaml
```diff
@@ -33,3 +33,10 @@ repos:
   hooks:
     - id: shfmt
       args: ['-l', '-i', '2', '-ci', '-sr', '-w']
+
+- repo: https://github.com/rahulmlokurte/pre-commit-terraform
+  rev: 90bd2a5727d2bbd5d5a27ed5390073f11c359a0c
+  hooks:
+    - id: terrafmt
+      args:
+        - --args=fmt
```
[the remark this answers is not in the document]
yermulnik: Also I think it's worth to be more restrictive on what this hook should do. We basically don't really want to allow users run e.g. `terrafmt help` or `terrafmt blocks`, but rather allow users to pass custom `terrafmt` flags to commands like `fmt` and `diff`.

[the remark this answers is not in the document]
MaxymVlasov: This repo has no terraform code at all, so that check here is useless.

At the same time, it can be used as good usage example in README. So, please move

```yaml
- id: terrafmt
  args:
    - --args=fmt
```

to `Hooks usage notes and examples` section in README with related comments, if they needed

README.md
```diff
@@ -62,6 +62,7 @@ If you are using `pre-commit-terraform` already or want to support its developme
 * [`TFSec`](https://github.com/liamg/tfsec) required for `terraform_tfsec` hook.
 * [`infracost`](https://github.com/infracost/infracost) required for `infracost_breakdown` hook.
 * [`jq`](https://github.com/stedolan/jq) required for `infracost_breakdown` hook.
+* [`terrafmt`](https://github.com/katbyte/terrafmt) required for `terraform_fmt` hook.
```
MaxymVlasov, line 65: Please add needed installation instructions both for macOS and Ubuntu in the sections below.

README.md
rahulmlokurte: Added the example on how to pass custom args to the hook

PR #314, opened 2022-01-05
[PR to discuss the concept] Improve performance for hooks when all files are scanned

### LINKED ISSUE #309: Improve perfomance during `pre-commit --all` (`-a`) run
Some hooks recursively check all files in provided dir.

So performance degradation exists only in the `pre-commit run --all` situation, because it will provide all existing repo files to hook:

https://github.com/antonbabenko/pre-commit-terraform/blob/e6ffbcd3def77c7caa471d008daefa13f9095127/terrascan.sh#L15-L19

Then, unique paths are found and run `terrascan` for each repo folder:

https://github.com/antonbabenko/pre-commit-terraform/blob/e6ffbcd3def77c7caa471d008daefa13f9095127/terrascan.sh#L29-L30

It works literally how it should work: checks only diffs.

So, good to know when the `--all` (`-a`) argument passed to `pre-commit` and just run `terrascan -d GIT_REPO_ROOT`, not all-all dirs.

---

## Useful info

* [How to run hook performance test](https://github.com/antonbabenko/pre-commit-terraform/blob/master/.github/CONTRIBUTING.md#run-hook-performance-test)
* `pre-commit` automatically parallel checks to exiting cores, so you need to run tests on a repo that has at least 2x more tf-dirs than CPU cores you have. If you have not so big repo - just copy-paste code-structure a few times, and you'll get needed.

* Create solution as function, that can be called from `terrascan_()` hook function and depends on the result, in `if` run other flow and stop execution with `exit 0`

---

- [ ] `checkov` run checks recursively. need `-d GIT_REPO_POOT_PATH`

- [ ] `tfsec` runs recursively without any additional args.

- [ ] `terragrunt_fmt`  ( [`terragrunt hclfmt`](https://github.com/gruntwork-io/terragrunt/blob/33b07484c81dd3302c579c28d6a03f915ce3eac2/docs/_docs/04_reference/cli-options.md#hclfmt)) runs recursively without any additional args.

- [ ] `terrascan` runs recursively without any additional args.

- [x] `terragrunt_validate` have option [`terragrunt run-all validate`](https://github.com/gruntwork-io/terragrunt/blob/33b07484c81dd3302c579c28d6a03f915ce3eac2/docs/_docs/04_reference/cli-options.md#validate-all-deprecated-use-run-all) that run checks recursively.   #327

---

Founded in https://github.com/antonbabenko/pre-commit-terraform/pull/305

MaxymVlasov:

- [ ] This PR fixes a bug.

Close #309

```bash
git clone git@github.com:carlosbustillordguez/pre-commit-terraform.git /tmp/309
cd /tmp/309
git checkout chore/refactoring

cd YOUR_TEST_REPO

pre-commit try-repo -a /tmp/309 terrascan
```

### DIFF
--- a/terrascan.sh
+++ b/terrascan.sh
@@ -1,6 +1,10 @@
 #!/usr/bin/env bash
 set -eo pipefail
 
+# globals variables
+# hook ID, see `- id` for details in .pre-commit-hooks.yaml file
+HOOK_ID="terrascan"
+
 function main {
   common::initialize
   common::parse_cmdline "$@"
@@ -52,9 +56,8 @@ function common::is_hook_run_on_whole_repo {
   local sorted_string
   local git_ls_files
   local included_files
-  local HOOK_ID="terrascan"
-
   local SCRIPT_DIR
+
   # get directory containing this script
   SCRIPT_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
 
--- a/terrascan.sh
+++ b/terrascan.sh
@@ -47,10 +47,50 @@ function common::parse_cmdline {
   done
 }
 
+function common::is_hook_run_on_whole_repo {
+  local -a -r files=("$@")
+  local sorted_string
+  local git_ls_files
+  local included_files
+  local HOOK_ID="terrascan"
+
+  local SCRIPT_DIR
+  # get directory containing this script
+  SCRIPT_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
+
+  # get included and excluded files from .pre-commit-hooks.yaml file
+  included_files=$(sed -n "/^- id: $HOOK_ID$/,/^$/p" "$SCRIPT_DIR/.pre-commit-hooks.yaml" | awk '$1 == "files:" {print $2; exit}')
+  excluded_files=$(sed -n "/^- id: $HOOK_ID$/,/^$/p" "$SCRIPT_DIR/.pre-commit-hooks.yaml" | awk '$1 == "exclude:" {print $2; exit}')
+
+  # sorted string with the files passed to the hook by pre-commit
+  sorted_string=$(printf '%s\n' "${files[@]}" | sort | tr '\n' ' ')
+
+  # git ls-files sorted string
+  if [[ "$excluded_files" != "" ]]; then
+    git_ls_files=$(git ls-files | sort | grep -v "$excluded_files" | grep -e "$included_files" | tr '\n' ' ')
+  else
+    git_ls_files=$(git ls-files | sort | grep -e "$included_files" | tr '\n' ' ')
+  fi
+
+  # This is a workaround to improve performance when all files are passed
+  # See: https://github.com/antonbabenko/pre-commit-terraform/issues/309
+  if [[ "$sorted_string" == "$git_ls_files" ]]; then
+    echo "true"
+  else
+    echo "false"
+  fi
+}
+
 function common::per_dir_hook {
   local -r args="$1"
   shift 1
   local -a -r files=("$@")
+
+  # check is (optional) function defined
+  if [[ $(type -t run_hook_on_whole_repo) == function ]] && [[ $(common::is_hook_run_on_whole_repo "${files[@]}") == "true" ]]; then
+    run_hook_on_whole_repo "$args"
+    exit 0
+  fi
 
   # consume modified files passed from pre-commit so that
   # hook runs against only those relevant directories
@@ -102,6 +142,14 @@ function per_dir_hook_unique_part {
   # return exit code to common::per_dir_hook
   local exit_code=$?
   return $exit_code
+}
+
+function run_hook_on_whole_repo {
+  local -r args="$1"
+
+  # pass the arguments to hook
+  # shellcheck disable=SC2068 # hook fails when quoting is used ("$arg[@]")
+  terrascan scan -i terraform ${args[@]}
 }
 
 [ "${BASH_SOURCE[0]}" != "$0" ] || main "$@"


terrascan.sh
```diff
@@ -1,6 +1,10 @@
 #!/usr/bin/env bash
 set -eo pipefail
 
+# globals variables
+# hook ID, see `- id` for details in .pre-commit-hooks.yaml file
+HOOK_ID="terrascan"
```
MaxymVlasov, line 6:
```suggestion
declare -r HOOK_ID='terrascan'
```

terrascan.sh
```diff
@@ -47,11 +51,50 @@ function common::parse_cmdline {
...
+  fi
+
+  # This is a workaround to improve performance when all files are passed
+  # See: https://github.com/antonbabenko/pre-commit-terraform/issues/309
+  if [[ "$sorted_string" == "$git_ls_files" ]]; then
+    echo "true"
+  else
+    echo "false"
+  fi
+}
```
MaxymVlasov, line 85:
```suggestion
# This is a workaround to improve performance when all files are passed
# See: https://github.com/antonbabenko/pre-commit-terraform/issues/309
function common::is_hook_run_on_whole_repo {
  local -a -r files=("$@")

  # get directory containing this script
  local -r SCRIPT_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
  # get included and excluded files from .pre-commit-hooks.yaml file
  local -r included_files=$(sed -n "/^- id: $HOOK_ID$/,/^$/p" "$SCRIPT_DIR/.pre-commit-hooks.yaml" | awk '$1 == "files:" {print $2; exit}')
  local -r excluded_files=$(sed -n "/^- id: $HOOK_ID$/,/^$/p" "$SCRIPT_DIR/.pre-commit-hooks.yaml" | awk '$1 == "exclude:" {print $2; exit}')
  # sorted string with the files passed to the hook by pre-commit
  local -r files_to_check=$(printf '%s\n' "${files[@]}" | sort | tr '\n' ' ')

  # git ls-files sorted string
  local all_files_that_can_be_checked
  if [ -z "$excluded_files" ]; then
    all_files_that_can_be_checked=$(git ls-files | sort | grep -e "$included_files" | tr '\n' ' ')
  else
    all_files_that_can_be_checked=$(git ls-files | sort | grep -v "$excluded_files" | grep -e "$included_files" | tr '\n' ' ')
  fi

  if [ "$files_to_check" == "$all_files_that_can_be_checked" ]; then
    echo true
  else
    echo false
  fi
}
```

terrascan.sh
```diff
@@ -47,11 +51,50 @@ function common::parse_cmdline {
...
+
 function common::per_dir_hook {
   local -r args="$1"
   shift 1
   local -a -r files=("$@")
 
+  # check is (optional) function defined
+  if [[ $(type -t run_hook_on_whole_repo) == function ]] && [[ $(common::is_hook_run_on_whole_repo "${files[@]}") == "true" ]]; then
```
MaxymVlasov, line 93:
```suggestion
  if [ "$(type -t run_hook_on_whole_repo)" == function ] &&
    # check is hook run via `pre-commit run --all`
    [ "$(common::is_hook_run_on_whole_repo "${files[@]}")" == true ]; then

```

terrascan.sh
```diff
@@ -47,11 +51,50 @@ function common::parse_cmdline {
   done
 }
 
+function common::is_hook_run_on_whole_repo {
+  local -a -r files=("$@")
+  local sorted_string
+  local git_ls_files
+  local included_files
```
yermulnik, line 58:
```suggestion
  local hook_config_block
  local included_files
  local excluded_files
```

terrascan.sh
```diff
@@ -47,11 +51,50 @@ function common::parse_cmdline {
...
+  local SCRIPT_DIR
+
+  # get directory containing this script
+  SCRIPT_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
+
+  # get included and excluded files from .pre-commit-hooks.yaml file
+  included_files=$(sed -n "/^- id: $HOOK_ID$/,/^$/p" "$SCRIPT_DIR/.pre-commit-hooks.yaml" | awk '$1 == "files:" {print $2; exit}')
+  excluded_files=$(sed -n "/^- id: $HOOK_ID$/,/^$/p" "$SCRIPT_DIR/.pre-commit-hooks.yaml" | awk '$1 == "exclude:" {print $2; exit}')
```
yermulnik, line 66: Just for the sake of reducing similar calls:
```suggestion
  # get included and excluded files from .pre-commit-hooks.yaml file
  hook_config_block=$(sed -n "/^- id: $HOOK_ID$/,/^$/p" "$SCRIPT_DIR/.pre-commit-hooks.yaml")
  included_files=$(awk '$1 == "files:" {print $2; exit}' <<<"$hook_config_block")
  excluded_files=$(awk '$1 == "exclude:" {print $2; exit}' <<<"$hook_config_block")
```

terrascan.sh
```diff
@@ -47,11 +51,50 @@ function common::parse_cmdline {
...
+
+  # sorted string with the files passed to the hook by pre-commit
+  sorted_string=$(printf '%s\n' "${files[@]}" | sort | tr '\n' ' ')
+
+  # git ls-files sorted string
+  if [[ "$excluded_files" != "" ]]; then
+    git_ls_files=$(git ls-files | sort | grep -v "$excluded_files" | grep -e "$included_files" | tr '\n' ' ')
+  else
+    git_ls_files=$(git ls-files | sort | grep -e "$included_files" | tr '\n' ' ')
```
yermulnik, line 75: `-e` arg to `grep` looks to be needless here, at least because this line's counterpart at line 73 doesn't have it.

MaxymVlasov:
@carlosbustillordguez I have no idea why I have no permission to apply review suggestions or push commit to fork branch, so I just commented some changes here.

About
>As commented above I had to use the SCRIPT_DIR variable to parse the .pre-commit-hooks.yaml file. I think that variable should be considered as global to use when is needed.
>
>Regarding global variables, should we pass them as arguments to the needed function or consume them directly inside the function? I mean for HOOK_ID and SCRIPT_DIR (if is defined as global)?

Let's not touch `SCRIPT_DIR` for now and leave it as is.

Regarding `HOOK_ID`. let's make it `local -r` in `function main` and populate it as a function argument.

terrascan.sh
```diff
@@ -47,11 +51,50 @@ function common::parse_cmdline {
...
+  # get included and excluded files from .pre-commit-hooks.yaml file
+  included_files=$(sed -n "/^- id: $HOOK_ID$/,/^$/p" "$SCRIPT_DIR/.pre-commit-hooks.yaml" | awk '$1 == "files:" {print $2; exit}')
+  excluded_files=$(sed -n "/^- id: $HOOK_ID$/,/^$/p" "$SCRIPT_DIR/.pre-commit-hooks.yaml" | awk '$1 == "exclude:" {print $2; exit}')
+
+  # sorted string with the files passed to the hook by pre-commit
+  sorted_string=$(printf '%s\n' "${files[@]}" | sort | tr '\n' ' ')
+
+  # git ls-files sorted string
+  if [[ "$excluded_files" != "" ]]; then
+    git_ls_files=$(git ls-files | sort | grep -v "$excluded_files" | grep -e "$included_files" | tr '\n' ' ')
```
yermulnik, line 73: I'm not quite sure how `pre-commit` interprets `files` and `exclude` parameters passed altogether but from what I lean to think it's something like "include `files` AND exclude `exclude`" which means that we need to reproduce this particular order for the sake of sameness:
```suggestion
    git_ls_files=$(git ls-files | sort | grep "$included_files" | grep -v "$excluded_files" | tr '\n' ' ')
```
What do you think?

terrascan.sh
```diff
@@ -47,11 +51,50 @@ function common::parse_cmdline {
...
+    git_ls_files=$(git ls-files | sort | grep -e "$included_files" | tr '\n' ' ')
+  fi
+
+  # This is a workaround to improve performance when all files are passed
+  # See: https://github.com/antonbabenko/pre-commit-terraform/issues/309
+  if [[ "$sorted_string" == "$git_ls_files" ]]; then
+    echo "true"
+  else
+    echo "false"
+  fi
```
yermulnik, line 84: Out of interest: why use `echo "{true,false}"` rather than `return {0,1}`? Does `return` break some logic?

terrascan.sh
yermulnik: The suggestion covers too big part of the code hence it's not quite obvious what's the change that is being suggested, hence the question: is the only difference the way `local` command being used (in the very beginning of the function definition vs right beside the variable)?

terrascan.sh
yermulnik: This is not essential though from my point of view `[[ condition && condition ]]` is more friendly and flexible as opposed to `[ condition ] && [ condition ]`. The only con is that `[[` is Bash-specific (and we explicitly use Bash).

terrascan.sh
MaxymVlasov: It have.

terrascan.sh
MaxymVlasov: `excluded_files` can be empty. `grep -v ""` just excludes all

yermulnik:
> I have no idea why I have no permission to apply review suggestions or push commit to fork branch

That's because you have no permission to write to Carlos's repo where the code, which is being proposed to be pulled in into this repo, is located 😏

MaxymVlasov:
>That's because you have no permission to write to Carlos's repo where the code, which is being proposed to be pulled in to this repo, is located 😏

Haha. So here I broke Github?
https://github.com/antonbabenko/pre-commit-terraform/pull/305

terrascan.sh
yermulnik: Apologies for the confusion, I mean the `grep -v "$excluded_files"` as counterpart.
What I mean we should use `grep -v -e` and `grep -e` OR `grep -v` and `grep` for the sake of sameness, but not mix different approaches (the main (and only?) purpose of `-e` parameter is to protect from patterns which begin with a dash — so it might worth to use it everywhere rather than on a random basis 🤔).

terrascan.sh
MaxymVlasov: Out of scope: bash break any logic. :) 

`[ $(common::is_hook_run_on_whole_repo "${files[@]}") ]` should be fine as well

terrascan.sh
yermulnik: Well, right above this line there a respective condition `if [[ "$excluded_files" != "" ]]; then` to protect from empty inverse `grep`.
My suggestion was to obey the ordering of inclusion and exclusion and have the order be the same as `pre-commit`'s logic has (I may be wrong assuming that `pre-commit` applies inclusion before exclusion though).

terrascan.sh
MaxymVlasov: it changes style and strict logic. Check diff locally, if you'd like

yermulnik:
> > That's because you have no permission to write to Carlos's repo where the code, which is being proposed to be pulled in to this repo, is located smirk
> 
> Haha. So in #305 I broke Github?

That's interesting. Didn't Carlos grant you write permission in his forked repo back then?

terrascan.sh
MaxymVlasov: `[[` added not needed interpretation step. I would like to have a code dummy and strict as possible, which helps avoid errors.

At the same time, [google style guide](https://google.github.io/styleguide/shellguide.html#test----and---) says that `[[ … ]] reduces errors as no pathname expansion or word splitting takes place between [[ and ]]`. 
So now I am not sure what is better.

terrascan.sh
yermulnik: Yep, that's what I implied: having a condition based on _true_ `true/false` is sort of much more straightforward in NIX'ish world as opposite to relying on a string comparison produced by a command/function/whatever where the consumer expects only success or failure 😃

yermulnik:
> > > That's because you have no permission to write to Carlos's repo where the code, which is being proposed to be pulled in to this repo, is located smirk
> > 
> > 
> > Haha. So in #305 I broke Github?
> 
> That's interesting. Didn't Carlos grant you write permission in his forked repo back then?

Oh, here it is: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

terrascan.sh
MaxymVlasov: Ah, my bad. Nevermind

terrascan.sh
yermulnik: How the `local` is used got changed, `readonly` attr added to local vars, some vars got their name change. For anything else — I'd better trust you 😂

By the way I liked the approach of putting definition of `local` var types right as the first thing in function definition more. It's sort of a best practice in development most of the time (irrespective to a way flexible Bash rules).

terrascan.sh
yermulnik: The `[` is the alias for old `test` command (despite `test` being builtin in case of Bash it's still POSIX complaint), hence lacks many modern features (comparing with `[[`) and is used for POSIX compatibility (which is not the case as we explicitly use Bash via shebang).

yermulnik: Btw w.r.t. google style guide: https://google.github.io/styleguide/shellguide.html#testing-strings would be unhappy with most of conditionals in this PR (testing whether the string is empty, comparing strings, etc).
In any case that's a good read to look through every so often (despite I'm not very comfortable with some statements Google incorporated and which are a bit out of POSIX standards and common traditional approaches 😛). Thanks for linking it! 👍🏻

terrascan.sh
MaxymVlasov: It's a best practice mostly only for Pascal and languages influenced/build on top of it.

First of all, you can't set up var as read-only if you do not combine var definition and value assignment.
Second thing, when you introduce var as later as possible, you decrease the count of lines where this var can be changed/reassigned that simplifying debugging.

terrascan.sh
antonbabenko: Let's follow one standard - the one by Google seems to be rather accurate. If necessary, we should rewrite some conditions and configure `shellcheck` (or other tools) to check accordingly to that standard.

Anyway, one standard is better than no standards :)

terrascan.sh
yermulnik: > you can't set up var as read-only if you do not combine var definition and value assignment.

Gotcha. I'd expect this to be a comment for a suggested change rather then guessing why the change was suggested 😏 Although I'm a bit hesitant about the need to make most (if not all) local vars readonly.

> … you decrease the count of lines …

Sort of questionable reason imho.

> It's a best practice mostly only for Pascal and languages influenced/build on top of it.

Quote debatable statement, though is a way out of scope.
What I mean is more of let's have one common convention and code style.

terrascan.sh
carlosbustillordguez: Thanks, @yermulnik good catch!

terrascan.sh
carlosbustillordguez: > My suggestion was to obey the ordering of inclusion and exclusion and have the order be the same as pre-commit's logic has (I may be wrong assuming that pre-commit applies inclusion before exclusion though).

@yermulnik I was reading the [pre-commit documentation](https://pre-commit.com/#:~:text=of%20files%20to%20run%20on.-,exclude,(optional%3A%20default%20%5E%24)%20exclude%20files%20that%20were%20matched%20by%20files.,-types) and I understand that `pre-commit` first include the `files` and then exclude files that were matched by `files`. You are right!!

terrascan.sh
carlosbustillordguez: I used `echo "{true,false}"` rather than `return {0,1}` because when I wrote the code I tested all possible combinations for `[[]]` and `[]`, but trying again I found a solution 😃. I think the issue here is `if` is not able to detect the function return when is used with `$()`.

Please, consider the following script where all expected output should be "is true". Noted that the lastest condition works with `return {0,1}`:

```bash
#!/bin/bash

## All case must be "is true"

function fn1() {
    local arg="$1"

    if [[ "$arg" == "foo" ]]; then
        echo "true"
    else
        echo "false"
    fi
}

function fn2() {
    local arg="$1"

    if [[ "$arg" == "foo" ]]; then
        return 0
    else
        return 1
    fi
}

echo "-->> expected 'is true'"
if [[ $(fn1 "foo") == "true" ]]; then
    echo "is true"
else
    echo "is false"
fi

echo "-->> expected 'is true'"
if [[ $(fn2 "foo") ]]; then
    echo "is true"
else
    echo "is false"
fi

echo "-->> expected 'is true'"
if [ "$(fn2 "foo")" ]; then
    echo "is true"
else
    echo "is false"
fi

echo "-->> expected 'is true'"
if [ $(fn2 "foo") ]; then
    echo "is true"
else
    echo "is false"
fi

echo "-->> expected 'is true' -- using 'if fn2 \"foo\"; then ...'"
if fn2 "foo"; then
    echo "is true"
else
    echo "is false"
fi
```

Outputs:

```bash
-->> expected 'is true'
is true
-->> expected 'is true'
is false
-->> expected 'is true'
is false
-->> expected 'is true'
is false
-->> expected 'is true' -- using 'if fn2 "foo"; then ...'
is true
```

carlosbustillordguez:
> > > > That's because you have no permission to write to Carlos's repo where the code, which is being proposed to be pulled in to this repo, is located smirk
> > > 
> > > 
> > > Haha. So in #305 I broke Github?
> > 
> > 
> > That's interesting. Didn't Carlos grant you write permission in his forked repo back then?
> 
> Oh, here it is: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork (explicit permission from PR creator is required).

@MaxymVlasov @yermulnik sorry for the issues regarding my repo permissions. I sent invitations to both, soo feel free to commit the suggestions.

PR #320, opened 2022-01-11
chore: Specify what we exactly mean

MaxymVlasov:
Last time new contributors provide useless info about testing. Let's make PR description more strict.


antonbabenko:
This PR is included in version 1.63.0 :tada:

PR #319, opened 2022-01-11
feat: Full revamp of Dockerfile

balihb:
Put an `x` into the box if that apply:

- [ ] This PR introduces breaking change.
- [ ] This PR fixes a bug.
- [ ] This PR adds new functionality.
- [x] This PR enhances existing functionality.

### Description of your changes

Rewritten and generalized the install steps in Dockerfile.

- hadolint
- dive
- container-structure-test
- install steps now easier to write (syntax highlight) and debug
- install steps can be checked separately with ShellCheck
- arm Docker image
- some shellcheck fix

### How has this code been tested

Lots of local Docker build :D

PR #322, opened 2022-01-11
chore: Add hadolint check for Dockerfiles

MaxymVlasov:

- [ ] This PR fixes a bug.

This PR is partly a subset of #319 originally made by @balihb, partly inspired by #319 PR.
It includes:
* [`hadolint`](https://github.com/hadolint/hadolint) check for Dockerfiles
* Fix Dockefiles issues
* Fix `infracost_breakdown` shellcheck issue. (other introduced fix not works)
* Fix performance test docker image, forgotten during refactoring a few months ago.

### How can we test changes

```yaml
repos:
  rev: 7e34de024972344787a1704e5952f812148aa33b
  hooks:
    - id: infracost_breakdown
      args:
        - --args=--path=./environment/qa
        - --hook-config='.totalHourlyCost >= 0.1'
        - --hook-config=".totalHourlyCost|tonumber > 1"
        - -h ".projects[].diff.totalMonthlyCost|tonumber!=10000"
        - -h [.projects[].diff.totalMonthlyCost | select (.!=null) | tonumber] | add > 1000
      verbose: true # Always show costs
```

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -72,7 +72,7 @@ sudo apt install -y datamash
 ### Run via Docker
 
 ```bash
-# Build `pre-commit` image
+# Build `pre-commit-terraform` image
 docker build -t pre-commit-terraform --build-arg INSTALL_ALL=true .
 # Build test image
 docker build -t pre-commit-tests tests/
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -73,7 +73,7 @@ sudo apt install -y datamash
 
 ```bash
 # Build `pre-commit` image
-docker build -t pre-commit --build-arg INSTALL_ALL=true .
+docker build -t pre-commit-terraform --build-arg INSTALL_ALL=true .
 # Build test image
 docker build -t pre-commit-tests tests/
 # Run
--- a/.github/workflows/pre-commit.yaml
+++ b/.github/workflows/pre-commit.yaml
@@ -25,19 +25,32 @@ jobs:
     - name: Install shellcheck
       run: |
         sudo apt update && sudo apt install shellcheck
+
+    - name: Install hadolint
+      run: |
+        curl -L "$(curl -s https://api.github.com/repos/hadolint/hadolint/releases/latest | grep -o -E -m 1 "https://.+?/hadolint-Linux-x86_64")" > hadolint \
+        && chmod +x hadolint && sudo mv hadolint /usr/bin/
     # Need to success pre-commit fix push
     - uses: actions/checkout@v2
       with:
         fetch-depth: 0
         ref: ${{ github.event.pull_request.head.sha }}
-
+    # Skip terraform_tflint which interferes to commit pre-commit auto-fixes
     - uses: actions/setup-python@v2
       with:
         python-version: '3.9'
     - name: Execute pre-commit
       uses: pre-commit/action@v2.0.0
       env:
-        SKIP: no-commit-to-branch
+        SKIP: no-commit-to-branch,hadolint
       with:
         token: ${{ secrets.GITHUB_TOKEN }}
+        extra_args: --color=always --show-diff-on-failure --files ${{ steps.file_changes.outputs.files }}
+    # Run only skipped checks
+    - name: Execute pre-commit check that have no auto-fixes
+      if: always()
+      uses: pre-commit/action@v2.0.0
+      env:
+        SKIP: check-added-large-files,check-merge-conflict,check-vcs-permalinks,forbid-new-submodules,no-commit-to-branch,end-of-file-fixer,trailing-whitespace,check-yaml,check-merge-conflict,check-executables-have-shebangs,check-case-conflict,mixed-line-ending,detect-aws-credentials,detect-private-key,shfmt,shellcheck
+      with:
         extra_args: --color=always --show-diff-on-failure --files ${{ steps.file_changes.outputs.files }}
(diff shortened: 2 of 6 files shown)


.github/CONTRIBUTING.md
```diff
@@ -73,7 +73,7 @@ sudo apt install -y datamash
 
 ```bash
 # Build `pre-commit` image
-docker build -t pre-commit --build-arg INSTALL_ALL=true .
+docker build -t pre-commit-terraform --build-arg INSTALL_ALL=true .
```
yermulnik, line 76: Probably the comment above needs to be updated respectively as well, does it?

.github/workflows/pre-commit.yaml
```diff
@@ -25,19 +25,32 @@ jobs:
     - name: Install shellcheck
       run: |
         sudo apt update && sudo apt install shellcheck
+
+    - name: Install hadolint
+      run: |
+        curl -L "$(curl -s https://api.github.com/repos/hadolint/hadolint/releases/latest | grep -o -E -m 1 "https://.+?/hadolint-Linux-x86_64")" > hadolint \
+        && chmod +x hadolint && sudo mv hadolint /usr/bin/
```
yermulnik, line 32: Not essential but makes use of single command
```suggestion
        && sudo install -m 0755 -o root -g root hadolint /usr/bin/
```

.github/workflows/pre-commit.yaml
```diff
@@ -25,19 +25,32 @@ jobs:
...
+        SKIP: no-commit-to-branch,hadolint
       with:
         token: ${{ secrets.GITHUB_TOKEN }}
         extra_args: --color=always --show-diff-on-failure --files ${{ steps.file_changes.outputs.files }}
+    # Run only skipped checks
+    - name: Execute pre-commit check that have no auto-fixes
+      if: always()
+      uses: pre-commit/action@v2.0.0
+      env:
+        SKIP: check-added-large-files,check-merge-conflict,check-vcs-permalinks,forbid-new-submodules,no-commit-to-branch,end-of-file-fixer,trailing-whitespace,check-yaml,check-merge-conflict,check-executables-have-shebangs,check-case-conflict,mixed-line-ending,detect-aws-credentials,detect-private-key,shfmt,shellcheck
```
yermulnik, line 54: I'd sort the list alphabetically to ease this list manegement.

yermulnik (approved):

hooks/infracost_breakdown.sh
```diff
@@ -73,8 +73,8 @@ function infracost_breakdown_ {
     }; then
       check="${check:1:-1}"
     fi
-    # shellcheck disable=SC2207 # Can't find working `read` command
-    operations=($(echo "$check" | grep -oE '[!<>=]{1,2}'))
+
+    mapfile -t operations < <(echo "$check" | grep -oE '[!<>=]{1,2}')
```
antonbabenko, line 77: what is `mapfile` for?

MaxymVlasov: https://github.com/koalaman/shellcheck/wiki/SC2207

MaxymVlasov: >On Unix-like operating systems, mapfile is a builtin command of the Bash shell. It reads lines from standard input into an indexed array variable.

antonbabenko: TIL :)

.github/workflows/pre-commit.yaml
MaxymVlasov: Now they sorted as they list in `pre-commit-config.yaml` order and include all hooks except `hadolint`.
I can sort them alphabetically, but not see what it can change.
Add a new hook? Just add it to the end
Remove hook? Can be simply ignored or removed from "array"

hooks/infracost_breakdown.sh
yermulnik: Agree with Anton's "TIL" 😄 I learned about this command some relatively short time ago but still always forget about it when it may be useful better than other approaches 😝

.github/workflows/pre-commit.yaml
MaxymVlasov: Hmm, this is the first time I've seen such an approach.
If the command took up less space, you could think about using it

.github/CONTRIBUTING.md
MaxymVlasov: Ah, yes. Thanks

.github/CONTRIBUTING.md
```diff
@@ -73,7 +73,7 @@ sudo apt install -y datamash
 
 ```bash
 # Build `pre-commit` image
```
MaxymVlasov, line 75:
```suggestion
# Build `pre-commit-terraform` image
```

.github/workflows/pre-commit.yaml
yermulnik: I was thinking more of avoiding additions of items which are already in the list.
Though it's not quire essential in this use case. Up to you dude to consider 😉
Resolving this convo.

.github/workflows/pre-commit.yaml
yermulnik: > Hmm, this is the first time I've seen such an approach.

There's always a space to learn. Just like the `mapfile` command 😏
`install` is more commonly used within packages' installation scripts and (just like `mapfile`) isn't commonly known by regular users.

> If the command took up less space, you could think about using it

This is not about space consumption, but about proper file installation including mode, user, and group (it is not a good practice to have user-owned files in system locations) along with using a single command which is designed specifically for the use case we have here instead if using chmod/chown/mv.

Again: this is not essential. Up to you to consider.

yermulnik: > If the command took up less space, you could think about using it

btw, the `mapfile` change is 9 chars longer than what existed before 😏

yermulnik (approved):

PR #323, opened 2022-01-12
chore: Adding version check arg for terraform_docs hook

### LINKED ISSUE #246: Add a check for specific version of a tool before applying a hook

It would be helpful to be able to set hard requirements on the version of certain tools (e.g. `terraform-docs`) inside the YAML config so that the hook fails&stops if the version mismatches at the very beginning.

Potential config:

```yaml
- id: terraform_docs
  args:
    - TERRAFORM_DOCS_ALLOWED_VERSION_REGEXP="0.13.[0-9]+"
    - tfvars hcl --output-file terraform.tfvars.model .
```

I think it should be applied to `terraform_docs` hooks first of all.

Ref: https://tldp.org/LDP/abs/html/bashver3.html#REGEXMATCHREF

mohitsaxenaknoldus:

- [ ] This PR fixes a bug.

Closes #246

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -16,6 +16,9 @@
   exclude: \.terraform\/.*$
 
 - id: terraform_docs
+  args:
+    - TERRAFORM_DOCS_ALLOWED_VERSION_REGEXP="0.13.[0-9]+"
+    - tfvars hcl --output-file terraform.tfvars.model .
   name: Terraform docs
   description: Inserts input and output documentation into README.md (using terraform-docs).
   require_serial: true


.pre-commit-hooks.yaml
```diff
@@ -16,6 +16,9 @@
   exclude: \.terraform\/.*$
 
 - id: terraform_docs
+  args:
+    - TERRAFORM_DOCS_ALLOWED_VERSION_REGEXP="0.13.[0-9]+"
+    - tfvars hcl --output-file terraform.tfvars.model .
```
yermulnik, line 21: Could you please add a description of what this change is for, what its benefits are, and whether it is backward compatible with existing setups?

MaxymVlasov: That should be in `.pre-commit-config.yaml` file.
Your proposal is hardcode.

PR #290, opened 2021-11-30
feat: Add `terraform_checkov`, that run per folder. Deprecate `checkov` hook

bmbferreira:
Put an `x` into the box if that apply:

- [x] This PR introduces breaking change.
- [ ] This PR fixes a bug.
- [ ] This PR adds new functionality.
- [x] This PR enhances existing functionality.

### Description of your changes

This change applies the same logic to `checkov` that is already being used for other tools, such as `tflint`. If we have a monorepo with multiple modules such as `module1` and `module2` and we run for example `pre-commit run --files module1/istio/*`, it would run `checkov` for all the modules but `tflint` would just run for `module1`, which is inconsistent with the argument that was passed to `pre-commit`.  

### How has this code been tested

In a monorepo with multiple modules, before if you ran `pre-commit run --files module1/istio/*`, `checkov` would execute for all the modules because of the `-d .` while other tools such as `tflint` would just execute for the paths passed as argument to `pre-commit`. With this change, `checkov` will start having the same behaviour as other tools, improving the consistency between the different tools.

PR #329, opened 2022-01-27
feat: Make checkov more flexible

Miscreancy:

- [ ] This PR fixes a bug.
- [X] This PR enhances existing functionality.

Currently, the checkov hook is hard-coded to run against the entire directory - no override is possible.

This change allows the hook to run in any of three modes - whole directory (the default), against all directories which contain a changed file, and against only changed files.

It should be noted that:

- There is a PR in place for similar but different functionality; the disparity between the two approaches I thought was sufficient to justify a separate PR as the code and logic is very different. I will post a link to this PR in that PR's discussion to raise the question as to which route should be followed.
- Common scripts were not used deliberately; this is for two reasons. 1. Using the `common::parse_cmdline` function would've forced the introduction of a breaking change (having to preface all arguments with `--args`) and 2. Using the `common::per_dir_hook` function would have removed the ability to pass multiple arguments to checkov, as the code block at line 70 assumes that only $1 is an argument and the rest are files.

### How can we test changes

- Use a dummy IaC repo with multiple modules and known security issues
- Use `pre-commit try-repo $path/to/clone` to confirm the default behaviour is still scanning the entire repository
- Use manual script runs of `checkov.sh` from within that dummy directory to confirm that when passed `--scan-change-directories` (and file names), the tool only scans directories that contain a change file, and while passed `--scan-change-files`, the tool only scans files, as well as any other changes and logic deemed necessary.

### DIFF
--- a/hooks/checkov.sh
+++ b/hooks/checkov.sh
@@ -1,0 +1,123 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+#######################################################################
+# Wrapper function for hook. Determines run mode based on bool set by
+# argparse, and triggers that run mode.
+# Globals:
+#   None
+# Arguments:
+#   None
+# Returns:
+#   0 if successful, non-zero on error
+#######################################################################
+function main {
+  parse_arguments "$@"
+  # shellcheck disable=SC2153 # False positive
+  if [[ ${CHANGE_DIRECTORY_SCAN} == true ]]; then
+    # directories containing changed .tf files only
+    directory_runner
+  elif [[ ${CHANGE_FILE_SCAN} == true ]]; then
+    # changed .tf files only
+    file_runner
+  else
+    # whole repository
+    checkov "${ARGS[@]}" -d .
+  fi
+}
+
+#######################################################################
+# Parses arguments provided to the hook to filter out changed files and
+# args for the hook, set run mode (if present) and remove run mode toggles
+# from the args as they're not valid for checkov. Enables backwards
+# compatibility for those using checkov with no args, or without --args
+# Globals:
+#   None
+# Arguments:
+#   None
+# Returns:
+#   0 if successful, non-zero on error
+#   Creates global arrays ARGS, FILES, may create a run mode global
+#######################################################################
+function parse_arguments {
+  declare -a -g ARGS=()
+  declare -a -g FILES=()
+
+  argv=("$@")
+  pattern='^.*\.tf$'
+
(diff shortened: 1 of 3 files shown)


hooks/checkov.sh
```diff
@@ -0,0 +1,123 @@
...
+# Returns:
+#   0 if successful, non-zero on error
+#   Creates global arrays ARGS, FILES, may create a run mode global
+#######################################################################
+function parse_arguments {
+  declare -a -g ARGS=()
+  declare -a -g FILES=()
+
+  argv=("$@")
+  pattern='^.*\.tf$'
```
yermulnik, line 47: The any_chars_from_beginning part of pattern makes no much sense and can be dropped
```suggestion
  pattern='\.tf$'
```

yermulnik (commented):
The code looks sensible to me, though I'll leave design appropriateness consideration and decision to @MaxymVlasov and @antonbabenko

MaxymVlasov:
>There is a PR in place for similar but different functionality; the disparity between the two approaches I thought was sufficient to justify a separate PR as the code and logic is very different. I will post a link to this PR in that PR's discussion to raise the question as to which route should be followed.
> ... 1. Using the common::parse_cmdline function would've forced the introduction of a breaking change (having to preface all arguments with --args)...

 #290 + compatibility in args parsing by replacing/changing `common::parse_cmdline`, but not sure it's worth it.

> 2. Using the common::per_dir_hook function would have removed the ability to pass multiple arguments to checkov, as the code block at line 70 assumes that only $1 is an argument and the rest are files.

Looks like you mean this code: 
https://github.com/antonbabenko/pre-commit-terraform/blob/321fb1669366f3087f6942af463b2d355120451a/hooks/_common.sh#L65-L72


`args` not `arg`. The answer already in var name and func description :) 
It was done in this way during bash limitation

This is a string with array:
https://github.com/antonbabenko/pre-commit-terraform/blob/321fb1669366f3087f6942af463b2d355120451a/hooks/terragrunt_fmt.sh#L13


That then passed as a string with array
https://github.com/antonbabenko/pre-commit-terraform/blob/321fb1669366f3087f6942af463b2d355120451a/hooks/_common.sh#L96
and then - expanded to an array
https://github.com/antonbabenko/pre-commit-terraform/blob/321fb1669366f3087f6942af463b2d355120451a/hooks/terragrunt_fmt.sh#L33


So this argument is not vaild.

PR #332, opened 2022-01-29
docs: Add workaround for configuration_aliases tf bug

### LINKED ISSUE #330: terraform_validate fails in modules with provider aliases
### Describe the bug

Terraform Validate is broken when running in a module that uses provider aliases, which in turn breaks pre_commit validation checks.

### How can we reproduce it?

When running the pre-commit with a module that has multiple providers defined like so:
```hcl
terraform {
  required_version = ">= 0.15"
  providers {
    aws = {
      version = ">= 3.0"
      configuration_aliases = [aws.other]
    }
  }
}
```
With some resource defined that uses `provider = aws.other`

When terraform validate runs it throws an error:

> Error: Provider configuration not present
> 
> To work with data.aws_kms_key.other its original provider configuration at provider["registry.terraform.io/hashicorp/aws"].primary is required, but it has been removed. This occurs when a provider configuration is removed while objects created by that provider still exist in the state. Re-add the provider configuration to destroy data.aws_kms_key.primary, after which you can
> remove the provider configuration again.     

### Environment information
The only that that matters in terms of the environment is that you're on a recent enough version of terraform to support module-based provider aliases (0.15+) and that you're using that feature and terraform_validate. It's a bug in terraform technically.

mbainter:

- [ ] This PR fixes a bug.
- [X] This PR updates documentation

As noted in the change, there is a bug affecting terraform validate
when used with a module that also requires multiple aliased providers.
There is little activity on this bug, leave it an open question on if
or when it will be fixed. This documents a couple of workaround options
in the interim.

Fixes #330 

### How can we test changes

Define a module with configuration aliases as described in #330, and run terraform validate.  This should error out.
Following the instructions provided in the documentation update prevents the error and offers workarounds other than disabling the terraform_valdiate hook.

### DIFF
--- a/README.md
+++ b/README.md
@@ -551,6 +551,51 @@ Example:
 
     **Warning:** If you use Terraform workspaces, DO NOT use this workaround ([details](https://github.com/antonbabenko/pre-commit-terraform/issues/203#issuecomment-918791847)). Wait to [`force-init`](https://github.com/antonbabenko/pre-commit-terraform/issues/224) option implementation.
 
+5. `terraform_validate` in a 0.15+ module with configuration_aliases errors out
+
+   When running the hook where you have configuration_aliases defined as required in the terraform block, terraform will throw an error like:
+   >
+   >
+   > Error: Provider configuration not present
+   > To work with <resource> its original provider configuration at provider["registry.terraform.io/hashicorp/aws"].<provider_alias> is required, but it has been removed. This occurs when a provider configuration is removed while
+   > objects created by that provider still exist in the state. Re-add the provider configuration to destroy <resource>, after which you can remove the provider configuration again.
+
+   This is a [known issue](https://github.com/hashicorp/terraform/issues/28490) with Terraform and how providers are initialized in 0.15 and later. To work around this for now you can add an exclude to the `terraform_validate` hook like this:
+   ```yaml
+   - id: terraform_validate
+     exclude: [^/].+$
+   ```
+   This will exclude the root directory from being processed by this hook. Then add a subdirectory like "examples" or "tests" and put an example implementation in place that defines the providers with the proper aliases, and this will give you validation of your module through the example.
+
+   Alternately, you can use [terraform-config-inspect](https://github.com/hashicorp/terraform-config-inspect) and use a variant of [this script](https://github.com/bendrucker/terraform-configuration-aliases-action/blob/main/providers.sh) to generate a providers file at runtime:
+
+   ```bash
+   terraform-config-inspect --json . | jq -r '
+     [.required_providers[].aliases]
+     | flatten
+     | del(.[] | select(. == null))
+     | reduce .[] as $entry (
+       {};
+       .provider[$entry.name] //= [] | .provider[$entry.name] += [{"alias": $entry.alias}]
+     )
+   ' | tee aliased-providers.tf.json
+   ```
+
+   Save this as .generate-providers.sh in the root of your repository and place it in the and add a hook to run that first, like so:
+   ```yaml
+   - repo: local
+     hooks:
+       - id: generate-terraform-providers
+          name: generate-terraform-providers
+          require_serial: true
+          entry: .generate-providers.sh
+          language: script
+          files: (\.tf|\.tfvars)$
+          pass_filenames: false
+   ```
+
+   **Note:** This latter method will leave an "aliased-providers.tf.json" file in your path. You will either want to automate a way to clean this up or add it to your gitignore or both.
+
 ### terrascan
 
 1. `terrascan` supports custom arguments so you can pass supported flags like `--non-recursive` and `--policy-type` to disable recursive inspection and set the policy type respectively:
--- a/README.md
+++ b/README.md
@@ -551,21 +551,21 @@ Example:
 
     **Warning:** If you use Terraform workspaces, DO NOT use this workaround ([details](https://github.com/antonbabenko/pre-commit-terraform/issues/203#issuecomment-918791847)). Wait to [`force-init`](https://github.com/antonbabenko/pre-commit-terraform/issues/224) option implementation.
 
-5. `terraform_validate` in a 0.15+ module with configuration_aliases errors out
+5. `terraform_validate` in a repo with Terraform module, written using Terraform 0.15+ and which uses provider `configuration_aliases` ([Provider Aliases Within Modules](https://www.terraform.io/language/modules/develop/providers#provider-aliases-within-modules)), errors out.
 
-   When running the hook where you have configuration_aliases defined as required in the terraform block, terraform will throw an error like:
+   When running the hook against Terraform code where you have provider `configuration_aliases` defined in a `required_providers` configuration block, terraform will throw an error like:
    >
    >
    > Error: Provider configuration not present
    > To work with <resource> its original provider configuration at provider["registry.terraform.io/hashicorp/aws"].<provider_alias> is required, but it has been removed. This occurs when a provider configuration is removed while
    > objects created by that provider still exist in the state. Re-add the provider configuration to destroy <resource>, after which you can remove the provider configuration again.
 
-   This is a [known issue](https://github.com/hashicorp/terraform/issues/28490) with Terraform and how providers are initialized in 0.15 and later. To work around this for now you can add an exclude to the `terraform_validate` hook like this:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -551,6 +551,51 @@ Example:
 
     **Warning:** If you use Terraform workspaces, DO NOT use this workaround ([details](https://github.com/antonbabenko/pre-commit-terraform/issues/203#issuecomment-918791847)). Wait to [`force-init`](https://github.com/antonbabenko/pre-commit-terraform/issues/224) option implementation.
 
+5. `terraform_validate` in a 0.15+ module with configuration_aliases errors out
```
yermulnik, line 554:
```suggestion
5. `terraform_validate` in a repo with Terraform module, written using Terraform 0.15+ and which uses provider `configuration_aliases` ([Provider Aliases Within Modules](https://www.terraform.io/language/modules/develop/providers#provider-aliases-within-modules)), errors out.
```

README.md
```diff
@@ -551,6 +551,51 @@ Example:
...
+
+   When running the hook where you have configuration_aliases defined as required in the terraform block, terraform will throw an error like:
```
yermulnik, line 556:
```suggestion
   When running the hook against Terraform code where you have provider `configuration_aliases` defined in a `required_providers` configuration block, terraform will throw an error like:
```

README.md
```diff
@@ -551,6 +551,51 @@ Example:
...
+   >
+   >
+   > Error: Provider configuration not present
+   > To work with <resource> its original provider configuration at provider["registry.terraform.io/hashicorp/aws"].<provider_alias> is required, but it has been removed. This occurs when a provider configuration is removed while
+   > objects created by that provider still exist in the state. Re-add the provider configuration to destroy <resource>, after which you can remove the provider configuration again.
+
+   This is a [known issue](https://github.com/hashicorp/terraform/issues/28490) with Terraform and how providers are initialized in 0.15 and later. To work around this for now you can add an exclude to the `terraform_validate` hook like this:
```
yermulnik, line 563:
```suggestion
   This is a [known issue](https://github.com/hashicorp/terraform/issues/28490) with Terraform and how providers are initialized in Terraform 0.15 and later. To work around this you can add an `exclude` parameter to the configuration of `terraform_validate` hook like this:
```

README.md
```diff
@@ -551,6 +551,51 @@ Example:
...
+   ```yaml
+   - id: terraform_validate
+     exclude: [^/].+$
```
yermulnik, line 566: This is a bit ambiguous regex, which means "every item, consisting of a minimum of two chars, with NOT `/` in it".
Could you please elaborate on what the real-life example would be?
Is it sort of "exclude everything"? If yes, then it should be `^.+$`, imho.

README.md
```diff
@@ -551,6 +551,51 @@ Example:
...
+   ```
+   This will exclude the root directory from being processed by this hook. Then add a subdirectory like "examples" or "tests" and put an example implementation in place that defines the providers with the proper aliases, and this will give you validation of your module through the example.
```
yermulnik, line 568:
```suggestion
   This will exclude the root directory from being processed by this hook. Then add a subdirectory like "examples" or "tests" and put an example implementation in place that defines the provider(s) with `configuration_aliases`, and this will give you validation of your Terraform module through the example.
```

yermulnik: Also could you please add an example of how to «add a subdirectory like "examples" or "tests"»?

README.md
```diff
@@ -551,6 +551,51 @@ Example:
...
+     | flatten
+     | del(.[] | select(. == null))
+     | reduce .[] as $entry (
+       {};
+       .provider[$entry.name] //= [] | .provider[$entry.name] += [{"alias": $entry.alias}]
+     )
+   ' | tee aliased-providers.tf.json
+   ```
+
+   Save this as .generate-providers.sh in the root of your repository and place it in the and add a hook to run that first, like so:
```
yermulnik, line 584:
```suggestion
   Save it as `.generate-providers.sh` in the root of your repository and add a `pre-commit` hook to run it before all other hooks, like so:
```

README.md
```diff
@@ -551,6 +551,51 @@ Example:
...
+   ```yaml
+   - repo: local
+     hooks:
+       - id: generate-terraform-providers
+          name: generate-terraform-providers
+          require_serial: true
+          entry: .generate-providers.sh
+          language: script
+          files: (\.tf|\.tfvars)$
+          pass_filenames: false
```
yermulnik, line 594:
```suggestion
   repos:
     - repo: local
       hooks:
         - id: generate-terraform-providers
           name: generate-terraform-providers
           require_serial: true
           entry: .generate-providers.sh
           language: script
           files: \.tf(vars)?$
           pass_filenames: false

     - repo: https://github.com/pre-commit/pre-commit-hooks
[…]
```

ps: please adjust indentation in general if I messed it up

README.md
```diff
@@ -551,6 +551,51 @@ Example:
...
+   ```
+
+   **Note:** This latter method will leave an "aliased-providers.tf.json" file in your path. You will either want to automate a way to clean this up or add it to your gitignore or both.
```
yermulnik, line 597:
```suggestion
   **Note:** The latter method will leave an "aliased-providers.tf.json" file in your repo. You will either want to automate a way to clean this up or add it to your `.gitignore` or both.
```

yermulnik (commented):
Thank you for the contribution.
Please consider my comments.

README.md
mbainter: That could be challenging to do without making this excessively long.  Would it be sufficient, do you think, if I were to link to a module in terraform-aws-modules that has examples?  Maybe [terraform-examples-sqs](https://github.com/terraform-aws-modules/terraform-aws-sqs/tree/master/examples/complete)?

README.md
mbainter: Oops, that's actually wrong.  That should be `exclude: [^/]+$`.

The idea is to exclude the files in directory where the module in question is, but not the subdirectories so that the examples get processed.  So the regex should be "match any string that is 1 or more characters long that doesn't have a slash in it".  So if it's a root module like say, tf-aws-modules/terraform-aws-sqs then you'd use the above regex, and it would not process the module itself, but would validate all of the examples.

OTOH, if it's a monorepo with a lot of modules under `modules/` then you would want something like `exclude: modules/[^/]+$/[^/]+$` to exclude just the contents of each of the module directories.

mbainter: I fixed the regex and amended it to also explain how to handle where the module is not in the root of the repository.

README.md
yermulnik: I was thinking of something like `include: ^examples/`

Also I didn't get what you mean about linking a module in terraform-aws-modules 🤔

README.md
```diff
@@ -551,6 +551,54 @@ Example:
...
+   ```yaml
+   - repos:
+     - repo: local
+       hooks:
+         - id: generate-terraform-providers
+            name: generate-terraform-providers
+            require_serial: true
+            entry: .generate-providers.sh
+            language: script
+            files: (\.tf|\.tfvars)$
```
yermulnik, line 594:
```suggestion
            files: \.tf(vars)?$
```

README.md
```diff
@@ -551,6 +551,54 @@ Example:
...
+            pass_filenames: false
+     - repo: https://github.com/pre-commit/pre-commit-hooks
```
yermulnik, line 596:
```suggestion

     - repo: https://github.com/pre-commit/pre-commit-hooks
```

README.md
mbainter: Oh.. that's not necessary with my exclude regex.  Since it only excludes the root path for the module, the examples under it still get run.  I thought you mean how to create an example for a module.

yermulnik: Aha, got you. Indeed... Apologies for the confusion.

yermulnik (approved):
LGTM
Any change pending on your end @mbainter? Otherwise I'm fine with merging it — please let me know.

antonbabenko:
@mbainter please confirm that this PR is ready to be merged or do you want to add something else?

Thanks a lot for this PR!

mbainter:
I'm guessing those emoji responses didn't generate any notifications for you - but this is fine by me if you're happy with it.  I can squash these two commits together now to clean up the history or you can when you merge it - up to you.

antonbabenko:
@mbainter Thank you. You are right, emojis are nice but not they don't produce any notifications in my inbox. :)

antonbabenko:
This PR is included in version 1.63.0 :tada:

smiller171:
I cannot for the life of me get the `exclude` directive to work as expected for the root of the repo

smiller171:
The example listed throws an error. This `^([^/])+$` seems to work with `identity` but it doesn't stop terraform from running in the wrong dir with `terraform_validate`

PR #338, opened 2022-02-10
feat: Improved speed of `pre-commit run -a` for multiple hooks

### LINKED ISSUE #309: Improve perfomance during `pre-commit --all` (`-a`) run

MaxymVlasov:
…m_tfsec`, `terragrunt_fmt` and `terragrunt_validate`

- [ ] This PR fixes a bug.

Speedup hooks execution on `pre-commit run --all`

Closes #309

### How can we test changes

```yaml
repos:
  rev: 7ebbe57df62dce6d50cfe11a12566850d96a9613
  hooks:
    - id: terraform_tfsec
      #args:
      #  - >
      #    --args=--config-file=.tfsec.json
    - id: terragrunt_fmt
    - id: terragrunt_validate
```

### DIFF
--- a/hooks/terraform_tfsec.sh
+++ b/hooks/terraform_tfsec.sh
@@ -43,4 +43,22 @@ function per_dir_hook_unique_part {
   return $exit_code
 }
 
+#######################################################################
+# Unique part of `common::per_dir_hook`. The function is executed one time
+# in the root git repo
+# Arguments:
+#   args (string with array) arguments that configure wrapped tool behavior
+#######################################################################
+function run_hook_on_whole_repo {
+  local -r args="$1"
+
+  # pass the arguments to hook
+  # shellcheck disable=SC2068 # hook fails when quoting is used ("$arg[@]")
+  tfsec "$(pwd)" ${args[@]}
+
+  # return exit code to common::per_dir_hook
+  local exit_code=$?
+  return $exit_code
+}
+
 [ "${BASH_SOURCE[0]}" != "$0" ] || main "$@"
--- a/hooks/terragrunt_fmt.sh
+++ b/hooks/terragrunt_fmt.sh
@@ -40,4 +40,22 @@ function per_dir_hook_unique_part {
   return $exit_code
 }
 
+#######################################################################
+# Unique part of `common::per_dir_hook`. The function is executed one time
+# in the root git repo
+# Arguments:
+#   args (string with array) arguments that configure wrapped tool behavior
+#######################################################################
+function run_hook_on_whole_repo {
+  local -r args="$1"
+
+  # pass the arguments to hook
+  # shellcheck disable=SC2068 # hook fails when quoting is used ("$arg[@]")
+  terragrunt hclfmt "$(pwd)" ${args[@]}
+
+  # return exit code to common::per_dir_hook
+  local exit_code=$?
+  return $exit_code
+}
+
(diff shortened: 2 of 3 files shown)


antonbabenko:
This PR is included in version 1.64.0 :tada:

sergei-ivanov:
I am afraid that this may likely produce incorrect results with `tfsec`, see https://github.com/aquasecurity/tfsec/issues/1452

MaxymVlasov:
I didn't think that is bug. As I see (https://github.com/aquasecurity/tfsec/issues/1452#issuecomment-1026957993), this is expected `tfsec` behavior.:

>... input variables are taken into account when loading a module from a parent vs loading it directly by scanning it's directory, just like if you ran terraform in the root vs the module dir you would see a slightly different plan depending on input vars.



Anyway, if you'll see come incorrect results, feel free to submit issue/PR

sergei-ivanov:
From tfsec maintainers' point of view, it is expected behaviour. And I can see their point, that's why I did not challenge it there.

But from the end user's point of view, tfsec produces different results if I run it from the top of the repository. It flags up a lot of false positives that are not flagged up if I run it module-by-module.

That's why I am concerned that the pre-commit hook is going to produce different results if I run it for all files vs. changed files only. Previously it was consistent in that respect.

PR #340, opened 2022-02-10
chore: Fix docker test workflow

MaxymVlasov:
Invalid workflow file: .github/workflows/build-image-test.yaml#L22
The workflow is not valid. .github/workflows/build-image-test.yaml (Line: 22, Col: 13): Unexpected symbol: '"true"'. Located at position 53 within expression: steps.changed-files-specific.outputs.any_changed == "true"

### DIFF
--- a/hooks/_common.sh
+++ b/hooks/_common.sh
@@ -27,7 +27,7 @@ function common::initialize {
 function common::parse_cmdline {
   # common global arrays.
   # Populated via `common::parse_cmdline` and can be used inside hooks' functions
-  ARGS=() HOOK_CONFIG=() FILES=()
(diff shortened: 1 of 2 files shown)


hooks/_common.sh
```diff
@@ -27,7 +27,7 @@ function common::initialize {
 function common::parse_cmdline {
   # common global arrays.
   # Populated via `common::parse_cmdline` and can be used inside hooks' functions
-  declare -g -a ARGS=() HOOK_CONFIG=() FILES=()
+  ARGS=() HOOK_CONFIG=() FILES=()
```
MaxymVlasov, line 30:
```suggestion
  declare -g -a ARGS=() HOOK_CONFIG=() FILES=()
```

PR #347, opened 2022-03-11
feat: Adding getopt flag and feature: --env-passthrough

andrew-glenn:

- [ ] This PR fixes a bug. 
- [X] This PR adds new functionality.
- [X] This PR enhances existing functionality.

Context: I need to pass in the location of an alternate `tflint` config in CI.
The location is based on an environment variable in my CI environment, so
hard-coding the value as an argument isn't tenable.

This PR adds the `__ENV_*__` placeholder, and the argument `--env-passthrough`.
The idea is to replace the `__ENV_FOO__` placeholder with the value of
environment variable `${FOO}`.

Example config leveraging the changes in this PR:

```
repos:
    rev: v1.65.0
    hooks:
      - id: terraform_tflint
        args:
          - "--env-passthrough=FOO"
          - "--args=--config=__ENV_FOO__/.tflint.hcl"
```

_Note: hooks which do not use the common argument parser in `_common.sh` were not modified to use this functionality_

### How can we test changes

- From AWS Codebuild:
`pre-commit try-repo https://andrew-glenn/pc-sandbox env-debug`

For transparency, here's the `.pre-commit-hooks.yaml` config in my sandbox repo
https://github.com/andrew-glenn/pc-sandbox/blob/abcdad5b40aee08484185451912b8ad191f80138/.pre-commit-hooks.yaml

as well as the script that echo's args passed to it:

https://github.com/andrew-glenn/pc-sandbox/blob/abcdad5b40aee08484185451912b8ad191f80138/env.sh#L1-L19

- Others:

Modify your local `.pre-commit-config` file to add my sandbox repo, replacing
`FOO` with an environment variable name of your choice.

```
repos:
  - repo: https://github.com/andrew-glenn/pc-sandbox
    rev: abcdad5b40aee08484185451912b8ad191f80138
    hooks:
      - id: env-debug
        args:
          - "--env-passthrough=FOO"
          - "--args=--config=__ENV_FOO__"
```

Example output:
```
===============================================================================
[INFO] Initializing environment for https://github.com/andrew-glenn/pc-sandbox.
env debug................................................................Passed
- hook id: env-debug
- duration: 0.01s

--foo=/codebuild/output/src982431412/src
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -303,7 +303,7 @@ All notable changes to this project will be documented in this file.
 
 - fix: Change terraform_validate hook functionality for subdirectories with terraform files ([#100](https://github.com/antonbabenko/pre-commit-terraform/issues/100))
 
-### 
+###
 
 configuration for the appropriate working directory.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -331,7 +331,7 @@ All notable changes to this project will be documented in this file.
 
 - fix: Change terraform_validate hook functionality for subdirectories with terraform files ([#100](https://github.com/antonbabenko/pre-commit-terraform/issues/100))
 
-###
+### 
 
 configuration for the appropriate working directory.
 
--- a/hooks/_common.sh
+++ b/hooks/_common.sh
@@ -27,10 +27,10 @@ function common::initialize {
 function common::parse_cmdline {
   # common global arrays.
   # Populated via `common::parse_cmdline` and can be used inside hooks' functions
-  declare -g -a ARGS=() HOOK_CONFIG=() FILES=()
+  declare -g -a ARGS=() HOOK_CONFIG=() FILES=() ENV_PASSTHRU=()
 
   local argv
-  argv=$(getopt -o a:,h: --long args:,hook-config: -- "$@") || return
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -331,7 +331,7 @@ All notable changes to this project will be documented in this file.
 
 - fix: Change terraform_validate hook functionality for subdirectories with terraform files ([#100](https://github.com/antonbabenko/pre-commit-terraform/issues/100))
 
-### 
+###
```
MaxymVlasov, line 334:
```suggestion
### 
```

MaxymVlasov (commented):
I think we can introduce logic that covers more cases.

Firstly, we should avoid difficult-to-understand logic for users.
Secondly, it should be simple to understand and know how to use (UX).

Hooks are not run in environment isolation (that's where from your use case) so, the first idea - is just to use the POSIX standard of environment variables.

I spend some time with @yermulnik brainstorming how it can be done and created #347. 
Please check if that will work for you.

P.S. Before I start working on #347, I quickly review your PR and leave some comments. No more need to implement suggestions in them, I keep them JFYI, hope it will be helpful :)

antonbabenko:
This issue has been resolved in version 1.69.0 :tada:

PR #352, opened 2022-03-16
feat: Adding init to terraform_tflint hook

andrew-glenn:

- [ ] This PR fixes a bug.
- [x ] This PR enhances existing functionality.

Just a quick `tflint --init` before running tflint to ensure plugins are installed.

### DIFF
--- a/hooks/terraform_tflint.sh
+++ b/hooks/terraform_tflint.sh
@@ -36,6 +36,11 @@ function per_dir_hook_unique_part {
 
   # Print checked PATH **only** if TFLint have any messages
   # shellcheck disable=SC2091,SC2068 # Suppress error output
+  $(tflint --init 2>&1) 2> /dev/null || {
+    common:colorify "yellow" "tflint init:"
+    tflint --init
+  }
+  # shellcheck disable=SC2091,SC2068 # Suppress error output
   $(tflint ${args[@]} 2>&1) 2> /dev/null || {
     common::colorify "yellow" "TFLint in $dir_path/:"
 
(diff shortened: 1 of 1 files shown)


hooks/terraform_tflint.sh
```diff
@@ -36,6 +36,11 @@ function per_dir_hook_unique_part {
 
   # Print checked PATH **only** if TFLint have any messages
   # shellcheck disable=SC2091,SC2068 # Suppress error output
+  $(tflint --init 2>&1) 2> /dev/null || {
+    common:colorify "yellow" "tflint init:"
+    tflint --init
+  }
```
yermulnik, line 42: The `init` command seems to be a one-off command if there were no changes to `tflint` plugins and hence the first run of `tflint --init` will do the thing and invalidate the second run and hide the initial `tflint` plugins initialization.
I think it's worth to do something like the below:
```
TFLINT_INIT=$(tflint --init 2>&1) 2> /dev/null || {
  local exit_code=$?
  common:colorify "yellow" "tflint init:"
  echo "$TFLINT_INIT"
  return $exit_code
}

This needs proper testing as I wrote this from the top of my head w/o running the script to ensure the code is Okay.

andrew-glenn: Done, and you were spot on.

andrew-glenn:
@yermulnik  - feedback addressed.

mbainter:
@andrew-glenn I think you need to capitalize the `a` in `Adding` in your PR title for it to pass that still failing check.

yermulnik (approved):
LGTM

yermulnik:
@antonbabenko @MaxymVlasov I haven't tested this PR, though it looks good and should not break anything. The failed PR check fix is here #358 358

antonbabenko:
This PR is included in version 1.65.0 :tada:

MaxymVlasov:
Ok, I'll create separate PR

Issue #54, opened 2019-07-04
It should show the file path when something is changed or goes wrong


robsonpeixoto:
When something get wrong, it doesn't show the file path and it is very useful.

```
❯ git commit -v
[WARNING] Unstaged files detected.
[INFO] Stashing unstaged files to /Users/robinho/.cache/pre-commit/patch1562252730.
Terraform fmt............................................................Failed
hookid: terraform_fmt

Files were modified by this hook. Additional output:

cluster.tf

Error: Invalid block definition

  on vars.tf line 20:
  (source code not available)

A block definition must have block content delimited by "{" and "}", starting
on the same line as the block header.
```

varac:
Oh just noticed I created a duplicate of this (#161) and a PR to fix it (#162)

antonbabenko:
@MaxymVlasov Let's make sure that the path is also shown for all other hooks we have (including tflint, tfsec, fmt).

davidh-unmind:
Did you get anywhere with this? Appreciate the world is pretty F'd up right now. No rush at all.

It's pretty irritating getting messages like this though.

```
Terragrunt validate......................................................Failed
- hook id: terragrunt_validate
- files were modified by this hook
```

albertorm95:
> @MaxymVlasov Let's make sure that the path is also shown for all other hooks we have (including tflint, tfsec, fmt).

Any update on this?

PR #357, opened 2022-04-07
fix: Print out the full path in tflint when run through pre-commit

thomaslast:

- [x] This PR fixes a bug.

This PR aims to fix the issue here: https://github.com/antonbabenko/pre-commit-terraform/issues/54

The issue for tflint is resolved by running the pre-commit at the repo level and the full path is then displayed on an error. 

### How can we test changes

```yaml
repos:
- repo: https://github.com/thomaslast/pre-commit-terraform
  rev: afde258978090e4d2382fd90bb230347d0b349ac
```

✅ Work as expected 
☑️ - Able to check only passed, not fail scenario
✖️ - work, but not print full path
❌ - failed to start
Not checked

✅ `checkov` (python, not affected hook)
✅ `infracost_breakdown`
✅ `terraform_docs_replace` (deprecated, python, not affected hook)
✖️ `terraform_docs_without_aggregate_type_defaults`
✖️  `terraform_docs`
✅  `terraform_fmt`
❌ `terraform_providers_lock`
❌   `terraform_tflint`
✅  `terraform_tfsec`
☑️  `terraform_validate`
✖️  `terragrunt_fmt`
☑️  `terragrunt_validate`
❌  `terrascan`

### DIFF
--- a/hooks/_common.sh
+++ b/hooks/_common.sh
@@ -138,7 +138,7 @@ function common::per_dir_hook {
   # run hook for each path
   for dir_path in $(echo "${dir_paths[*]}" | tr ' ' '\n' | sort -u); do
     dir_path="${dir_path//__REPLACED__SPACE__/ }"
-    pushd "$dir_path" > /dev/null || continue
+    [ -f "$dir_path" ] || continue
 
     per_dir_hook_unique_part "$args" "$dir_path"
 
@@ -147,7 +147,6 @@ function common::per_dir_hook {
       final_exit_code=$exit_code
     fi
 
-    popd > /dev/null
   done
 
   # restore errexit if it was set before the "for" loop
--- a/hooks/_common.sh
+++ b/hooks/_common.sh
@@ -138,7 +138,7 @@ function common::per_dir_hook {
   # run hook for each path
   for dir_path in $(echo "${dir_paths[*]}" | tr ' ' '\n' | sort -u); do
     dir_path="${dir_path//__REPLACED__SPACE__/ }"
-    [ -f "$dir_path" ] || continue
+    [ ! -d "$dir_path" ] && continue
 
     per_dir_hook_unique_part "$args" "$dir_path"
 
--- a/hooks/terragrunt_fmt.sh
+++ b/hooks/terragrunt_fmt.sh
@@ -33,7 +33,7 @@ function per_dir_hook_unique_part {
 
   # pass the arguments to hook
   # shellcheck disable=SC2068 # hook fails when quoting is used ("$arg[@]")
-  terragrunt hclfmt ${args[@]}
+  terragrunt hclfmt ${args[@]} "$dir_path"
 
   # return exit code to common::per_dir_hook
   local exit_code=$?
--- a/hooks/terrascan.sh
+++ b/hooks/terrascan.sh
@@ -33,7 +33,7 @@ function per_dir_hook_unique_part {
 
   # pass the arguments to hook
   # shellcheck disable=SC2068 # hook fails when quoting is used ("$arg[@]")
-  terrascan scan -i terraform ${args[@]}
+  terrascan scan -i terraform ${args[@]} "$dir_path"
 
   # return exit code to common::per_dir_hook
   local exit_code=$?
--- a/hooks/terraform_fmt.sh
+++ b/hooks/terraform_fmt.sh
@@ -58,7 +58,7 @@ function terraform_fmt_ {
   # run hook for each path
   for dir_path in $(echo "${dir_paths[*]}" | tr ' ' '\n' | sort -u); do
     dir_path="${dir_path//__REPLACED__SPACE__/ }"
-    pushd "$dir_path" > /dev/null || continue
+    [ ! -d "$dir_path" ] && continue
 
     per_dir_hook_unique_part "$args" "$dir_path"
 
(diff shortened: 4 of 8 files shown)


hooks/_common.sh
```diff
@@ -138,7 +138,7 @@ function common::per_dir_hook {
   # run hook for each path
   for dir_path in $(echo "${dir_paths[*]}" | tr ' ' '\n' | sort -u); do
     dir_path="${dir_path//__REPLACED__SPACE__/ }"
-    pushd "$dir_path" > /dev/null || continue
+    [ -f "$dir_path" ] || continue
```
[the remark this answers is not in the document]
thomaslast: I've added in the dir_path so that it should run against the file for each hook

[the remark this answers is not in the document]
thomaslast: I've added in the dir_path so that it should run against the file for each hook

hooks/terragrunt_fmt.sh
```diff
@@ -33,7 +33,7 @@ function per_dir_hook_unique_part {
 
   # pass the arguments to hook
   # shellcheck disable=SC2068 # hook fails when quoting is used ("$arg[@]")
-  terragrunt hclfmt ${args[@]}
+  terragrunt hclfmt ${args[@]} "$dir_path"
```
yermulnik, line 36: Based on https://terragrunt.gruntwork.io/docs/reference/cli-options/#hclfmt this wouldn't work for `terragrunt`. The `hclfmt` recursively searches current working dir and isn't compatible with providing target working dir as a command line argument, hence the `terragrunt hclfmt` hook need an explicit `pushd`/`popd` workaround.

hooks/terrascan.sh
```diff
@@ -33,7 +33,7 @@ function per_dir_hook_unique_part {
 
   # pass the arguments to hook
   # shellcheck disable=SC2068 # hook fails when quoting is used ("$arg[@]")
-  terrascan scan -i terraform ${args[@]}
+  terrascan scan -i terraform ${args[@]} "$dir_path"
```
yermulnik, line 36: To provide target working directory to `terrascan` the `-d, --iac-dir string` command line flag must be provided:
```
> terrascan scan -h | fgrep -- --iac-dir
  -d, --iac-dir string            path to a directory containing one or more IaC files (default ".")
```
Please fix.

yermulnik (commented):
I added some more comments re the new portion of changes — please take a look.
Also could you please test-run all hooks you're requesting changes too and provide the output, so that we are sure there's no breakage of existing functionality which may impact other consumers of `pre-commit-terraform`. Thank you.

@MaxymVlasov Could you please take a look into this PR when you've got a chance so that we all agree upon this kind of changes? Thanks.

MaxymVlasov (changes requested):
Good try, but that will not work, because you are already in `$dir_path`. Usage of `"$(pwd)"` will not help too:

```bash
Oslo/environment/qa ➜ terraform fmt "$(pwd)"      
main.tf
````

You need to show an error msg like "that was changed in next dir" but not show it if exit code == 0

File: hooks/_common.sh [hunk 4]
```diff
@@ -138,7 +138,7 @@ function common::per_dir_hook {
   # run hook for each path
   for dir_path in $(echo "${dir_paths[*]}" | tr ' ' '\n' | sort -u); do
     dir_path="${dir_path//__REPLACED__SPACE__/ }"
-    pushd "$dir_path" > /dev/null || continue
+    [ ! -d "$dir_path" ] && continue
```
MaxymVlasov, line 141: Hmm, not mention that. Lmc

MaxymVlasov: Yeah, with that change it can work, but need to check that all hooks will work properly

MaxymVlasov:
Hmm, it does not print changed files at all, interesting

```bash
➜ pre-commit run
- hook id: terraform_docs
```

```bash
repos:
- repo: https://github.com/thomaslast/pre-commit-terraform
  rev: afde258978090e4d2382fd90bb230347d0b349ac
  hooks:
    - id: terraform_docs
```

hooks/_common.sh
MaxymVlasov: TFLint can't find TF modules by relative `source` without `pushd`

MaxymVlasov:
```yaml
repos:
- repo: https://github.com/thomaslast/pre-commit-terraform
  rev: afde258978090e4d2382fd90bb230347d0b349ac
```

✅ Work as expected 
☑️ - Able to check only passed, not fail scenario
✖️ - work, but not print full path
❌ - failed to start
Not checked

✅ `checkov` (python, not affected hook)
✅ `infracost_breakdown`
✅ `terraform_docs_replace` (deprecated, python, not affected hook)
✖️ `terraform_docs_without_aggregate_type_defaults`
✖️  `terraform_docs`
✅  `terraform_fmt`
❌ `terraform_providers_lock`
❌   `terraform_tflint`
✅  `terraform_tfsec`
✅  `terraform_validate`
✖️  `terragrunt_fmt`
☑️  `terragrunt_validate`
❌  `terrascan`

thomaslast:
Sorry I've gone AWOL on this, thanks @MaxymVlasov and @yermulnik for your help on this. This seems to be breaking change for something that could be quite simple. Is it worth making another version of `common::per_dir_hook` which will not use pushd & popd that I make use of just for tflint for the moment...?

PR #361, opened 2022-04-13
fix: Improve `tflint --init` command execution

MaxymVlasov:
Relates #352

### DIFF
--- a/hooks/terraform_tflint.sh
+++ b/hooks/terraform_tflint.sh
@@ -17,6 +17,20 @@ function main {
   # shellcheck disable=SC2178 # It's the simplest syntax for that case
   ARGS=${ARGS[*]/__GIT_WORKING_DIR__/$(pwd)\/}
   # shellcheck disable=SC2128 # It's the simplest syntax for that case
+
+  # Run `tflint --init` for check that plugins installed.
+  # It should run once on whole repo.
+  {
+    TFLINT_INIT=$(tflint --init 2>&1) 2> /dev/null &&
+      common::colorify "green" "Command 'tflint --init' successfully done:" &&
+      echo -e "${TFLINT_INIT}\n\n\n"
+  } || {
+    local exit_code=$?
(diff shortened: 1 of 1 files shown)


hooks/terraform_tflint.sh
```diff
@@ -17,6 +17,20 @@ function main {
   # shellcheck disable=SC2178 # It's the simplest syntax for that case
   ARGS=${ARGS[*]/__GIT_WORKING_DIR__/$(pwd)\/}
   # shellcheck disable=SC2128 # It's the simplest syntax for that case
+
+  # Run `tflint --init` for check that plugins installed.
+  # It should run once on whole repo.
+  {
+    TFLINT_INIT=$(tflint --init 2>&1) 2> /dev/null &&
+      common::colorify "green" "Command 'tflint --init' successfully done:" &&
+      echo -e "${TFLINT_INIT}\n\n\n"
```
yermulnik, line 26: Not sure that we really need to have this output even when `tflint --init` produced empty output (e.g. when all its plugins are up-to-date). Maybe make it conditional like below?
```suggestion
      {
        [[ $TFLINT_INIT ]] && {
          common::colorify "green" "Command 'tflint --init' successfully done:"
          echo -e "${TFLINT_INIT}\n\n\n"
        }
      }
```

MaxymVlasov: It does not produce empty output, it shows a list of all installed plugins and writes that they are up-to-date.
Also, it will go to stdout ONLY if something goes wrong.

I prefer to see all operations/changes that were made by hook if something goes wrong - that can be useful for debugging.

MaxymVlasov: I.e. it can be used as hint when it will fail in CI but still works locally

MaxymVlasov: that's can happen, because TFlint plugins have compatibility matrix with tflint

yermulnik: Make sense. Agree. Thanks.

PR #386, opened 2022-05-18
feat: When a config file is given, do not specify formatter on cli (terraform_docs)

acodeninja:
- [x] This PR fixes a bug.

The change is to the `terraform_docs` hook. Currently this hook will ignore the formatter set in any specified config file, defaulting to the `md` formatter. On our project we use the `markdown document` formatter but this is ignored by the hook. I consider this a bug but thought a PR would be better than a new issue for you.

### How can we test changes

Run:

```shellscript
$ mkdir test-tf-docs-pre-commit
$ cd test-tf-docs-pre-commit
$ git init
$ pre-commit install
```

Create the following files.

`.terraform-docs.yml`
```yaml
formatter: toml
```

`.pre-commit-config.yaml`
```yaml
repos:
  - repo: https://github.com/acodeninja/pre-commit-terraform.git
    rev: v1.71.0
    hooks:
      - id: terraform_docs
        args:
          - --args=--config=.terraform-docs.yml
```

`main.tf`
```hcl
variable "some-input" {
  description = "An input variable"
  type        = string
}

output "some-output" {
  value = var.some-input
}
```

`README.md`
```markdown
<!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->

<!-- END OF PRE-COMMIT-TERRAFORM DOCS HOOK -->
```

Then run `pre-commit run --all-files`.

You will see the output in the readme after running will be:

```markdown
<!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->
## Requirements

No requirements.

## Providers

No providers.

## Modules

No modules.

## Resources

No resources.

## Inputs

| Name | Description | Type | Default | Required |
|------|-------------|------|---------|:--------:|
| <a name="input_some-input"></a> [some-input](#input\_some-input) | An input variable | `string` | n/a | yes |

## Outputs

| Name | Description |
|------|-------------|
| <a name="output_some-output"></a> [some-output](#output\_some-output) | n/a |
<!-- END OF PRE-COMMIT-TERRAFORM DOCS HOOK -->
```

This is not the configured and desired format according to the `.terraform-docs.yml` configuration.

Now, change the `.pre-commit-config.yaml` file to the following:

```yaml
repos:
  - repo: https://github.com/acodeninja/pre-commit-terraform.git
    rev: 855f9c52ed0e717dd953a3cbe445bfee1cd17662
    hooks:
      - id: terraform_docs
        args:
          - --args=--config=.terraform-docs.yml
```

And run `pre-commit run --all-files` again.

You will now see the readme file contents is as expected from the configuration of terraform docs.

```markdown
<!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->
header = ""
footer = ""
modules = []
providers = []
requirements = []
resources = []

[[inputs]]
  name = "some-input"
  type = "string"
  description = "An input variable"
  required = true
  [inputs.default]

[[outputs]]
  name = "some-output"
  description = ""
<!-- END OF PRE-COMMIT-TERRAFORM DOCS HOOK -->
```

### DIFF
--- a/hooks/terraform_docs.sh
+++ b/hooks/terraform_docs.sh
@@ -138,7 +138,7 @@ function terraform_docs {
   #
   # Override formatter if no config file set
   #
-  [[ "$args" != *"--config"* ]] && local tf_docs_formatter="md"
+  [[ ! $args =~ (^|[[:space:]])--config=[^[:space:]]+ ]] && local tf_docs_formatter="md"
 
   local dir_path
   for dir_path in $(echo "${paths[*]}" | tr ' ' '\n' | sort -u); do
--- a/hooks/terraform_docs.sh
+++ b/hooks/terraform_docs.sh
@@ -135,6 +135,15 @@ function terraform_docs {
     esac
   done
 
+  #
+  # Decide formatter to use
+  #
+  local tf_docs_formatter="md"
+  if [[ "$args" == *"--config"* ]]; then
+    # Allow config file to specify formatter
+    tf_docs_formatter=""
+  fi
+
   local dir_path
   for dir_path in $(echo "${paths[*]}" | tr ' ' '\n' | sort -u); do
     dir_path="${dir_path//__REPLACED__SPACE__/ }"
@@ -181,7 +190,7 @@ function terraform_docs {
 
     if [[ "$terraform_docs_awk_file" == "0" ]]; then
       # shellcheck disable=SC2086
-      terraform-docs md $args ./ > "$tmp_file"
+      terraform-docs $tf_docs_formatter $args ./ > "$tmp_file"
     else
       # Can't append extension for mktemp, so renaming instead
       local tmp_file_docs
@@ -192,7 +201,7 @@ function terraform_docs {
 
       awk -f "$terraform_docs_awk_file" ./*.tf > "$tmp_file_docs_tf"
       # shellcheck disable=SC2086
-      terraform-docs md $args "$tmp_file_docs_tf" > "$tmp_file"
+      terraform-docs $tf_docs_formatter $args "$tmp_file_docs_tf" > "$tmp_file"
       rm -f "$tmp_file_docs_tf"
     fi
 
--- a/hooks/terraform_docs.sh
+++ b/hooks/terraform_docs.sh
@@ -138,7 +138,7 @@ function terraform_docs {
   #
   # Override formatter if no config file set
   #
-  [[ ! $args =~ (^|[[:space:]])--config=[^[:space:]]+ ]] && local tf_docs_formatter="md"
+  [[ "$args" != *"--config="* ]] && local tf_docs_formatter="md"
 
   local dir_path
   for dir_path in $(echo "${paths[*]}" | tr ' ' '\n' | sort -u); do
--- a/hooks/terraform_docs.sh
+++ b/hooks/terraform_docs.sh
@@ -141,7 +141,7 @@ function terraform_docs {
   local tf_docs_formatter="md"
   if [[ "$args" == *"--config"* ]]; then
     # Allow config file to specify formatter
-    tf_docs_formatter=""
+    unset tf_docs_formatter
   fi
 
   local dir_path
--- a/hooks/terraform_docs.sh
+++ b/hooks/terraform_docs.sh
@@ -136,13 +136,9 @@ function terraform_docs {
   done
 
   #
-  # Decide formatter to use
+  # Override formatter if no config file set
   #
-  local tf_docs_formatter="md"
-  if [[ "$args" == *"--config"* ]]; then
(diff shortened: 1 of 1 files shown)


hooks/terraform_docs.sh
```diff
@@ -135,6 +135,15 @@ function terraform_docs {
     esac
   done
 
+  #
+  # Decide formatter to use
+  #
+  local tf_docs_formatter="md"
+  if [[ "$args" == *"--config"* ]]; then
+    # Allow config file to specify formatter
+    tf_docs_formatter=""
```
yermulnik, line 144:
```suggestion
    unset tf_docs_formatter
```

hooks/terraform_docs.sh
```diff
@@ -135,6 +135,15 @@ function terraform_docs {
     esac
   done
 
+  #
+  # Decide formatter to use
+  #
+  local tf_docs_formatter="md"
+  if [[ "$args" == *"--config"* ]]; then
```
yermulnik, line 142: Thanks for the contribution. Good point.
Can this `if` conditional be re-worked by adding check for `--config` arg into `case` block on lines 125-135 and setting some var conditionally, e.g. `config_file="true"`, and here instead of `if` statement do something like `[[ $config_file != "true" ]] && local tf_docs_formatter="md"`? Thanks.

acodeninja: So I had a look at the case block which appears to be looking at the values in the `$hook_config` variable. This doesn't actually contain details passed in by the args property, those are contained in the `$args` variable. I'm happy to alter the PR, but this doesn't appear to be the best way. 

Would something like this be ok?

```shellscript
  #
  # Override formatter if no config file set
  #
  [[ "$args" != *"--config"* ]] && local tf_docs_formatter="md"
```

I can of course break this up a little and do something like:

```shellscript
  #
  # Override formatter if no config file set
  #
  [[ "$args" == *"--config"* ]] && local has_config_file="true"

  [[ "$has_config_file" != "true" ]] && local tf_docs_formatter="md"
```

Let me know what you think.

yermulnik: Oh, my bad. I indeed misread that part of code 🤦🏻
Yep `[[ "$args" != *"--config"* ]] && local tf_docs_formatter="md"` looks good to me. Thanks.

acodeninja: It's no worries, that was actually my first solution this morning.

acodeninja: On the short code I agree, however this currently isn't supported in any case. See line 19:

```
ARGS=${ARGS[*]/--config=/--config=$(pwd)\/}
```

The absolute path resolution only kicks in if the full option `--config` is set. I can include this fix in my PR, but it feels out of scope. Happy to alter as required though.

yermulnik: Oh, so `=` sign is allowed. Looks like we need to improve code on line 19 since it doesn't cover short options and space instead of equal sign 🤔 Though this indeed is out of scope of your PR. Let's pitch upon `[[ ! $args =~ (^|[[:space:]])--config=[^[:space:]]+ ]] && local tf_docs_formatter="md"` so that these two are aligned?

acodeninja: I've swapped out for the regex approach so hopefully good to go.

yermulnik (approved):
LGTM
Anyone has capacity to test this with Bash v3 please?

MaxymVlasov:
Maybe we can just pin the bash version used in Mac here
https://github.com/antonbabenko/pre-commit-terraform/blob/51147f1712117fa4d2c902cf6c9066c4fc8cafab/Dockerfile#L177

yermulnik:
> Anyone has capacity to test this with Bash v3 please?

So the "(..|..)" thing doesn't work in Bash v3. @acodeninja Looks like `[[ "$args" != *"--config="* ]]` is our best choice. Apologies for the fuss 🤷🏻

yermulnik (approved):

yermulnik:
> Maybe we can just pin the bash version used in Mac here

I just build bash3 locally which appeared to be quickest wat for me =)

MaxymVlasov:
> > Maybe we can just pin the bash version used in Mac here
> 
> I just built bash3 locally which appeared to be quickest way for me =) And I'm having an objection on pinning Docker to one version of Bash, while allowing and supporting another version of Bash outside Docker.

Well, not possible to install bash 3 via apk add. The oldest bash in Alpine is 4.3
https://pkgs.alpinelinux.org/packages?name=bash&branch=v3.3

MaxymVlasov (approved):
Tested, works fine

antonbabenko:
This PR is included in version 1.72.0 :tada:

PR #392, opened 2022-05-27
docs: Updated docs for infracost_breakdown to match Infracost 0.10

antonbabenko:
Infracost 0.10 has some changes in the names of the flags.

Ref: https://twitter.com/AliKhajeh/status/1530081914865868802

### DIFF
--- a/README.md
+++ b/README.md
@@ -322,14 +322,14 @@ Unlike most other hooks, this hook triggers once if there are any changed files 
     <!-- markdownlint-disable-next-line no-inline-html -->
     </details>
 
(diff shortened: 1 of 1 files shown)


yermulnik (approved):
Is this backward compatible with `infracost` versions prior to `v0.10`?

yermulnik:
> I don't think so. Infracost 0.10 has been released 3 days ago.

Maybe revert this change then and add it back as a supplemental note for Infracost 0.10 specifically? 🤔

antonbabenko:
I don't think it is necessary because we always expect (and recommend) users to update their tools if something seems off. This change is mentioning one of an argument but there are several more that users may be using (depending on the version of the tool they have installed).

antonbabenko:
This PR is included in version 1.72.2 :tada:

PR #402, opened 2022-06-30
chore: Reuse logic from common functions

nshenry03:
From @MaxymVlasov in https://github.com/antonbabenko/pre-commit-terraform/pull/401#discussion_r903065694:
> Heh, you found code written in Paleolithic.
>
> Let's try reuse logic that used in common functions here:
>
> https://github.com/antonbabenko/pre-commit-terraform/blob/master/hooks/_common.sh#L178
>
> Absolutely have no idea why it was written in so complex way

Put an `x` into the box if that apply:

- [ ] This PR introduces breaking change.
- [ ] This PR fixes a bug.
- [ ] This PR adds new functionality.
- [ ] This PR enhances existing functionality.

### Description of your changes

Just doing as @MaxymVlasov asked in https://github.com/antonbabenko/pre-commit-terraform/pull/401#discussion_r903065694

### How can we test changes

None. It's a fairly simple and straightforward change.

PR #404, opened 2022-07-04
chore: Rewrite `terraform_validate` to `common::per_dir_hook`

MaxymVlasov:
That's one of many PRs to the `pre-release` branch, each one will depend on the previous one.
That is the way to include all that info in one release...

Also, close #402.

### DIFF
--- a/hooks/terraform_validate.sh
+++ b/hooks/terraform_validate.sh
@@ -36,6 +36,7 @@ function main {
 # global variables with appropriate values
 # Globals (init and populate):
 #   ARGS (array) arguments that configure wrapped tool behavior
+#   HOOK_CONFIG (array) arguments that configure hook behavior
 #   INIT_ARGS (array) arguments to `terraform init` command
 #   ENVS (array) environment variables that will be used with
 #     `terraform` commands
@@ -54,6 +55,11 @@ function parse_cmdline_ {
       -a | --args)
         shift
         ARGS+=("$1")
+        shift
+        ;;
+      -h | --hook-config)
+        shift
+        HOOK_CONFIG+=("$1;")
         shift
         ;;
       -i | --init-args)
--- a/hooks/terraform_validate.sh
+++ b/hooks/terraform_validate.sh
@@ -17,7 +17,18 @@ function main {
   common::initialize "$SCRIPT_DIR"
   parse_cmdline_ "$@"
   common::parse_and_export_env_vars
-  terraform_validate_
+
+  # Export provided env var K/V pairs to environment
+  local var var_name var_value
+  for var in "${ENVS[@]}"; do
+    var_name="${var%%=*}"
+    var_value="${var#*=}"
+    # shellcheck disable=SC2086
+    export $var_name="$var_value"
+  done
+
+  # shellcheck disable=SC2153 # False positive
+  common::per_dir_hook "${ARGS[*]}" "$HOOK_ID" "${FILES[@]}"
 }
 
 #######################################################################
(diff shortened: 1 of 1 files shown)


hooks/terraform_validate.sh
```diff
@@ -17,14 +17,26 @@ function main {
...
-  terraform_validate_
+
+  # Export provided env var K/V pairs to environment
+  local var var_name var_value
+  for var in "${ENVS[@]}"; do
+    var_name="${var%%=*}"
+    var_value="${var#*=}"
+    # shellcheck disable=SC2086
+    export $var_name="$var_value"
+  done
```
MaxymVlasov, line 28: That's will be moved to separate function later

hooks/terraform_validate.sh
```diff
@@ -45,6 +57,11 @@ function parse_cmdline_ {
         ARGS+=("$1")
         shift
         ;;
+      -h | --hook-config)
```
MaxymVlasov, line 60: `parse_cmdline_` will be merged in `common::parse_cmdline` later

hooks/terraform_validate.sh
```diff
@@ -65,88 +82,53 @@ function parse_cmdline_ {
...
+    init_output=$(terraform init -backend=false "${INIT_ARGS[@]}" 2>&1)
+    exit_code=$?
+
+    if [ $exit_code -ne 0 ]; then
+      common::colorify "yellow" "'terraform init' failed, 'terraform validate' skipped: $dir_path"
+      echo "$init_output"
+      return $exit_code
     fi
-  done
+  fi
```
MaxymVlasov, line 118: Will be moved to `common::terraform_init` function

MaxymVlasov (commented):
Leave some notes about future improvements, to avoid extra attention to them in this PR

hooks/terraform_validate.sh
```diff
@@ -65,88 +82,53 @@ function parse_cmdline_ {
...
+  local init_output
+  local validate_output
+
+  if [ ! -d .terraform ]; then
+    init_output=$(terraform init -backend=false "${INIT_ARGS[@]}" 2>&1)
+    exit_code=$?
+
+    if [ $exit_code -ne 0 ]; then
+      common::colorify "yellow" "'terraform init' failed, 'terraform validate' skipped: $dir_path"
+      echo "$init_output"
```
yermulnik, line 115: should the `echo` use `-e` opt to interpret escape sequences or `terraform init` should use `-no-color`?

yermulnik (approved):

hooks/terraform_validate.sh
MaxymVlasov: Ok, will replace `echo` with  `common::colorify`

yermulnik: The other `echo` used to output `terraform validate` then needs this change too.
Though... I'd better keep native TF output colorize specs to keep it usual and recognized by users.

yermulnik: Especially given that there's not `-no-color` passed to terraform, its escape sequences may (and probably will) get mixed up with sequences used by `common::colorify` and the output may become vomit of all the colours of the rainbow 🤣

MaxymVlasov: Just tested - terraform ignore any color instructions specified to pre-commit or by `common::colorify`

Will be done in another PR (need to disable that across all hooks)

hooks/terraform_validate.sh
MaxymVlasov: Done in #406

hooks/terraform_validate.sh
MaxymVlasov: Done in #408

hooks/terraform_validate.sh
MaxymVlasov: Done in #408

hooks/terraform_validate.sh
MaxymVlasov: Done in #409

PR #422, opened 2022-08-02
chore: Fix `terraform_validate` exclude example

### LINKED ISSUE #421: exclusions are install dependent and documented example for terraform_valudate is incorrect
The example given for running the validate hook against non root modules advises to create a test root module and to exclude the top level directory as follows:

````
- id: terraform_validate
  exclude: '[^/]+$'
`````

This does not work in my ubuntu environment. Looking into the code I see it is using grep -e, however that is environment dependent. So for eample

`git ls-files | sort | grep -e "" | grep -v -e "[^/]+$"`
returns all files

adding the required escape
`git ls-files | sort | grep -e "" | grep -v -e "[^/]\+$"`
returns no files because all files have a non slash character at the end

correct pattern is to anchor to the start of the string
`git ls-files | sort | grep -e "" | grep -v -e "^[^/]\+$"`
returns only the files in a sub directory

There is an option to ue -E on grep so this works too
`git ls-files | sort | grep -e "" | grep -v -E "^[^/]+$"`
however that would be a change to pre-commit

All that said, I still can't get exclude to work from within the pre commit yaml - maybe some further escaping at play that I haven't got my head around yet.

What does work is to turn it around and use files instead of exclude
````
- id: terraform_validate
  files: tests/validate
`````

Maybe a note in the documentation and/or perhaps show using files instead of exclude would help other people.
Thanks.

MaxymVlasov:
Fixes #421

### DIFF
--- a/README.md
+++ b/README.md
@@ -677,7 +677,7 @@ Example:
    This is a [known issue](https://github.com/hashicorp/terraform/issues/28490) with Terraform and how providers are initialized in Terraform 0.15 and later. To work around this you can add an `exclude` parameter to the configuration of `terraform_validate` hook like this:
    ```yaml
    - id: terraform_validate
-     exclude: '[^/]+$'
+     exclude: '^[^/]+$'
    ```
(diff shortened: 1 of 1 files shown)


antonbabenko (approved):

antonbabenko:
This PR is included in version 1.74.2 :tada:

PR #424, opened 2022-08-04
chore: Add new deps to check to issue template

MaxymVlasov:


.github/ISSUE_TEMPLATE/bug_report_local_install.md
```diff
@@ -87,6 +87,8 @@ terragrunt --version                      2>/dev/null || echo "terragrunt SKIPPE
 echo -n "terrascan " && terrascan version 2>/dev/null || echo "terrascan SKIPPED"
 tflint --version                          2>/dev/null || echo "tflint SKIPPED"
 echo -n "tfsec " && tfsec --version       2>/dev/null || echo "tfsec SKIPPED"
+echo -n "tfupdate " && tfupdate --version 2>/dev/null || echo "tfupdate SKIPPED"
+echo -n "hcledit " && hcledit version     2>/dev/null ||  echo "hcledit SKIPPED"
```
MaxymVlasov, line 91:
```suggestion
echo -n "hcledit " && hcledit version     2>/dev/null || echo "hcledit SKIPPED" 
```

antonbabenko (approved):

.github/ISSUE_TEMPLATE/bug_report_local_install.md
```diff
@@ -75,7 +75,7 @@ Linux DESKTOP-C7315EF 5.4.72-microsoft-standard-WSL2 #1 SMP Wed Oct 28 23:40:43
 
 <!--  For check all needed version run next script:
 
-$0 << EOF
+bash << EOF
```
yermulnik, line 78: To properly handle errors from commands which use pipes, we should ensure `pipefail` is turned on so that exit code of chain of commands is the one of last non-successful commands:
```suggestion
bash -o pipefail << EOF
```

Also if commands below that use pipes to filter output should not output errors, we need to wrap the chain of commands into curly brackets like this:
```
{ terraform --version | head -n 1; } 2>/dev/null || echo "terraform SKIPPED"
```
Also the same should better be done for commands with control operator (`&&` in the case of below code block) so that error output of any of cmds in chain is suppressed and so that control `||` operator takes place for the chain of commands, rather than for the one before previous `&&` operator in ternary.

As always: needs testing on ancient versions of Bash 😉

MaxymVlasov: pre-commit-terraform will not work without `terraform`, `pre-commit`, and `git`, so let's says that is not a bug - it's a feature :D

yermulnik: Yep, that's obvious 🤣 My comment was about proper use and syntax of bash code snippet to — again — properly handle exit codes and errors 😺 Not essential though, though looks like a low quality code created by low quality engineers like ourselves 🤣😜

antonbabenko:
This PR is included in version 1.74.2 :tada:

PR #436, opened 2022-10-03
feat: Add support for version constraints in `tfupdate`

MaxymVlasov:
Put an `x` into the box if that apply:

- [ ] This PR introduces breaking change.
- [ ] This PR fixes a bug.
- [x] This PR adds new functionality.
- [ ] This PR enhances existing functionality.

### Description of your changes

Fixes #388 

### How can we test changes

* `pre-commit-config.yaml`:

```yaml
repos:
- repo: https://github.com/antonbabenko/pre-commit-terraform
  rev: 8e7a029eb1f796f1ff60ad8e07b911466381cdd1
  hooks:
    - id: tfupdate
      name: Autoupdate AWS provider versions
      args:
        - --args=--version '~> 4.2.0'
        - --args=provider aws

    - id: tfupdate
      name: Autoupdate Helm provider versions
      args:
        - --args=provider helm
    - id: tfupdate
      name: Update Google provider to specified version
      args:
        - --args=provider google
        - --args=--version ">= 4.18.0, < 5"
    - id: tfupdate
      name: Update Terraform provider to specified version
      args:
        - --args=terraform
        - --args=--version "> 1.1.8"
```

`versions.tf`

```tf
terraform {
  required_version = ">= 0.14"
  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = ">= 3.0.0"
    }
    helm = {
      source  = "hashicorp/helm"
      version = "2.6.0"
    }
    google = {
      source  = "google/google"
      version = "2.7.0"
    }

  }
}
```

Issue #388, opened 2022-05-25
tfupdate version constraints


chris3ware:
### What problem are you facing?

I am trying to integrate tfupdate into my pre-commit configuration. It works perfectly for exact versions i.e:

```yaml
          - --args=--version 0.7.1
```

tfupdate also supports terraform version constraints - but I am struggling to use this functionality with pre-commit so I wondered if it was supported or if I was doing something wrong? 

My attempted configuration looks like this:

```yaml
      - id: tfupdate
        name: Terraform provider version SOPS
        args:
          - --args=provider sops
          - --args=--version ">= 0.7.1"
```

The error returned when running `pre-commit run` is this:

```zsh
Terraform provider version SOPS..........................................Failed
- hook id: tfupdate
- exit code: 1

The command expects 2 arguments, but got 3
```

The associated tfupdate command (`tfupdate provider --version ">= 0.7.1" --recursive sops .`) run from the CLI works as expected. 

### How could pre-commit-terraform help solve your problem?

Should terraform pre-commit accept version constraint arguments for tfupdate? A good use case for this is required_providers in a module. For example for current AWS provider looks like this:

```hcl
terraform {
  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = ">= 3.71.0, < 4.0.0"
    }
}
```

MaxymVlasov:
This issue has been resolved in version 1.76.0 🎉

PR #437, opened 2022-10-04
feat: Add support for version constraints in `tfupdate`

MaxymVlasov:
Replace for #436 ?
Close #436 #388

- [x] shellcheck violation
- [x] fix other hooks, by adding `"${#ARGS[@]}"` and change `per_dir_hook_unique_part` args
- [x] fix pre-commit --all
- [x] fix the PASSED case, when args in the wrong order
```
      args:
        - --args=--version '~> 4.2.0'
        - --args=provider aws
```

### DIFF
--- a/hooks/_common.sh
+++ b/hooks/_common.sh
@@ -164,17 +164,21 @@ function common::is_hook_run_on_whole_repo {
 # 3. Complete hook execution and return exit code
 # Arguments:
 #   hook_id (string) hook ID, see `- id` for details in .pre-commit-hooks.yaml file
-#   args (string with array) arguments that configure wrapped tool behavior
+#   args_array_length (integer) Count of arguments in args array.
+#   args (array) arguments that configure wrapped tool behavior
 #   files (array) filenames to check
 #######################################################################
 function common::per_dir_hook {
   local -r hook_id="$1"
-  # Expand args to a true array
-  local -a args=()
-  while read -r -d '' ARG; do
-    args+=("$ARG")
-  done < <(echo "$2" | xargs printf '%s\0')
+  local -i args_array_length=$2
   shift 2
+  local -a args=()
+  # Expand args to a true array.
+  # Based on https://stackoverflow.com/a/10953834
+  while ((args_array_length-- > 0)); do
+    args+=("$1")
+    shift
+  done
   # assign rest of function's positional ARGS into `files` array,
   # despite there's only one positional ARG left
   local -a -r files=("$@")
--- a/hooks/_common.sh
+++ b/hooks/_common.sh
@@ -164,21 +164,17 @@ function common::is_hook_run_on_whole_repo {
 # 3. Complete hook execution and return exit code
 # Arguments:
 #   hook_id (string) hook ID, see `- id` for details in .pre-commit-hooks.yaml file
-#   args_array_length (integer) Count of arguments in args array.
-#   args (array) arguments that configure wrapped tool behavior
+#   args (string with array) arguments that configure wrapped tool behavior
 #   files (array) filenames to check
 #######################################################################
 function common::per_dir_hook {
   local -r hook_id="$1"
-  local -i args_array_length=$2
-  shift 2
+  # Expand args to a true array
   local -a args=()
-  # Expand args to a true array.
-  # Based on https://stackoverflow.com/a/10953834
-  while ((args_array_length-- > 0)); do
-    args+=("$1")
-    shift
-  done
+  while read -r -d '' ARG; do
+    args+=("$ARG")
+  done < <(echo "$2" | xargs printf '%s\0')
+  shift 2
   # assign rest of function's positional ARGS into `files` array,
   # despite there's only one positional ARG left
   local -a -r files=("$@")
--- a/hooks/_common.sh
+++ b/hooks/_common.sh
@@ -129,7 +129,7 @@ function common::parse_and_export_env_vars {
 #######################################################################
 function common::is_hook_run_on_whole_repo {
   local -r hook_id="$1"
-  shift 1
+  shift
   local -a -r files=("$@")
   # get directory containing `.pre-commit-hooks.yaml` file
   local -r root_config_dir="$(dirname "$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)")"
@@ -170,16 +170,17 @@ function common::is_hook_run_on_whole_repo {
 #######################################################################
 function common::per_dir_hook {
   local -r hook_id="$1"
-  declare -i args_array_length=$2
+  local -i args_array_length=$2
   shift 2
-  declare -a args=()
+  local -a args=()
   # Expand args to a true array.
   # Based on https://stackoverflow.com/a/10953834
   while ((args_array_length-- > 0)); do
     args+=("$1")
     shift
   done
-
+  # assign rest of function's positional ARGS into `files` array,
+  # despite there's only one positional ARG left
   local -a -r files=("$@")
 
   # check is (optional) function defined
--- a/hooks/_common.sh
+++ b/hooks/_common.sh
@@ -168,8 +168,8 @@ function common::is_hook_run_on_whole_repo {
 #   files (array) filenames to check
 #######################################################################
 function common::per_dir_hook {
-  local -r args="$1"
-  local -r hook_id="$2"
+  local -r hook_id="$1"
+  local -r args="$2"
   shift 2
   local -a -r files=("$@")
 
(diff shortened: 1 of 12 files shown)


hooks/_common.sh
```diff
@@ -163,21 +163,30 @@ function common::is_hook_run_on_whole_repo {
...
 function common::per_dir_hook {
-  local -r args="$1"
-  local -r hook_id="$2"
+  local -r hook_id="$1"
+  declare -i args_array_length=$2
   shift 2
+  declare -a args=()
+  # Expand args to a true array.
+  # Based on https://stackoverflow.com/a/10953834
+  while ((args_array_length-- > 0)); do
```
yermulnik, line 178: Would a simpler code work here? This also makes no need for `args_array_length` var.
```suggestion
  while [[ $@ ]]; do
```
On the other hand the whole loop looks quite like simple `args=("$@")` 🤔

hooks/tfupdate.sh
```diff
@@ -15,26 +15,102 @@ function main {
...
+#   args (array) arguments that configure wrapped tool behavior
 # Outputs:
 #   If failed - print out hook checks status
 #######################################################################
 function per_dir_hook_unique_part {
-  local -r args="$1"
   # shellcheck disable=SC2034 # Unused var.
-  local -r dir_path="$2"
+  local -r dir_path="$1"
+  shift 1
```
yermulnik, line 33: it defaults to `1` if no args provided
```suggestion
  shift
```

hooks/tfupdate.sh
```diff
@@ -15,26 +15,102 @@ function main {
...
+  declare -a -r args=("$@")
```
yermulnik, line 34: should this be `local -a -r args=("$@")`?

hooks/tfupdate.sh
```diff
@@ -15,26 +15,32 @@ function main {
...
+
+  local expand_args=()
```
yermulnik, line 36: For consistency and precision:
```suggestion
  local -a expand_args=()
```

hooks/tfupdate.sh
```diff
@@ -48,12 +54,16 @@ function per_dir_hook_unique_part {
 #   args (string with array) arguments that configure wrapped tool behavior
 #######################################################################
 function run_hook_on_whole_repo {
-  local -r args="$1"
+  declare -a -r args=("$@")
+
+  local expand_args=()
```
yermulnik, line 59: For consistency and precision:
```suggestion
  local -a expand_args=()
```

hooks/tfupdate.sh
```diff
@@ -48,12 +54,16 @@ function per_dir_hook_unique_part {
 #   args (string with array) arguments that configure wrapped tool behavior
 #######################################################################
 function run_hook_on_whole_repo {
-  local -r args="$1"
+  declare -a -r args=("$@")
```
yermulnik, line 57: Should this be `local -a -r args=("$@")` or is it non-local var?

hooks/_common.sh
yermulnik: Oh, I seem to think I understand why `args_array_length` is needed.
Though given `common::per_dir_hook "$HOOK_ID" "${ARGS[*]}" "${FILES[@]}"`, which is used across this PR, the 2nd ARG to this current function has ARGS array rather than number of its elements — is this expected? So it's either `common::per_dir_hook "$HOOK_ID" "${#ARGS[*]}" "${ARGS[*]}" "${FILES[@]}"`, or I am missing something essential 🤔 (probably it's right about time to leave this for tomorrow 😛)

yermulnik: Okay, I couldn't leave it for tomorrow.
This `while` thing doesn't work as expected for me, and given `common::per_dir_hook "$HOOK_ID" "${ARGS[*]}" "${FILES[@]}"` this should be that all same thing from our discussion in Slack:
```
  while read -r -d '' ARG; do
    args+=("$ARG")
  done < <(echo "$1" | xargs printf '%s\0')
  shift
```

hooks/terraform_tfsec.sh
```diff
@@ -13,17 +13,17 @@ function main {
...
-  ARGS=${ARGS[*]/__GIT_WORKING_DIR__/$(pwd)\/}
+  for i in "${!ARGS[@]}"; do
+    ARGS[i]=${ARGS[i]/__GIT_WORKING_DIR__/$(pwd)\/}
+  done
 
   # Suppress tfsec color
   if [ "$PRE_COMMIT_COLOR" = "never" ]; then
     # shellcheck disable=SC2178,SC2128 # It's the simplest syntax for that case
-    ARGS+=" --no-color"
+    ARGS+=("--no-color")
```
yermulnik, line 23: Is comment above still relevant as of the change to this line?

hooks/_common.sh
```diff
@@ -163,21 +163,30 @@ function common::is_hook_run_on_whole_repo {
...
+    args+=("$1")
+    shift
+  done
+
   local -a -r files=("$@")
```
yermulnik, line 183: For clarity (otherwise we need to be more picky on which positional ARG should be considered a list of files passed in to this function):
```suggestion
  # assign rest of function's positional ARGS into `files` array,
  # despite there's only one positional ARG left
  local -a -r files=("$@")
```

hooks/_common.sh
```diff
@@ -163,21 +163,30 @@ function common::is_hook_run_on_whole_repo {
...
 #   hook_id (string) hook ID, see `- id` for details in .pre-commit-hooks.yaml file
+#   args_array_length (integer) Count of arguments in args array.
+#   args (array) arguments that configure wrapped tool behavior
 #   files (array) filenames to check
 #######################################################################
 function common::per_dir_hook {
-  local -r args="$1"
-  local -r hook_id="$2"
+  local -r hook_id="$1"
+  declare -i args_array_length=$2
```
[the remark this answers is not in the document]
MaxymVlasov: As I read, `declare` in fucntion automaically marc vars as local, except you specify `-g`. But for consistency, let me rewrite it to `local`, especially after you mention that `-i` available for `local` too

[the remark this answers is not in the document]
yermulnik: Yeah, you're right, in a function it makes var to be local, I keep forgetting this. And that's why I lean to use `local` in functions, sort of more clear to understand by dummies like me 🤣 And yep `local` accepts all options which are acceptable by `declare`.

hooks/_common.sh
MaxymVlasov: No idea why, but the suggested solution does not work here 🤔 

[`7866b64` (#437)](https://github.com/antonbabenko/pre-commit-terraform/pull/437/commits/7866b64c3a1644bfd1049013329f381e379da190)

I try to replace `common::per_dir_hook "${ARGS[*]}"` with `common::per_dir_hook "${ARGS[@]}"` but got a weird error
```
dirname: unrecognized option '--version__REPLACED__SPACE__">__REPLACED__SPACE__1.1.8"'
Try 'dirname --help' for more information.
```

Also, no luck with changing `per_dir_hook_unique_part "${args[@]}"` to `per_dir_hook_unique_part "${args[*]}"`

MaxymVlasov: >Is the __REPLACED__SPACE__ to be replaced with actual space later down the code? 

Should be. Firstly, it replaces spaces with placeholders and then placeholders with spaces. Check file for `__REPLACED__SPACE__ `

yermulnik: Nah, I mean is this expected that at this place the `__REPLACED__SPACE__` isn't replaced with spaces yet? This may influence how we should be parsing array elements since elements values are not space-separated string, but `__REPLACED__SPACE__`-separated string.

hooks/terraform_docs.sh
```diff
@@ -12,12 +12,12 @@ function main {
   common::parse_cmdline "$@"
   common::export_provided_env_vars "${ENV_VARS[@]}"
   common::parse_and_export_env_vars
-  # Support for setting relative PATH to .terraform-docs.yml config.
-  # shellcheck disable=SC2178 # It's the simplest syntax for that case
-  ARGS=${ARGS[*]/--config=/--config=$(pwd)\/}
-  # shellcheck disable=SC2128 # It's the simplest syntax for that case
+  # Support for setting PATH to repo root.
+  for i in "${!ARGS[@]}"; do
+    ARGS[i]=${ARGS[i]/__GIT_WORKING_DIR__/$(pwd)\/}
```
MaxymVlasov, line 17:
```suggestion
    ARGS[i]=${ARGS[i]/--config=/--config=$(pwd)\/}
```

yermulnik (approved):
Black magic 🤣
Great job hardworking on this, Max. Thank you! 👍🏻

Issue #203, opened 2021-06-03
hook terraform validate fails 


loque1:
Hi. I upgraded to terraform 0.15.4 and the terraform_validate hook fails with the following
```bash
Terraform validate.......................................................Failed
- hook id: terraform_validate
- exit code: 1

Validation failed: iac/common
╷
│ Warning: Experimental feature "module_variable_optional_attrs" is active
│ 
│   on terraform.tf line 13, in terraform:
│   13:   experiments = [module_variable_optional_attrs]
│ 
│ Experimental features are subject to breaking changes in future minor or
│ patch releases, based on feedback.
│ 
│ If you have feedback on the design of this feature, please open a GitHub
│ issue to discuss it.
╵
╷
│ Error: Could not load plugin
│ 
│ 
│ Plugin reinitialization required. Please run "terraform init".
│ 
│ Plugins are external binaries that Terraform uses to access and manipulate
│ resources. The configuration provided requires plugins which can't be
│ located,
│ don't satisfy the version constraints, or are otherwise incompatible.
│ 
│ Terraform automatically discovers provider requirements from your
│ configuration, including providers used in child modules. To see the
│ requirements and constraints, run "terraform providers".
│ 
│ 2 problems:
│ 
│ - failed to instantiate provider "registry.terraform.io/hashicorp/azuread"
│ to obtain schema: unknown provider
│ "registry.terraform.io/hashicorp/azuread"
│ - failed to instantiate provider "registry.terraform.io/hashicorp/azurerm"
│ to obtain schema: unknown provider
│ "registry.terraform.io/hashicorp/azurerm"
│ 
╵
```
If I run the command outside of pre-commit in the same directory it all passes fine
```bash
╷
│ Warning: Experimental feature "module_variable_optional_attrs" is active
│ 
│   on terraform.tf line 13, in terraform:
│   13:   experiments = [module_variable_optional_attrs]
│ 
│ Experimental features are subject to breaking changes in future minor or patch releases, based on feedback.
│ 
│ If you have feedback on the design of this feature, please open a GitHub issue to discuss it.
│ 
│ (and 5 more similar warnings elsewhere)
╵
Success! The configuration is valid, but there were some validation warnings as shown above.

```
Any help on this would be ideal

mustafababil:
You can try deleting your `.terraform` folder as you upgraded TF version earlier.

amontalban:
[...]

PR #441, opened 2022-10-21
feat: Add `--retry-once-with-cleanup` to `terraform_validate`

baolsen:

- [x] This PR fixes a bug.

Adds an argument to `terraform_validate` hook to cleanup `.terraform` directory if validate fails. 

    - id: terraform_validate
      args:
        - --hook-config=--retry-once-with-cleanup=true     # Boolean. true or false

This is a workaround to a known issue with `terraform_validate`.
Number 3: https://github.com/antonbabenko/pre-commit-terraform#terraform_validate

Related links below:

Related #224 
Related #301 

I try to address the performance concern raised in #301 by only clearing `.terraform` if validate fails once per directory.

### How can we test changes

Tested with:

```
    rev: '850bb967e2d38a89665f0640467a3d5d2d6f634e'
    hooks:
      - id: terraform_fmt
      - id: terraform_validate
        args:
```

Steps:
1 Run hook on a known-good repo          
2 Enter .terraform folder and corrupt it by deleting some providers
3 Run hook without the flag above. Validation fails.
4 Run hook with the flag above. Validation runs for a while (doing init behind the scenes) and then passes
5 Run hook with the flag above. Validation runs quickly and then passes


yermulnik (commented):
@baolsen Thanks for the contribution.
Overall PR looks good to me, apart from the below points:
1. Do I get it right that having this change in place `terraform_validate` hook will unconditionally remove `.terraform` dir from current working dir and run again even if there are no issues with the content of `.terraform` dir? E.g. when validation fails because of an error with TF code rather than because of a broken content of `.terraform` dir. This doesn't seem to be right. I'm not able to verify `terraform validate` exit codes and output at the moment, hence could you please look up whether TF does output different error code on broken `.terraform` dir content and if not then it may worth to capture its output (this is already in place) and lookup a common to this use case set of error messages to make a decision on whether to re-init TF by removing `.terraform` dir or not.
2. This doesn't look good to unconditionally/inadvertently remove user content, hence could you please update this PR to prompt for user approval to remove `.terraform` dir before proceeding?
3. Since `.terraform` dir may contain files which belong to GIT repo (e.g. cloned TF modules) and since these files may be sort of protected (w/o writable bit attached), it should better force removal via `-f` option to `rm`.

Thanks.

hooks/terraform_validate.sh
```diff
@@ -43,17 +43,57 @@ function per_dir_hook_unique_part {
...
 
-  # pass the arguments to hook
-  validate_output=$(terraform validate "${args[@]}" 2>&1)
+  do_validate
   exit_code=$?
 
+  if [ $exit_code -ne 0 ] && [ "$retry_once_with_cleanup" = true ]; then
+    if [ -d .terraform ]; then
+      # Will only be displayed if validation fails again.
+      common::colorify "yellow" "Validation failed. Re-initialising: $dir_path"
```
MaxymVlasov, line 90: I prefer to display that message in all cases. Please add to README example "`verbose: true` with the related comment about displaying that msg

MaxymVlasov:
>2. This doesn't look good to unconditionally/inadvertently remove user content, hence could you please update this PR to prompt for user approval to remove .terraform dir before proceeding?

Not sure that `pre-commit` is able to add so much interaction. + Then we need a flag that will skip interaction, for jobs run on CI.

Also, I think that specified in 2. functional is too protective and not needed at all

yermulnik:
> Not sure that `pre-commit` is able to add so much interaction.

Mmm, I didn't think of it indeed... Wouldn't `read -p "Nuke .terraform/ dir? (Y/n)"` work?

> Then we need a flag that will skip interaction, for jobs run on CI.

Yep, right! User should be able to explicitly put their consent via hook parameter. Thank for the pointer.

> Also, I think that specified in 2. functional is too protective and not needed at all

It's when we know for sure it was `pre-commit-terraform` who created `.terraform/` dir and not a user who wanted to store there some important data (like e.g. TF module or TF provider that user had to manually download and extract into that dir, or even when user had to make manual changes to any of them for testing, debugging, development, whatever). This is why I think we should not remove user data unconditionally or implicitly but let user explicitly approve such an action.
Maybe you're right and we may go with an optional parameter for this hook, which when set does delete `.terraform/` directory, otherwise hook produces notification message that user has to explicitly set this parameter to make re-init code do its action.

baolsen:
Thanks for the great comments :)

> Do I get it right that having this change in place terraform_validate hook will unconditionally remove .terraform dir from current working dir and run again even if there are no issues with the content of .terraform dir?

Yes, good catch. Whether there is a validation error or a problem with .terraform directory it will re-validate at the moment.

> terraform validate exit codes

Unfortunately it seems to only return exit code 1 for validation errors or an error caused by invalid .terraform config. Thanks for the tip for parsing the error message, I'll definitely look into that now.

> User should be able to explicitly put their consent via hook parameter. Thank for the pointer.

I was not aware that some users would be heavily customising `.terraform` contents.

However it seems a bit strange to need an extra hook parameter for giving consent for the first hook parameter's intended behaviour. If the user has specified the current parameter `--retry-once-with-cleanup` is that enough consent? Or would we prefer to have something like `--retry-with-prompt-for-cleanup` and `--retry-auto-approve`. The second reviewer seemed to think this is overkill, and we only need the current parameter.

Happy to take guidance here. 

I'll do some fixups for the other minor comments.
Thanks!

baolsen:
PR updated with ability to parse `terraform validate` output for specific error summaries, and only delete + retry on those.

The list can be built up over time from other use cases.

yermulnik:
> I was not aware that some users would be heavily customising `.terraform` contents.

Yep, that's more of a rare use case for regular users, though those who develop TF providers might not be happy with implicit removal of `.terraform` dir with the development code. We have to count such uses cases in.

> If the user has specified the current parameter `--retry-once-with-cleanup` is that enough consent?

Yeah, I have to agree with this given Max's opinion. Though hook's behavior should be described in detail in README then, including a warning/notice about implicit/unconditional removal of `.terraform` dir, so that users are aware and agree with that.

> PR updated

Thanks. I'll give it a look in a min.

hooks/terraform_validate.sh
```diff
@@ -43,16 +43,88 @@ function per_dir_hook_unique_part {
...
-  validate_output=$(terraform validate "${args[@]}" 2>&1)
-  exit_code=$?
+  function do_validate {
+    validate_output=$(terraform validate "${args[@]}" 2>&1)
+    exit_code=$?
+    return $exit_code
+  }
+
+  function parse_validate {
+    # Requires jq
```
yermulnik, line 78: Could you please update `jq`'s item in https://github.com/antonbabenko/pre-commit-terraform#1-install-dependencies by adding `terraform_validate` to the list? Thanks.

hooks/terraform_validate.sh
```diff
@@ -43,16 +43,88 @@ function per_dir_hook_unique_part {
...
+    exit_code=$?
+
+    valid="$(jq -rc '.valid' <<< "$validate_output")"
+
+    if [ "$valid" == "true" ]; then
+      return 0
+    fi
+
+    # Pretty-print error information
+    echo "$validate_output" | jq '.diagnostics[]'
```
yermulnik, line 94: For consistency with surrounding code it may worth to use here-string here too:
```suggestion
    jq '.diagnostics[]' <<< "$validate_output"
```

hooks/terraform_validate.sh
```diff
@@ -43,16 +43,88 @@ function per_dir_hook_unique_part {
...
+    # Requires jq
+    local exit_code
+    local validate_output
+    local valid
+    local summary
+
+    validate_output=$(terraform validate -json "${args[@]}" 2>&1)
+    exit_code=$?
+
+    valid="$(jq -rc '.valid' <<< "$validate_output")"
```
yermulnik, line 87: Outer double quotes seem to be extraneous.

hooks/terraform_validate.sh
```diff
@@ -43,16 +43,88 @@ function per_dir_hook_unique_part {
...
+    if [ "$valid" == "true" ]; then
+      return 0
+    fi
+
+    # Pretty-print error information
+    echo "$validate_output" | jq '.diagnostics[]'
+
+    # Parse error message
+    while IFS= read -r error_message; do
+      summary="$(jq -rc '.summary' <<< "$error_message")"
```
yermulnik, line 98: Outer double quotes seem to be extraneous.

hooks/terraform_validate.sh
```diff
@@ -43,16 +43,88 @@ function per_dir_hook_unique_part {
...
+      case $summary in
+        "missing or corrupted provider plugins")
+          return 10
+          ;;
+        "Module source has changed")
+          return 10
+          ;;
+      esac
+    done < <(jq -rc '.diagnostics[]' <<< "$validate_output")
+    return $exit_code
```
yermulnik, line 108: Would it make sense to add an explanatory comment here?
```suggestion

    # Return `terraform validate`'s original exit code for
    # when `$summary` isn't covered by `case` block above
    return $exit_code
```

hooks/terraform_validate.sh
```diff
@@ -43,16 +43,88 @@ function per_dir_hook_unique_part {
...
+    local validate_output
+    local valid
+    local summary
+
+    validate_output=$(terraform validate -json "${args[@]}" 2>&1)
+    exit_code=$?
+
+    valid="$(jq -rc '.valid' <<< "$validate_output")"
+
+    if [ "$valid" == "true" ]; then
```
yermulnik, line 89: Could you please make code more consistent: down the code the comparison against `true` is done w/o double quotes. Either remove double quotes from this conditional's RHS, or add it down the code please.

hooks/terraform_validate.sh
```diff
@@ -43,16 +43,88 @@ function per_dir_hook_unique_part {
...
+    do_validate
+    exit_code=$?
+  fi
+
+  if [ $exit_code -eq 10 ] && [ "$retry_once_with_cleanup" = true ]; then
+    if [ -d .terraform ]; then
+      # Will only be displayed if validation fails again.
+      common::colorify "yellow" "Validation failed. Removing .terraform from: $dir_path"
+      rm -rf .terraform
+      do_validate
```
yermulnik, line 124: Would it make sense to throw another diagnostic message here about re-validation being done? Sort of this:
```suggestion
      common::colorify "yellow" "Re-validating: $dir_path"
      do_validate
```

yermulnik: ...just to make it obvious that re-validation is being done and not just the dir has been removed.

yermulnik:
Other than comments above I think the PR looks good to me.

hooks/terraform_validate.sh
baolsen: Agree, thanks for the suggestion

baolsen:
Hey @yermulnik thanks for the further comments.
I think I've addressed them in the latest push, please take a look.

yermulnik (approved):
LGTM

yermulnik:
Let's summon @MaxymVlasov to take another look and put his opinion.

hooks/terraform_validate.sh
```diff
@@ -43,16 +43,97 @@ function per_dir_hook_unique_part {
   local -a -r args=("$@")
 
   local exit_code
-  local validate_output
+  #
+  # Get hook settings
+  #
+  local retry_once_with_cleanup=false
+
+  IFS=";" read -r -a configs <<< "${HOOK_CONFIG[*]}"
```
MaxymVlasov, line 51: Okay, that usage a little bit breaks the conception of "provide all vars from one function to another explicitly as much as possible in bash"

Also, I see that it needs some refactoring across a few different hooks and _common.sh, so it may be done as a separate PR because the final result should look like this:

```bash
 common::per_dir_hook "$HOOK_ID" "${#ARGS[@]}" "${ARGS[@]}" "${#HOOK_CONFIG[@]}" "${HOOK_CONFIG[@]}" "${FILES[@]}"
```

That will add the ability setup configs for all available (and not deprecated) hooks

hooks/terraform_validate.sh
```diff
@@ -43,16 +43,97 @@ function per_dir_hook_unique_part {
...
-  validate_output=$(terraform validate "${args[@]}" 2>&1)
-  exit_code=$?
+  function do_validate {
+    validate_output=$(terraform validate "${args[@]}" 2>&1)
+    exit_code=$?
+    return $exit_code
+  }
+
+  function parse_validate {
+    # Requires jq
```
MaxymVlasov, line 78: Let's move that function outside `per_dir_hook_unique_part` function body, before it.

And please add documentation to both, like in `per_dir_hook_unique_part`

hooks/terraform_validate.sh
```diff
@@ -43,16 +43,97 @@ function per_dir_hook_unique_part {
...
+
+  if [ "$retry_once_with_cleanup" == "true" ]; then
+    parse_validate
+    exit_code=$?
+  else
+    do_validate
+    exit_code=$?
+  fi
+
+  if [ $exit_code -eq 10 ] && [ "$retry_once_with_cleanup" == "true" ]; then
```
MaxymVlasov, line 127: 10 - is the "magic number". Add a comment before about "why 10"

hooks/terraform_validate.sh
```diff
@@ -43,16 +43,97 @@ function per_dir_hook_unique_part {
...
+    if [ -d .terraform ]; then
+      # Will only be displayed if validation fails again.
+      common::colorify "yellow" "Validation failed. Removing .terraform from: $dir_path"
+      rm -rf .terraform
```
MaxymVlasov, line 131: Let's address https://github.com/antonbabenko/pre-commit-terraform/issues/203#issuecomment-918791847 and make it from workaround to useful for all solutions (also need changes to README)
```suggestion
      common::colorify "yellow" "Validation failed. Removing cached providers and modules from: $dir_path/.terraform"
      # Remove all files in the `.terraform` dir except `environment` - where stored info about TF workspaces.
      rm -rf .terraform/!("environment")
```

yermulnik: Not sure about default value of `extglob` across different versions of Bash and don't like to take a chance on it, so let's be more explicit on what we need to delete then: `rm -rf .terraform/{modules,providers}/` — this should be very much sufficient to `Removing cached providers and modules from` and needs no Bash-specific quirks.

Just for reference here's [the description](https://developer.hashicorp.com/terraform/cli/init#working-directory-contents) of `.terraform` dir:
> A hidden `.terraform` directory, which Terraform uses to manage cached provider plugins and modules, record which [workspace](https://developer.hashicorp.com/terraform/cli/workspaces) is currently active, and record the last known backend configuration in case it needs to migrate state on the next run. This directory is automatically managed by Terraform, and is created during initialization.

PS: I don't count in custom setups like https://developer.hashicorp.com/terraform/cli/config/config-file#explicit-installation-method-configuration 😄

MaxymVlasov: You're right
```suggestion
      common::colorify "yellow" "Validation failed. Removing cached providers and modules from: $dir_path/.terraform"
      # .terraform` dir includes could some extra files, like `environment`,
      # where stored info about TF workspaces, so we can't just remove
      # `.terraform` dir complitely.
      rm -rf .terraform/{modules,providers}/
```

yermulnik: Not sure how to suggest change to your suggested change, so posting it as is:
```
      # `.terraform` dir may comprise some extra files, like `environment`
      # which stores info about current TF workspace, so we can't just remove
      # `.terraform` dir completely.
```

hooks/terraform_validate.sh
baolsen: Thanks. Def going to leave this for a separate PR because I copy-pasted this approach from elsewhere. At least it is now "consistently wrong"

baolsen:
Hey @yermulnik & @MaxymVlasov 

Thanks for the feedback

I think I've got all of it included in the latest commit, please take a look.

Only part I'm not 100% sure is the feel of the README - eg should we still include the 2nd option for the workaround or is this option 1 sufficient for the "force-init" idea and we just remove option 2 entirely.

(Moved the warning back to the bottom of Option 2 since we now avoid deleting any environment file)

yermulnik (approved):
LGTM

hooks/terraform_validate.sh
```diff
@@ -43,17 +84,51 @@ function per_dir_hook_unique_part {
...
+    esac
+  done
 
   common::terraform_init 'terraform validate' "$dir_path" || {
     exit_code=$?
     return $exit_code
   }
 
-  # pass the arguments to hook
   validate_output=$(terraform validate "${args[@]}" 2>&1)
```
MaxymVlasov, line 112:
```suggestion
  # pass the arguments to hook
  validate_output=$(terraform validate "${args[@]}" 2>&1)
```

hooks/_common.sh
```diff
@@ -299,13 +299,16 @@ function common::terraform_init {
...
-  if [ ! -d .terraform ]; then
+  if [ ! -d .terraform/modules ] || [ ! -d .terraform/providers ]; then
     init_output=$(terraform init -backend=false "${TF_INIT_ARGS[@]}" 2>&1)
     exit_code=$?
 
     if [ $exit_code -ne 0 ]; then
       common::colorify "red" "'terraform init' failed, '$command_name' skipped: $dir_path"
       echo -e "$init_output\n\n"
+    else
+
```
MaxymVlasov, line 310:
```suggestion
```

MaxymVlasov:
The proposed test wasn't work...

<details>

```yaml
    rev: '850bb967e2d38a89665f0640467a3d5d2d6f634e'
    hooks:
      - id: terraform_fmt
      - id: terraform_validate
        args:
```

```bash
✘1 ➜ ga .pre-commit-config.yaml && pre-commit run
Terraform validate.......................................................Failed
- hook id: terraform_validate
- exit code: 2

/home/vm/.cache/pre-commit/reponxjr7ygn/hooks/terraform_validate.sh: line 85: syntax error near unexpected token `exit_code=$?'
```
</details>

So it wasn't work 🤔 
But I make some changes to improve readability, etc., and now it works, at least for me.

@baolsen please double-check that stuff like that works fine

```yaml
repos:
  rev: 15cafcddcce89ef85dfc123d55085cf2255df842
  hooks:
    - id: terraform_validate
      args:
        - --tf-init-args=-upgrade
        - --tf-init-args=-get=true
        - --envs=AWS_DEFAULT_REGION="us-west-2"
        - --envs=AWS_ACCESS_KEY_ID="anaccesskey"
        - --envs=AWS_SECRET_ACCESS_KEY="asecretkey"
        - --hook-config=--retry-once-with-cleanup=true     # Boolean. true or false
      verbose: true
```

hooks/terraform_validate.sh
```diff
@@ -24,6 +24,40 @@ function main {
   common::per_dir_hook "$HOOK_ID" "${#ARGS[@]}" "${ARGS[@]}" "${FILES[@]}"
 }
 
+#######################################################################
+# Run `terraform validate` and handle errors. Requires `jq``
```
MaxymVlasov, line 28:
```suggestion
# Run `terraform validate` and handle errors. Requires `jq`
```

baolsen:
@MaxymVlasov thanks for the updates.

I did a few tests on 955cb56 and picked up a few bugs, fixed them locally and pushed. Please check again.

baolsen:
I also see a lot of unrelated changes in the README, not sure if those are intended.

hooks/terraform_validate.sh
```diff
@@ -24,6 +24,43 @@ function main {
...
+#######################################################################
+# Run `terraform validate` and match errors. Requires `jq`
+# Arguments:
+#   validate_output (string with json) output of `terraform validate` command
+# Outputs:
+#   Returns integer:
+#    - 0 (no errors)
+#    - 1 (matched errors; retry)
+#    - 2 (no matched errors; do not retry)
+#######################################################################
```
baolsen, line 36: Renamed from `find_validate_errors` to `match_validate_errors` to align better with the functions behaviour. We dont care about all errors, only specific ones.

MaxymVlasov:
> I also see a lot of unrelated changes in the README, not sure if those are intended.

Yeah, that's `chore` changes, which mostly fix style and language, because I had time to check it in this PR, and they are so tiny, so no need to create a new PR and then, possible, deal with conflicts

MaxymVlasov (approved):

README.md
```diff
@@ -652,7 +653,30 @@ Example:
...
+    ```
+
+    > Note: That flag require additional installed dependency: `jq`.
+
+    If `--retry-once-with-cleanup=true`, then in each failed directory the cached modules and providers from the `.terraform` directory will be deleted, before retrying once more. To avoid unnecessary deletion of this directory, the cleanup and retry will only happen if Terraform produces any of the following error messages:
+
+    * Missing or corrupted provider plugins
+    * Module source has changed
+    * Module version requirements have changed
+    * Module not installed
```
yermulnik, line 673: Might it worth to wrap sentences into double quotes (and/or make the use _italic_ font?) to explicitly denote they are error messages from `terraform validate` rather than a free-form text of what we think is validation error message?
Also added missing message based on code and lowercased 1st message to match what's in code:
```suggestion
    * "_missing or corrupted provider plugins_"
    * "_Module source has changed_"
    * "_Module version requirements have changed_"
    * "_Module not installed_"
    * "_Could not load plugin_"
```

README.md
```diff
@@ -666,22 +690,23 @@ Example:
 
    `terraform_validate` hook will try to reinitialize them before running the `terraform validate` command.
 
-    **Warning:** If you use Terraform workspaces, DO NOT use this workaround ([details](https://github.com/antonbabenko/pre-commit-terraform/issues/203#issuecomment-918791847)). Wait to [`force-init`](https://github.com/antonbabenko/pre-commit-terraform/issues/224) option implementation.
+    **Warning:** If you use Terraform workspaces, DO NOT use this option ([details](https://github.com/antonbabenko/pre-commit-terraform/issues/203#issuecomment-918791847)). Consider the first option, or wait to [`force-init`](https://github.com/antonbabenko/pre-commit-terraform/issues/224) option implementation.
```
yermulnik, line 693:
```suggestion
    **Warning:** If you use Terraform workspaces, DO NOT use this option ([details](https://github.com/antonbabenko/pre-commit-terraform/issues/203#issuecomment-918791847)). Consider the first option, or wait for [`force-init`](https://github.com/antonbabenko/pre-commit-terraform/issues/224) option implementation.
```

hooks/terraform_validate.sh
```diff
@@ -43,16 +80,64 @@ function per_dir_hook_unique_part {
...
+
+  for c in "${configs[@]}"; do
+
+    IFS="=" read -r -a config <<< "$c"
+    key=${config[0]}
+    value=${config[1]}
+
+    case $key in
+      --retry-once-with-cleanup)
+        retry_once_with_cleanup=$value
```
yermulnik, line 98: Should we take first or last parameter occurance when there are more than one specified (e.g. by mistake)? For the former (first takes precedence) we should add something like this:
```suggestion
        if [ ! $retry_once_with_cleanup ]; then retry_once_with_cleanup=$value; fi
```

hooks/terraform_validate.sh
```diff
@@ -43,16 +80,64 @@ function per_dir_hook_unique_part {
...
+    validate_output=$(terraform validate -json "${args[@]}" 2>&1)
+    exit_code=$?
+
+    # Match specific validation errors
+    local -i validate_errors_matched
+    match_validate_errors "$validate_output"
+    validate_errors_matched=$?
+
+    # Errors matched; Retry validation
+    if [ "$validate_errors_matched" = "1" ]; then
```
yermulnik, line 123: Arithmetic evaluation (comparison of integers in this case) should use `-eq` instead of `=` (or `==`, which is the same as `=`):
```suggestion
    if [ "$validate_errors_matched" -eq 1 ]; then
```

hooks/terraform_validate.sh
```diff
@@ -43,16 +80,64 @@ function per_dir_hook_unique_part {
...
+      common::colorify "yellow" "Validation failed. Removing cached providers and modules from $dir_path/.terraform"
```
yermulnik, line 124:
```suggestion
      common::colorify "yellow" "Validation failed. Removing cached providers and modules from \"$dir_path/.terraform\" directory"
```

yermulnik (approved):
LGTM with tiny remarks left for your guys consuderation

hooks/terraform_validate.sh
MaxymVlasov: Better stop execution and ask what user check configuration.


```suggestion
        if [ $retry_once_with_cleanup ]; then 
          common::colorify "yellow" 'Invalid hook config. You specify more than one "--retry-once-with-cleanup" flag"'
          exit 1
        fi
        retry_once_with_cleanup=$value
```

yermulnik: Yep, makes sense!

hooks/terraform_validate.sh
```diff
@@ -43,16 +80,64 @@ function per_dir_hook_unique_part {
   local -a -r args=("$@")
 
   local exit_code
-  local validate_output
+  #
+  # Get hook settings
+  #
+  local retry_once_with_cleanup=false
```
MaxymVlasov, line 86:
```suggestion
  local retry_once_with_cleanup
```

antonbabenko (approved):

antonbabenko:
This PR is included in version 1.77.0 :tada:

brettcurtis:
Here to say thanks, this PR is a big help!

MaxymVlasov:
Hi @brettcurtis.
`terraform init -upgrade` could act differently than you expected, eg. download a new major provider version. See details of [how it works here](https://discuss.hashicorp.com/t/what-does-terraform-init-upgrade-do/32638/4)

brettcurtis:
Thanks for reply @MaxymVlasov - Yeah that could mess up folks I suppose. For us we always pull current release shift testing left and fix any problems before we hit production. So basically, terraform init -upgrade is part of our flow already on any PR.

brettcurtis:
Ah, so I just tested this: 

```none
Validation failed: regional/infra
{
  "format_version": "1.0",
  "valid": false,
  "error_count": 1,
  "warning_count": 0,
  "diagnostics": [
    {
      "severity": "error",
      "summary": "missing or corrupted provider plugins:
  - registry.terraform.io/hashicorp/google: there is no package for registry.terraform.io/hashicorp/google 4.44.1 cached in .terraform/providers
  - registry.terraform.io/hashicorp/google-beta: there is no package for registry.terraform.io/hashicorp/google-beta 4.44.1 cached in .terraform/providers
  - registry.terraform.io/hashicorp/random: there is no package for registry.terraform.io/hashicorp/random 3.4.3 cached in .terraform/providers",
      "detail": ""
    }
  ]
}
```

The reason I'm in this state is because dependabot manages our terraform provider versions and merged a couple PRs with provider updates since the last time I worked on the given repo. So, I guess this PR doesn't solve that problem? Another thing I noticed is that run init on directories that only include .tfvars.

MaxymVlasov:
@brettcurtis better create an issue and describe which problems you face, how you get JSON output (because hook should always return non-JSON outputs to the end-user), and other things that exist in bug issue template

brettcurtis:
Will do, I'm trying to come up with a simple test case but in my tests, I can't reproduce what I was expecting to be the problem. The assumption was that my local .terraform has an older version of a provider cached and I pulled down a new .terraform.lock.hcl that was updated with a new provider version it would fail. I'll report back once I can figure out how to reproduce.

PR #496, opened 2023-02-21
feat(ci): Build multi-arch Docker images (`amd64`, `arm64`)

smiller171:
Just looking to add ARM builds for my needs

- [ ] This PR fixes a bug.

Add ARM builds for running on M1/M2 Macs and other Arm CPUs

### How can we test changes

After building, run the following command and check that the output contains `amd64`, `arm` and `arm64`

`docker manifest inspect ghcr.io/antonbabenko/pre-commit-terraform:nightly | jq -r '.manifests[].platform.architecture' | sort -u`


MaxymVlasov:
Are you sure that this is all, which need to make the amd64 packages happily work with arm64? IE

https://github.com/antonbabenko/pre-commit-terraform/blob/1e623e8d87fb285318cb99d2148dc4d1d0374f37/Dockerfile#L75

smiller171:
Nope. I hadn't checked that far yet. Give me a bit and I can make it work though.

smiller171:
@MaxymVlasov I think this should actually work now. I tested locally and could build it on both platforms, and I _think_ I got the GH Actions config correct.

MaxymVlasov:
>buildx failed with: ERROR: multiple platforms feature is currently not supported for docker driver. Please switch to a different driver (eg. "docker buildx create --use")

MaxymVlasov:
@smiller171 ?

Dockerfile
```diff
@@ -94,19 +98,22 @@ RUN . /.env \
...
     ) && chmod +x terragrunt \
     ; fi
 
 
 # Terrascan
 RUN . /.env && \
     if [ "$TERRASCAN_VERSION" != "false" ]; then \
+    if [ "$TARGETARCH" != "amd64" ]; then ARCH="$TARGETARCH"; else ARCH="x86_64"; fi; \
+    # Convert the first letter to Uppercase
+    OS="$(echo ${TARGETOS} | cut -c1 | tr '[:lower:]' '[:upper:]' | xargs echo -n; echo ${TARGETOS} | cut -c2-100)"; \
```
yermulnik, line 112: I'm wondering whether we can use something like below?
```suggestion
    OS="$(echo ${TARGETOS} | sed 's/[^ ]\+/\L\u&/g')"; \
```
If not, then the `cut -c2-100` should probably be replaced with `cut -c2-` (not sure why would we want to crop original string by 100 chars).

Dockerfile
```diff
@@ -1,5 +1,9 @@
 ARG TAG=3.11.1-alpine3.17@sha256:d8b0703ce84fe5a52d485f212e9d852bcdb8606798064f5f21af57325a7cf73f
 FROM python:${TAG} as builder
+ARG TARGETPLATFORM
+ARG TARGETOS
+ARG TARGETARCH
```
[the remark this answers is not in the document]
MaxymVlasov: They are provided by `docker buildx` automatically.

Dockerfile
```diff
@@ -1,5 +1,9 @@
 ARG TAG=3.11.1-alpine3.17@sha256:d8b0703ce84fe5a52d485f212e9d852bcdb8606798064f5f21af57325a7cf73f
 FROM python:${TAG} as builder
+ARG TARGETPLATFORM
```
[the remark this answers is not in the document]
MaxymVlasov:
```suggestion
```

Dockerfile
MaxymVlasov:
```
/ # TARGETOS=darwin
/ # echo ${TARGETOS} | sed 's/[^ ]\+/\L\u&/g'
Ludarwin
/ # 
```

Dockerfile
```diff
@@ -94,19 +96,22 @@ RUN . /.env \
...
     ) && chmod +x terragrunt \
     ; fi
 
 
 # Terrascan
 RUN . /.env && \
     if [ "$TERRASCAN_VERSION" != "false" ]; then \
+    if [ "$TARGETARCH" != "amd64" ]; then ARCH="$TARGETARCH"; else ARCH="x86_64"; fi; \
+    # Convert the first letter to Uppercase
+    OS="$(echo ${TARGETOS} | cut -c1 | tr '[:lower:]' '[:upper:]' | xargs echo -n; echo ${TARGETOS} | cut -c2-100)"; \
```
MaxymVlasov, line 110:
```suggestion
    OS="$(echo ${TARGETOS} | cut -c1 | tr '[:lower:]' '[:upper:]' | xargs echo -n; echo ${TARGETOS} | cut -c2-)"; \
```

.github/workflows/build-image-test.yaml
```diff
@@ -22,17 +22,25 @@ jobs:
             .dockerignore
             tools/entrypoint.sh
 
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@4b4e9c3e2d4531116a6f8ba8e71fc6e2cb6e6c8c # v2.5.0
```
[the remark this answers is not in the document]
yermulnik: > Because the tag can be repinned

This is why GIT tags are better: they allow maintainers to fix things quick.
Though Renovate bot... okay then =)

Dockerfile
yermulnik: Does Docker use the same odd version of `sed` as your MacOS does? 😉

.github/workflows/build-image-test.yaml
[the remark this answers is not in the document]
MaxymVlasov: >This is why GIT tags are worse: they allow maintainers to break things quick.

🌚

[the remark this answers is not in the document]
MaxymVlasov: https://reproducible-builds.org/

Dockerfile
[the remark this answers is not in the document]
yermulnik: Gotcha. Thanks.
Does plain `docker build` provide them  as well? If not, should we update README for those who want to build image on their own with a notice re the `buildx` being a requirement?

Dockerfile
MaxymVlasov: No idea what it use, because I use Ubuntu, and tests was inside image

MaxymVlasov: Otherwise here was be `OS="${TARGETOS^}"`

.github/workflows/build-image-test.yaml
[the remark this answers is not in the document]
yermulnik: Avoid relying on projects that don't deserve to be trusted.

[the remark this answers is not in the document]
MaxymVlasov: Nice Friday joke!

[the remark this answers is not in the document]
yermulnik: No kidding.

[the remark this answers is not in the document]
yermulnik: In this particular use case what you're talking about is more of a oversecuring where there's no need to.

yermulnik (approved):

Dockerfile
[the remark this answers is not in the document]
MaxymVlasov: Documented

README.md
```diff
@@ -97,7 +97,10 @@ All available tags [here](https://github.com/antonbabenko/pre-commit-terraform/p
 
 **Build from scratch**:
 
-When `--build-arg` is not specified, the latest version of `pre-commit` and `terraform` will be only installed.
+> **Note**: To build image you need have [`docker buildx`](https://docs.docker.com/build/install-buildx/) enabled as default builder.
```
yermulnik, line 100:
```suggestion
> **Note**: To build image you need to have [`docker buildx`](https://docs.docker.com/build/install-buildx/) enabled as default builder.  
```

README.md
```diff
@@ -97,7 +97,10 @@ All available tags [here](https://github.com/antonbabenko/pre-commit-terraform/p
 
 **Build from scratch**:
 
-When `--build-arg` is not specified, the latest version of `pre-commit` and `terraform` will be only installed.
+> **Note**: To build image you need have [`docker buildx`](https://docs.docker.com/build/install-buildx/) enabled as default builder.  
+> Otherwise - provide `TARGETOS` and `TARGETARCH` as additional `--build-arg`'s to `docker build`.
```
yermulnik, line 101: Still don't get why don't we just set default value in Dockerfile =) This shouldn't prevent `buildx` from overriding them. Or am I getting this incorrectly?

README.md
```diff
@@ -97,7 +97,10 @@ All available tags [here](https://github.com/antonbabenko/pre-commit-terraform/p
...
+
+When hooks-related `--build-arg`'s is not specified, only the latest version of `pre-commit` and `terraform` will be installed.
```
yermulnik, line 103:
```suggestion
When hooks-related `--build-arg`s are not specified, only the latest version of `pre-commit` and `terraform` will be installed.
```

README.md
MaxymVlasov: `buildx` is required for multi-arch builds. I'd like to force users to use `buildx` when they build images from scratch, and because it's not our "core functional" there is no reason to maintain compatibility with legacy `docker build`

MaxymVlasov: Buildx exist at least for 3 years

yermulnik: > `buildx` is required for multi-arch builds. I'd like to force users to use `buildx` when they build images from scratch, and because it's not our "core functional" there is no reason to maintain compatibility with legacy `docker build`

Makes sens.

PR #509, opened 2023-04-24
fix: Terraform validate - only run tf init if providers directory is missing in .terraform

smelchior:

- [x] This PR fixes a bug.

If a terraform states does not include modules, a `.terraform/modules` directory is not present. For these cases, the tf init is run every time the validate hook runs, which takes extra time + causes issues if e.g. a remote state is used which causes the init to fail.

### How can we test changes

I checked it with my repositories as described above, and the tf init did now only run if the .terraform/providers directly is not present.

### DIFF
--- a/hooks/_common.sh
+++ b/hooks/_common.sh
@@ -299,7 +299,8 @@ function common::terraform_init {
     TF_INIT_ARGS+=("-no-color")
   fi
 
-  if [ ! -d .terraform/modules ] || [ ! -d .terraform/providers ]; then
+  # only check for providers directory here as /modules is only present if code uses terraform modules
+  if [ ! -d .terraform/providers ]; then
     init_output=$(terraform init -backend=false "${TF_INIT_ARGS[@]}" 2>&1)
     exit_code=$?
 


yermulnik (approved):
Makes sense.

@baolsen JFYI (since this was a part of https://github.com/antonbabenko/pre-commit-terraform/pull/441)

yermulnik:
@MaxymVlasov Can you think of any reason to check both dirs instead of only `.terraform/providers/`?

MaxymVlasov:
>If a terraform states does not include modules

If a terraform state does not include providers, what then?

yermulnik:
> but let's brainstorm first are there any modules that don't need providers

Output-only modules.

yermulnik:
Maybe add something like `--force-no-init` or `--ignore-init` arg for `common::terraform_init()` function so that it always return success when this flag is passed by user? 🤔

MaxymVlasov:
> Maybe add something like `--force-no-init` or `--ignore-init` arg for `common::terraform_init()` function so that it always returns success when this flag is passed

Hmm, or ask the user to provide a comma-separated list of what paths should be checked, ie:

`- --run-tf-init-if-not-exits-any-of=.terraform/providers,.terraform/foobar`

yermulnik:
> Hmm, or ask the user to provide a comma-separated list of what paths should be checked, ie:

Might be a bit of overkilling as of quite rare use case though. Those who would want to provide a list of options, most probably were much more smarter and already use `mkdir` to circumvent this hook's unwanted behavior 😃

MaxymVlasov:
@smelchior good idea, let's do it

Here are my errors when no providers and modules provided

```json
➜ time t validate -json
{
  "format_version": "1.0",
  "valid": false,
  "error_count": 4,
  "warning_count": 0,
  "diagnostics": [
    {
      "severity": "error",
      "summary": "Module not installed",
      "detail": "This module is not yet installed. Run \"terraform init\" to install all modules required by this configuration.",
      "range": {
        "filename": "main.tf",
        "start": {
          "line": 75,
          "column": 1,
          "byte": 2450
        },
        "end": {
          "line": 75,
          "column": 23,
          "byte": 2472
        }
      },
      "snippet": {
        "context": null,
        "code": "module \"account_label\" {",
        "start_line": 75,
        "highlight_start_offset": 0,
        "highlight_end_offset": 22,
        "values": []
      }
    },
    {
      "severity": "error",
      "summary": "Module not installed",
      "detail": "This module is not yet installed. Run \"terraform init\" to install all modules required by this configuration.",
      "range": {
        "filename": "providers.tf",
        "start": {
          "line": 20,
          "column": 1,
          "byte": 388
        },
        "end": {
          "line": 20,
          "column": 21,
          "byte": 408
        }
      },
      "snippet": {
        "context": null,
        "code": "module \"email_label\" {",
        "start_line": 20,
        "highlight_start_offset": 0,
        "highlight_end_offset": 20,
        "values": []
      }
    },
    {
      "severity": "error",
      "summary": "Module not installed",
      "detail": "This module is not yet installed. Run \"terraform init\" to install all modules required by this configuration.",
      "range": {
        "filename": "providers.tf",
        "start": {
          "line": 15,
          "column": 1,
          "byte": 285
        },
        "end": {
          "line": 15,
          "column": 19,
          "byte": 303
        }
      },
      "snippet": {
        "context": null,
        "code": "module \"iam_roles\" {",
        "start_line": 15,
        "highlight_start_offset": 0,
        "highlight_end_offset": 18,
        "values": []
      }
    },
    {
      "severity": "error",
      "summary": "Module not installed",
      "detail": "This module is not yet installed. Run \"terraform init\" to install all modules required by this configuration.",
      "range": {
        "filename": "context.tf",
        "start": {
          "line": 23,
          "column": 1,
          "byte": 918
        },
        "end": {
          "line": 23,
          "column": 14,
          "byte": 931
        }
      },
      "snippet": {
        "context": null,
        "code": "module \"this\" {",
        "start_line": 23,
        "highlight_start_offset": 0,
        "highlight_end_offset": 13,
        "values": []
      }
    }
  ]
}
terraform validate -json  0.12s user 0.05s system 141% cpu 0.124 total
```

Because it took only 0.12s, in case of a problem, that absolutely OK to do that check

I suppose it could be done using next or little-bit changed function: https://github.com/antonbabenko/pre-commit-terraform/blob/e1d78ae91739b2881a4f2d9f6264c3fb9ce29150/hooks/terraform_validate.sh#L37


That code is mostly the same as what we need to execute then

https://github.com/antonbabenko/pre-commit-terraform/blob/e1d78ae91739b2881a4f2d9f6264c3fb9ce29150/hooks/terraform_validate.sh#L117-L139

As replace to 
https://github.com/antonbabenko/pre-commit-terraform/blob/e1d78ae91739b2881a4f2d9f6264c3fb9ce29150/hooks/terraform_validate.sh#L107-L110

smelchior:
perfect, i will create a new PR and close this one afterwards :)

smelchior:
As discussed, I made the change in https://github.com/antonbabenko/pre-commit-terraform/pull/524 in a "simple" way, i think this should be sufficient to achieve what we want :-)

Issue #303, opened 2021-12-16
Add unit-tests that will found problems with new code


MaxymVlasov:
### What problem are you facing?

New PR's time to time break existing functional, and that can't be found by PR review.
Here one of the examples why we should test corner cases on PR's: https://github.com/antonbabenko/pre-commit-terraform/pull/302

MaxymVlasov:
Tried `bach` - no idea how to test ARGS results after `common::parse_cmdline` execution. Test just not go inside for-loop

MaxymVlasov:
Need to check 
https://opensource.com/article/19/2/testing-bash-bats

MaxymVlasov:
Go/Ruby
https://github.com/cloudfoundry-incubator/kubo-deployment/tree/v0.14.0/src/kubo-deployment-tests

https://t.me/catops_chat/61258

MaxymVlasov:
That works awesome
https://github.com/kward/shunit2

Issue #508, opened 2023-04-21
terraform_fmt fails with tfenv and "min-required" version


jcarlson:
### Describe the bug

`tfenv` allows for [the use of `min-required`](https://github.com/tfutils/tfenv/tree/v3.0.0#min-required) as the version specified in a project's `.terraform-version` file, which then parses the Terraform `required_version` from the configuration files.

The `terraform_fmt` hook runs `terraform fmt` using the directory of the changed file as the working directory, rather than the repository root as the working directory.

In a configuration that includes sub-modules as sub-directories, where that sub-module requires a different, less-restrictive version of Terraform, the `terraform_fmt` hook then attempts to install the wrong version of Terraform by installing the minimum version for that sub-module, rather than the version for that project.

For example, I have a root configuration that uses Terraform 1.4.4, and a sub-module which requires Terraform `>= 0.14`, which 1.4.4 satisfies. However, when files in the sub-module change, `terraform_fmt` failed because `tfenv` is attempting to install Terraform 0.14, and on my M1 Apple Silicon Mac, this fails, since there is no ARM build of Terraform 0.14.

### How can we reproduce it?

Observe the following sample configuration:

`main.tf`
```hcl
terraform {
  required_version = "1.4.4"
}

module "sub_module" {
  source = "./modules/sub"
}

output "lorem_ipsum" {
  value = join([module.sub_module.foo, module.sub_module.bar], " ")
}

```

`modules/sub/main.tf`
```hcl
terraform {
  required_version = ">= 0.14"
}

output "foo" {
  value = "Foo is fun."
}

output "bar" {
  value = "Bar is bad."
}
```

The repository root also includes the following files:

`.terraform-version`
```
min-required
```

`.pre-commit-config.yaml`
```yaml
---
repos:
  - repo: https://github.com/antonbabenko/pre-commit-terraform
    rev: v1.77.2
    hooks:
      - id: terraform_fmt
```

With the above configuration in a clean git working tree:
```bash
cat <<-EOF | tee -a modules/sub/main.tf

output "baz" {
  value = "Baz is not used"
}
EOF

git add modules/sub/main.tf
git commit
```

The following error is produced:

```
[INFO] Initializing environment for https://github.com/antonbabenko/pre-commit-terraform.
Terraform fmt............................................................Failed
- hook id: terraform_fmt
- exit code: 127

[...]

PR #512, opened 2023-04-26
feat: TFLint: Add `--hook-config=--delegate-chdir` to use `tflint -chdir`

### LINKED ISSUE #511: tflint doesn't include the full path to warnings.
### Describe the bug

When running pre-commit the errors are always bubbled up using relative paths:
```
TFLint in <my_relative_path>:
1 issue(s) found:

variables.tf:6:1: Warning - variable "unused" is declared but not used (terraform_unused_declarations)
```

We want the errors to provide the full path.

```
TFLint in <my_relative_path>:
1 issue(s) found:

<my_relative_path>/variables.tf:6:1: Warning - variable "unused" is declared but not used (terraform_unused_declarations)
```

This makes it hard for editors + CI checks to ensure that the errors are passed to the full path to make it easier to debug.

### How can we reproduce it?

Create a TFlint warning error in a nested module. 

### Environment information
```
OS: MacOS
Darwin  22.4.0 Darwin Kernel Version 22.4.0: Mon Mar  6 20:59:28 PST 2023; root:xnu-8796.101.5~3/RELEASE_ARM64_T6000 arm64
```

```bash
pre-commit 3.1.1
Terraform v1.2.0
Python 3.11.2+
Python 3.11.2+
checkov 2.3.90
terraform-docs version v0.16.0 darwin/arm64
terragrunt version v0.36.7-3-gc1a70b07ac3d
terrascan terrascan SKIPPED
TFLint version 0.44.1
+ ruleset.aws (0.17.1)
+ ruleset.terraform (0.1.1)
tfsec v1.28.1
tfupdate tfupdate SKIPPED
hcledit hcledit SKIPPED
```

* `.pre-commit-config.yaml`:

```bash
    rev: v1.75.0
    hooks:
      - id: terraform_tflint
        alias: tflint_init
        name: run tflint --init
        stages: [commit]
        args:
          - --args=--init
          - --args=--no-color
      - id: terraform_tflint
        alias: tflint_init
        name: run tflint --init
        stages: [commit]
        args:
          - --args=--init
          - --args=--no-color
      - id: terraform_tflint
        stages: [commit]
        args:
          - --args=-f compact
```

</details>

lexton:

- [ ] This PR fixes a bug.

This change leverages the `tflint -chdir` flag to do the nested lookup and provide fully qualified paths to the output.

It unwraps the `pushd`/`popd` functionality and delegates the directory changing to the binary process via chdir.

This would be a breaking behavior if a user tried to add a `-chdir` flag to the `args` - but I can't think of a valid reason that anyone would do this - or if that functionality would even work with the pre-commit-terraform hooks.

The `-chdir` flag was added in [tflint 0.44.0](https://github.com/terraform-linters/tflint/releases/tag/v0.44.0) - I don't know how to enforce specific versions of the binary or to trivially change the commit hook driver behavior based upon the version. This flag is forwards compatible, while the `dir` arg to tflint is being deprecated.

Totally happy to revisit the approach or just keep maintaining a fork for my use cases.

Fixes #511 

### How can we test changes

Locally examine the output when the tflint succeeds vs fails.

### DIFF
--- a/README.md
+++ b/README.md
@@ -599,6 +599,14 @@ To replicate functionality in `terraform_docs` hook:
         - --args=--config=__GIT_WORKING_DIR__/.tflint.hcl
     ```
 
+3. By default pre-commit-terraform performs directory switching into the terraform modules for you. If you want to delgate the directory changing to the binary - this will allow tflint to determine the full paths for error/warning messages, rather than just module relative paths. *Note: this requires `tflint>=0.44.0`.* For example:
+
+    ```yaml
+    - id: terraform_tflint
+          args:
+            - --hook-config=--delegate-chdir
+    ```
+
 
 ### terraform_tfsec
 
--- a/hooks/_common.sh
+++ b/hooks/_common.sh
@@ -257,7 +257,7 @@ function common::per_dir_hook {
       final_exit_code=$exit_code
     fi
 
-    if ! $DELEGATE_CHDIR; then
+    if [[ $DELEGATE_CHDIR != true ]]; then
       popd > /dev/null
     fi
 
--- a/hooks/_common.sh
+++ b/hooks/_common.sh
@@ -218,7 +218,7 @@ function common::per_dir_hook {
   done
 
   # Lookup hook-config for modifiers that impact common behavior
-  local DELEGATE_CHDIR=false
+  local change_dir_in_unique_part=false
   IFS=";" read -r -a configs <<< "${HOOK_CONFIG[*]}"
   for c in "${configs[@]}"; do
     IFS="=" read -r -a config <<< "$c"
@@ -230,7 +230,7 @@ function common::per_dir_hook {
         # this flag will skip pushing and popping directories
         # delegating the responsibility to the hooked plugin/binary
         if [[ ! $value || $value == true ]]; then
-          DELEGATE_CHDIR=true
+          change_dir_in_unique_part="delegate_chdir"
         fi
         ;;
     esac
@@ -246,18 +246,18 @@ function common::per_dir_hook {
   for dir_path in $(echo "${dir_paths[*]}" | tr ' ' '\n' | sort -u); do
     dir_path="${dir_path//__REPLACED__SPACE__/ }"
 
-    if [[ $DELEGATE_CHDIR != true ]]; then
+    if [[ $change_dir_in_unique_part == false ]]; then
       pushd "$dir_path" > /dev/null || continue
     fi
 
-    per_dir_hook_unique_part "$dir_path" "${args[@]}"
+    per_dir_hook_unique_part "$dir_path" "$change_dir_in_unique_part" "${args[@]}"
 
     local exit_code=$?
     if [ $exit_code -ne 0 ]; then
       final_exit_code=$exit_code
     fi
 
-    if [[ $DELEGATE_CHDIR != true ]]; then
+    if [[ $change_dir_in_unique_part == false ]]; then
       popd > /dev/null
     fi
 
--- a/hooks/_common.sh
+++ b/hooks/_common.sh
@@ -217,6 +217,25 @@ function common::per_dir_hook {
     ((index += 1))
   done
 
+  # Lookup hook-config for modifiers that impact common behavior
+  local DELEGATE_CHDIR=false
+  IFS=";" read -r -a configs <<< "${HOOK_CONFIG[*]}"
+  for c in "${configs[@]}"; do
+    IFS="=" read -r -a config <<< "$c"
+    key=${config[0]}
+    value=${config[1]}
+
+    case $key in
+      --delegate-chdir)
+        # this flag will skip pushing and popping directories
+        # delegating the responsibility to the hooked plugin/binary
+        if [[ ! $value || $value == true ]]; then
+          DELEGATE_CHDIR=true
+        fi
+        ;;
+    esac
+  done
+
   # preserve errexit status
   shopt -qo errexit && ERREXIT_IS_SET=true
   # allow hook to continue if exit_code is greater than 0
@@ -226,7 +245,10 @@ function common::per_dir_hook {
   # run hook for each path
   for dir_path in $(echo "${dir_paths[*]}" | tr ' ' '\n' | sort -u); do
     dir_path="${dir_path//__REPLACED__SPACE__/ }"
-    pushd "$dir_path" > /dev/null || continue
+
+    if [[ $DELEGATE_CHDIR != true ]]; then
+      pushd "$dir_path" > /dev/null || continue
+    fi
 
     per_dir_hook_unique_part "$dir_path" "${args[@]}"
 
@@ -235,7 +257,10 @@ function common::per_dir_hook {
       final_exit_code=$exit_code
     fi
 
-    popd > /dev/null
+    if ! $DELEGATE_CHDIR; then
+      popd > /dev/null
+    fi
+
   done
 
   # restore errexit if it was set before the "for" loop
--- a/hooks/terraform_tflint.sh
+++ b/hooks/terraform_tflint.sh
@@ -28,7 +28,7 @@ function main {
   } || {
     local exit_code=$?
     common::colorify "red" "Command 'tflint --init' failed:"
-    echo "${TFLINT_INIT}"
+    echo -e "${TFLINT_INIT}"
     return ${exit_code}
   }
 
@@ -50,15 +50,19 @@ function per_dir_hook_unique_part {
   shift
   local -a -r args=("$@")
 
-  TFLINT_OUTPUT=$(tflint "${args[@]}" 2>&1)
+  if [[ $DELEGATE_CHDIR == true ]]; then
+    local dir_args="--chdir=$dir_path"
+  fi
+
+  # shellcheck disable=SC2086 # we need to remove the arg if its unset
+  TFLINT_OUTPUT=$(tflint ${dir_args:-} "${args[@]}" 2>&1)
   local exit_code=$?
 
   if [ $exit_code -ne 0 ]; then
     common::colorify "yellow" "TFLint in $dir_path/:"
-    echo "$TFLINT_OUTPUT"
+    echo -e "$TFLINT_OUTPUT"
   fi
 
-  # return exit code to common::per_dir_hook
   return $exit_code
 }
 
--- a/hooks/terraform_tflint.sh
+++ b/hooks/terraform_tflint.sh
@@ -41,16 +41,20 @@ function main {
 # Arguments:
 #   dir_path (string) PATH to dir relative to git repo root.
 #     Can be used in error logging
+#   change_dir_in_unique_part (string/false) Modifier which creates
+#     possibilities to use non-common chdir strategies.
+#     Availability depends on hook.
 #   args (array) arguments that configure wrapped tool behavior
 # Outputs:
 #   If failed - print out hook checks status
 #######################################################################
 function per_dir_hook_unique_part {
   local -r dir_path="$1"
-  shift
+  local -r change_dir_in_unique_part="$2"
+  shift 2
   local -a -r args=("$@")
 
-  if [[ $DELEGATE_CHDIR == true ]]; then
+  if [ "$change_dir_in_unique_part" == "delegate_chdir" ]; then
     local dir_args="--chdir=$dir_path"
   fi
 
--- a/hooks/terraform_tflint.sh
+++ b/hooks/terraform_tflint.sh
@@ -63,6 +63,7 @@ function per_dir_hook_unique_part {
     echo -e "$TFLINT_OUTPUT"
   fi
 
+  # return exit code to common::per_dir_hook
   return $exit_code
 }
 
--- a/hooks/terraform_checkov.sh
+++ b/hooks/terraform_checkov.sh
@@ -31,6 +31,9 @@ function main {
 # Arguments:
 #   dir_path (string) PATH to dir relative to git repo root.
 #     Can be used in error logging
+#   change_dir_in_unique_part (string/false) Modifier which creates
+#     possibilities to use non-common chdir strategies.
+#     Availability depends on hook.
 #   args (array) arguments that configure wrapped tool behavior
 # Outputs:
 #   If failed - print out hook checks status
@@ -38,7 +41,9 @@ function main {
 function per_dir_hook_unique_part {
   # shellcheck disable=SC2034 # Unused var.
   local -r dir_path="$1"
-  shift
+  # shellcheck disable=SC2034 # Unused var.
+  local -r change_dir_in_unique_part="$2"
+  shift 2
   local -a -r args=("$@")
 
   checkov -d . "${args[@]}"
--- a/hooks/terraform_fmt.sh
+++ b/hooks/terraform_fmt.sh
@@ -28,6 +28,9 @@ function main {
 # Arguments:
 #   dir_path (string) PATH to dir relative to git repo root.
 #     Can be used in error logging
+#   change_dir_in_unique_part (string/false) Modifier which creates
+#     possibilities to use non-common chdir strategies.
+#     Availability depends on hook.
 #   args (array) arguments that configure wrapped tool behavior
 # Outputs:
 #   If failed - print out hook checks status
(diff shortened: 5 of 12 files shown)


MaxymVlasov:
Hmm, that's an interesting solution. I definitely like that you find a way to fix that issue, but not sure about realization and our ability to merge it just as a minor release.

Can you please run some performance tests for your realization and for tag 1.77.3?
https://github.com/antonbabenko/pre-commit-terraform/blob/master/.github/CONTRIBUTING.md#run-hook-performance-test

0.44.0 (2022-12-26) was added not so long time ago and that's the main issue from my perspective - many folks could not updated yet, and I don't like to force them to update outside of breaking change release.

So for now - how it when will be merged and released - as 1.x or part of 2.0 - it's a question.

Maybe I need to find some time to resolve https://github.com/antonbabenko/pre-commit-terraform/issues/303 and then I'll be able to move some other hooks to git repo root too, as was asked in #508. With or without breaking changes - will see.

yermulnik:
> Hmm, that's an interesting solution. I definitely like that you find a way to fix that issue, but not sure about realization and our ability to merge it just as a minor release.

What is proposed in this PR contradicts with the essence of `per_dir_hook_unique_part` function and attempts to circumvent author's use case by implicitly manipulating dir stack in reverse order. So it's a bit of reversed copy&pasta from the parent function (also `continue` SHOULD NOT be used outside loops).
This doesn't seem to be a proper and reliable approach as it's more of a circumvention and a quirk. It also doesn't apply any version compatibility checks and hence is backward incompatible (for those who may need to use `tflint =< 0.44.0` by whichever reason).

To properly resolve author's use case, parent `common::per_dir_hook` function should be updated to exclude dir manipulation (the `pushd/popd` thing) for those hook(s) which underlying tool is capable with managing dir stack on their own.
Along with that it would great to maintain backward compatibility for those who may need to use older versions of `tflint` IMHO.

lexton:
Thanks for the feedback! I'll run the performance tests after we decide on a path forward. This is a bit of a hack - just reversing the caller behavior. It seemed to work well enough for me locally - but see the problems you are describing.

I'll adjust to add a `hook-config` that would bypass the `pushd/popd` behavior, and then switch on the `--chdir` flag for `tflint`.

Does a flag like this seem reasonable?
```
--hook-config=--skip--chdir=true
--hook-config=--skip--chdir=false
```

hooks/_common.sh
```diff
@@ -226,7 +245,10 @@ function common::per_dir_hook {
   # run hook for each path
   for dir_path in $(echo "${dir_paths[*]}" | tr ' ' '\n' | sort -u); do
     dir_path="${dir_path//__REPLACED__SPACE__/ }"
-    pushd "$dir_path" > /dev/null || continue
+
+    if ! $SKIP_CHDIR; then
+      pushd "$dir_path" > /dev/null || continue
+    fi
```
MaxymVlasov, line 251: @yermulnik ?

README.md
```diff
@@ -599,6 +599,14 @@ To replicate functionality in `terraform_docs` hook:
         - --args=--config=__GIT_WORKING_DIR__/.tflint.hcl
     ```
 
+3. When you want the root relative paths to the linting error messages rather than just module relative paths. You need to use TFlint native `--chdir` functionality. Note: this requires TFlint version 0.44.0 or later. For example:
+
+    ```yaml
+    - id: terraform_tflint
+          args:
+            - --hook-config=--skip-chdir
```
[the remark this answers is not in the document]
lexton: so the current behavior is to `--enable-chdir=true` - it wouldn't be backwards compatible without setting the default value if its unset, this would require me setting `--enable-chdir=true` in the tflint entrypoint. Is this the approach you are thinking about?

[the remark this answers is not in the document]
yermulnik: Maybe name it `--delegate-chdir` to make it sort of map one-to-one to the essence of this PR?

hooks/_common.sh
```diff
@@ -217,6 +217,25 @@ function common::per_dir_hook {
...
+    key=${config[0]}
+    value=${config[1]}
+
+    case $key in
+      --skip-chdir)
+        # this flag will skip pushing and popping directories
+        # delegating the responsibility to the hooked plugin/binary
+        if [ "$value" != "false" ]; then
+          SKIP_CHDIR=true
+        fi
```
yermulnik, line 234: I'd prefer to compare with truthy value to avoid unexpectedness when user e.g. sets `no` instead of `false` by a mistake. Something like this (using modern `[[` allows to drop double quotes):
```suggestion
        if [[ ! $value || $value == true ]]; then
          SKIP_CHDIR=true
```

hooks/_common.sh
```diff
@@ -226,7 +245,10 @@ function common::per_dir_hook {
   # run hook for each path
   for dir_path in $(echo "${dir_paths[*]}" | tr ' ' '\n' | sort -u); do
     dir_path="${dir_path//__REPLACED__SPACE__/ }"
-    pushd "$dir_path" > /dev/null || continue
+
+    if ! $SKIP_CHDIR; then
```
yermulnik, line 249: Given you define `SKIP_CHDIR` to a non-empty value on line 221 (`local SKIP_CHDIR=false`) the `! $SKIP_CHDIR` would always be false since Bash has no idea about boolean values. Hence need to compare strings explicitly:
```suggestion
    if [[ $SKIP_CHDIR != true ]]; then
```

hooks/terraform_tflint.sh
```diff
@@ -50,16 +50,29 @@ function per_dir_hook_unique_part {
...
-    tflint "${args[@]}"
-  }
+    case $key in
+      --skip-chdir)
+        if [ "$value" != "false" ]; then
+          DIR_ARGS="--chdir=$dir_path"
+        fi
+        ;;
+    esac
+  done
```
yermulnik, line 66: What's the need for this redundancy? Target dir path is passed in from the parent function unconditionally and is stored in `$dir_path` var.

hooks/terraform_tflint.sh
```diff
@@ -50,16 +50,29 @@ function per_dir_hook_unique_part {
...
 
-  # return exit code to common::per_dir_hook
+  COMMAND_OUTPUT=$({ tflint "${DIR_ARGS}" "${args[@]}"; } 2>&1)
   local exit_code=$?
+
+  if [ $exit_code -ne 0 ]; then
+    common::colorify "yellow" "TFLint in $dir_path/:"
+    echo "$COMMAND_OUTPUT"
+  fi
+
```
yermulnik, line 75: Please see https://github.com/antonbabenko/pre-commit-terraform/pull/514#discussion_r1180387879 on how I suggest to handle `tflint` execution.

hooks/_common.sh
yermulnik: replied down the PR

README.md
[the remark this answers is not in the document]
MaxymVlasov: Current behavior in `master` is `--enable-chdir=` which should be interpreted as `--enable-chdir=false`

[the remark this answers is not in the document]
MaxymVlasov: Below in the code right now you specify `local SKIP_CHDIR=false` which is equivalent of `--enable-chdir=true`, which is a Breaking change.

lexton:
Working on addressing these PR comments now - should have it re-pushed/rebased etc in a few mins.

MaxymVlasov:
And merge master branch too, please

hooks/terraform_tflint.sh
lexton: If you are on `tflint<=0.44` it results in a deprecation warning if you try and pass the $dir_path to the function itself. So this is just to do the backwards compatibility piece in a way that doesn't generate new deprecation warnings.

MaxymVlasov:
And you don't need to do force-pushes at all - we will squash it when it will be merged

hooks/terraform_tflint.sh
yermulnik: Since the `--chdir=…` is an arg with an exceptional handling (as in only one of a subset of all `tflint` cmd line args), wouldn't it be simpler to just refer to `$SKIP_CHDIR` var defined by parent function instead of this whole `for` loop?
Sort of this:
```bash
if [[ $SKIP_CHDIR == true ]]; then
  DIR_ARGS="--chdir=$dir_path"
fi
```

lexton:
sorry - not super familiar with the best way to work with forked repos - mostly used internal monorepos. Will re-open in a min - I screwed up the rebase.

hooks/terraform_tflint.sh
```diff
@@ -28,7 +28,7 @@ function main {
   } || {
     local exit_code=$?
     common::colorify "red" "Command 'tflint --init' failed:"
-    echo "${TFLINT_INIT}"
+    echo -e "${TFLINT_INIT}"
```
lexton, line 31: on the other PR this was a comment that we wanted the `-e` to properly escape - did we still want this? Here and below on L63 as well.

hooks/_common.sh
```diff
@@ -217,6 +217,25 @@ function common::per_dir_hook {
...
+  for c in "${configs[@]}"; do
+    IFS="=" read -r -a config <<< "$c"
+    key=${config[0]}
+    value=${config[1]}
+
+    case $key in
+      --delegate-chdir)
+        # this flag will skip pushing and popping directories
+        # delegating the responsibility to the hooked plugin/binary
+        if [[ ! $value || $value == true ]]; then
```
lexton, line 232: I may be missing something in the evaluation here - not super familiar with getopts expected behaviors:

```
--delegate-chdir=true == true 
--delegate-chdir == true
--delegate-chdir=on == false
--delegate-chdir=off == false
--delegate-chdir=no == false
--delegate-chdir=yes == false
--delegate-chdir=false == false
```

hooks/terraform_tflint.sh
yermulnik: If presence of `-e` arg doesn't break UX on different OSes, then I'd better leave it rather than drop.

hooks/_common.sh
yermulnik: My point is that `pre-commit-terraform` should delegate path manipulation to `tflint` only when `--delegate-chdir` has no value set or when `--delegate-chdir` has a value of `true` set. Adding `yes`, `on`, `1` to the list of truthy values might make sense to improve UX though.
The below should demonstrate what I mean:
```bash
#!/usr/bin/env bash

ITEMS=(
        --delegate-chdir=true
        --delegate-chdir
        --delegate-chdir=on
        --delegate-chdir=off
        --delegate-chdir=no
        --delegate-chdir=yes
        --delegate-chdir=false
)

for ITEM in "${ITEMS[@]}"; do
        set -- ${ITEM//=/ }
        key=$1
        value=$2
        if [[
                ! $value || \
                $value == true || \
                $value == yes || \
                $value == on || \
                $value == 1 ]]; then
                printf "%-25s: %s\n" "$ITEM" "true"
        else
                printf "%-25s: %s\n" "$ITEM" "false"
        fi
done
```

hooks/terraform_tflint.sh
yermulnik: Given we already have `echo -e …` used across this file I'd say let's have it in place if it doesn't break printing `${TFLINT_INIT}`

hooks/_common.sh
yermulnik: Let me know if this makes sense.

lexton: I think UX consistency would be preferable to an expanded selector in this one instance. I suspect most folks would just copy a `--delegate-chdir=true/false` straight from the docs and wouldn't think twice about it - but thats just my opinion.

yermulnik: Works for me

yermulnik (commented):
@lexton Could you please test this PR once again and if wit works as expected and as intended for all of `--delegate-chdir`, `--delegate-chdir=true`, `--delegate-chdir=false`, and when `--delegate-chdir` isn't used at all, then I'll be happy to approve.
Thank you

lexton:
Yep - just tested this locally. All these situations worked as expected.

```

Command 'tflint --init' successfully done:

TFLint in mod/:
1 issue(s) found:

variables.tf:1:1: Warning - variable "unused" is declared but not used (terraform_unused_declarations)
```



Interestingly when I ran it locally one of the surprising things:

```
../../private/tmp/new/mod/variables.tf:1:1: Warning - variable "unused" is declared but not used (terraform_unused_declarations)
```

Seems like the symlinking may resolve strangely here - not sure if we want or need to compensate or correct for this.
https://apple.stackexchange.com/questions/1043/why-is-tmp-a-symlink-to-private-tmp

yermulnik:
> Interestingly when I ran it locally one of the surprising things - it seem to resolve symlinks

This time this is for `tflint` not `pre-commit-terraform` 🤣 I'm going to approve PR now. @MaxymVlasov FYI.

yermulnik (approved):
@lexton Thank you for the contribution 🥳

hooks/terraform_tflint.sh
```diff
@@ -50,15 +50,19 @@ function per_dir_hook_unique_part {
...
   local exit_code=$?
 
   if [ $exit_code -ne 0 ]; then
     common::colorify "yellow" "TFLint in $dir_path/:"
-    echo "$TFLINT_OUTPUT"
+    echo -e "$TFLINT_OUTPUT"
   fi
 
-  # return exit code to common::per_dir_hook
   return $exit_code
```
MaxymVlasov, line 66:
```suggestion
  # return exit code to common::per_dir_hook
  return $exit_code
```

hooks/_common.sh
```diff
@@ -217,6 +217,25 @@ function common::per_dir_hook {
...
+          DELEGATE_CHDIR=true
```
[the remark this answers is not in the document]
yermulnik: `break` potentially breaks the idea of `Lookup hook-config for modifiers that impact common behavior` (modifier**s**)

hooks/_common.sh
```diff
@@ -217,6 +217,25 @@ function common::per_dir_hook {
...
+        fi
+        ;;
+    esac
+  done
```
MaxymVlasov, line 237: I don't like that it makes `common::per_dir_hook` more complex.
I thinking about moving it to a separate function but I don't see how it could be done, at least, without introducing a new global env.

The main problem here is `pushd/popd` IF's,

```
 if [[ $DELEGATE_CHDIR != true ]]; then
      pushd "$dir_path" > /dev/null || continue
    fi
```

which, in current realization (internal variable naming?) looks too "tflint-specific"

I suppose `CHANGE_DIR_INSIDE_HOOK` than `DELEGATE_CHDIR` will be more universal and easier to understand

MaxymVlasov: Or at least, I have those feelings. Not sure how they could be properly described and, which more importantly, done via code. 

Let me think about it for a few days, maybe then I can describe it better

yermulnik: I like the `delegate` wording. If `DELEGATE_CHDIR` is too short for clear understanding, then `DELEGATE_CHDIR_TO_HOOK` should suffice imho.

yermulnik: > I thinking about moving it to a separate function but I don't see how it could be done, at least, without introducing a new global env.

Parse `HOOK_CONFIG` array outside of the function and define this function with or w/o `pushd/popd` things based on presence or absence of `--delegate-chdir` hook arg 😜 
This may be expanded later for alike use cases. Though this would add complexity to the code anyways.

yermulnik: I'd leave this as is (the way it's coded by @lexton in this PR) as a one-off/unique instance, at least until we have more of this kind requested.

lexton: I think its relatively well encapsulated as is.

If you wanted to change/extend this in the future it wouldn't be too hard to update/change variable wording without major impact.

The biggest thing that would be hard to change would be the exact flag name. `--delegate-chdir` if you wanted to make this more future proof that makes sense to get correct right now.

yermulnik: > The biggest thing that would be hard to change would be the exact flag name. `--delegate-chdir` if you wanted to make this more future proof that makes sense to get correct right now.

My opinion for the flag is the same as for the var naming. JFYI.

MaxymVlasov: Okay, I double-checked #508 and this functional does not help avoid the mentioned issue, so, I'll make one tiny improvement and we are ready to go

yermulnik (approved):

antonbabenko:
This PR is included in version 1.79.0 :tada:

PR #514, opened 2023-04-28
fix: Speed up x2 TFLint hook execution in dirs with violations

lexton:

- [x] This PR fixes a bug.

Caches the output of the TFLint commands - upon failures the existing behavior was to re-run to present the output. This should speed up performance where there are lots of scattered failures in many directories.

### How can we test changes

I have tested locally with a single failing repo - showed a 48% reduction in wall clock time with the `try-repo` hooks.

Substantial speedup in a monorepo with scattered failures across many directories.

### DIFF
--- a/hooks/terraform_tflint.sh
+++ b/hooks/terraform_tflint.sh
@@ -50,12 +50,12 @@ function per_dir_hook_unique_part {
   shift
   local -a -r args=("$@")
 
-  COMMAND_OUTPUT=$({ tflint "${args[@]}"; } 2>&1)
+  TFLINT_OUTPUT=$({ tflint "${args[@]}"; } 2>&1)
   local exit_code=$?
 
   if [ $exit_code -ne 0 ]; then
     common::colorify "yellow" "TFLint in $dir_path/:"
-    echo "$COMMAND_OUTPUT"
+    echo "$TFLINT_OUTPUT"
   fi
 
   # return exit code to common::per_dir_hook
--- a/hooks/terraform_tflint.sh
+++ b/hooks/terraform_tflint.sh
@@ -50,7 +50,7 @@ function per_dir_hook_unique_part {
   shift
   local -a -r args=("$@")
 
-  TFLINT_OUTPUT=$({ tflint "${args[@]}"; } 2>&1)
+  TFLINT_OUTPUT=$(tflint "${args[@]}" 2>&1)
   local exit_code=$?
 
   if [ $exit_code -ne 0 ]; then
--- a/hooks/terraform_tflint.sh
+++ b/hooks/terraform_tflint.sh
@@ -50,16 +50,15 @@ function per_dir_hook_unique_part {
   shift
   local -a -r args=("$@")
 
-  # Print checked PATH **only** if TFLint have any messages
-  # shellcheck disable=SC2091,SC2068 # Suppress error output
-  $(tflint ${args[@]} 2>&1) 2> /dev/null || {
-    common::colorify "yellow" "TFLint in $dir_path/:"
+  COMMAND_OUTPUT=$({ tflint "${args[@]}"; } 2>&1)
+  local exit_code=$?
 
-    tflint "${args[@]}"
-  }
+  if [ $exit_code -ne 0 ]; then
+    common::colorify "yellow" "TFLint in $dir_path/:"
+    echo "$COMMAND_OUTPUT"
+  fi
 
   # return exit code to common::per_dir_hook
-  local exit_code=$?
   return $exit_code
 }
 


hooks/terraform_tflint.sh
```diff
@@ -50,16 +50,15 @@ function per_dir_hook_unique_part {
   shift
   local -a -r args=("$@")
 
-  # Print checked PATH **only** if TFLint have any messages
-  # shellcheck disable=SC2091,SC2068 # Suppress error output
-  $(tflint ${args[@]} 2>&1) 2> /dev/null || {
-    common::colorify "yellow" "TFLint in $dir_path/:"
+  COMMAND_OUTPUT=$({ tflint "${args[@]}"; } 2>&1)
```
MaxymVlasov, line 53:
```suggestion
  TFLINT_OUTPUT=$({ tflint "${args[@]}"; } 2>&1)
```

hooks/terraform_tflint.sh
```diff
@@ -50,16 +50,15 @@ function per_dir_hook_unique_part {
...
+  local exit_code=$?
 
-    tflint "${args[@]}"
-  }
+  if [ $exit_code -ne 0 ]; then
+    common::colorify "yellow" "TFLint in $dir_path/:"
+    echo "$COMMAND_OUTPUT"
```
MaxymVlasov, line 58:
```suggestion
    echo "$TFLINT_OUTPUT"
```

MaxymVlasov (approved):
Amazing! 
I remember how I spend hours trying to achieve that but with no luck

hooks/terraform_tflint.sh
```diff
@@ -50,16 +50,15 @@ function per_dir_hook_unique_part {
...
-    common::colorify "yellow" "TFLint in $dir_path/:"
+  TFLINT_OUTPUT=$({ tflint "${args[@]}"; } 2>&1)
+  local exit_code=$?
 
-    tflint "${args[@]}"
-  }
+  if [ $exit_code -ne 0 ]; then
+    common::colorify "yellow" "TFLint in $dir_path/:"
+    echo "$TFLINT_OUTPUT"
+  fi
```
yermulnik, line 59: I don't really remember why we didn't go with alike approach in the past, while we did this with `tflint` init part... 🤔

Btw grouping single command (the `{ …; }` thing) looks redundant.

From my point of view (for the sake of consistency and reduction of sort of explicitly independent but implicitly related code bits) the better approach would be to go with the same solution that we use for initializing of `tflint`: https://github.com/antonbabenko/pre-commit-terraform/blob/master/hooks/terraform_tflint.sh#L25-L27
(comments are for our consideration after testing)
```suggestion
  local exit_code
  
  # Do we need to wrap `${args[@]}` into double quotes?
  TFLINT_OUTPUT=$(tflint ${args[@]} 2>&1) exit_code=$? || {
    common::colorify "yellow" "TFLint in $dir_path/:"

    # `-e` is required on some systems which 
    # have `echo` with ascii-sequences turned off by default,
    echo -e "$TFLINT_OUTPUT"
  }
```

MaxymVlasov: @yermulnik  That construction is less readable.

```bash
TFLINT_OUTPUT=$(tflint ${args[@]} 2>&1) exit_code=$? || {}
```

hooks/terraform_tflint.sh
```diff
@@ -50,16 +50,15 @@ function per_dir_hook_unique_part {
   shift
   local -a -r args=("$@")
 
-  # Print checked PATH **only** if TFLint have any messages
-  # shellcheck disable=SC2091,SC2068 # Suppress error output
-  $(tflint ${args[@]} 2>&1) 2> /dev/null || {
-    common::colorify "yellow" "TFLint in $dir_path/:"
+  TFLINT_OUTPUT=$({ tflint "${args[@]}"; } 2>&1)
```
MaxymVlasov, line 53:
```suggestion
  TFLINT_OUTPUT=$(tflint "${args[@]}" 2>&1)
```

hooks/terraform_tflint.sh
MaxymVlasov: >I don't really remember why we didn't go with alike approach in the past

Because errors were shown right after command execution, no matter how I tried to suppress it

MaxymVlasov: And something was wrong with capturing it to variable

lexton: I quite liked the readability of this suggestion here as a format - what specifically was broken?

yermulnik: > That construction is less readable.

Though is a way more durable as e.g. there would be no option to add some more code in between `tflint` execution, `exit_code` definition and `if` conditional block that may accidentally override `exit_code` value.
Up to you to decide though 🤝

yermulnik (approved):

antonbabenko:
This PR is included in version 1.77.4 :tada:

hooks/terraform_tflint.sh
lexton: I wrote a quick script to demonstrate/test this behavior - just for posterity at this point.


``` bash
#!/bin/bash

HOOK_CONFIG+=("--hello;")

TMPFILE=$(mktemp /tmp/test.XXXXXX)
trap "rm -f $TMPFILE" EXIT

cat <<'EOF' > $TMPFILE
status=$1
shift
echo "status: $status"
echo "args: $@"
exit $status
EOF

# sucess case
OUTPUT=$(bash $TMPFILE 0 "${HOOK_CONFIG[@]}") exit_code=$? || {
    echo "command failed with exit code $exit_code"
    echo -e "$OUTPUT"
}

HOOK_CONFIG+=("--failing;")

# fail case
OUTPUT=$(bash $TMPFILE 1 "${HOOK_CONFIG[@]}") exit_code=$? || {
    echo "command failed with exit code $exit_code"
    echo -e "$OUTPUT"
}

echo "done"
```


Output:
```
$ pre-commit-terraform git:(master) bash /tmp/test.sh
command failed with exit code 1
status: 1
args: --hello; --world; --failing;
done
```

yermulnik: > The big gaps in this are that you can't trivially write `exit_code` as a `local` - and it wouldn't work if you `export OUTPUT=...`

I'm not sure I got your point, though 1) "you can't trivially write `exit_code` as a `local`" — this is why the very first line in my code suggestion (the line to define `exit_code` as `local` var) and 2) why would we need to export output?

lexton: Yeah. Just ways that you could accidentally swallow the errors and break the chaining here. 

Not that it's important at this point. I like the current implantation. 

Thanks for the really thorough reviews!

PR #519, opened 2023-05-07
docs: Add arm to supported architectures

MaxymVlasov:


yermulnik (approved):

VictoriaMOtus:
The text for the sub point is unreadable on the README.md, tested on Firefox and Chrome.

image

Also these might be better under a section about "Requirements" rather than "Install deps" as these are not installable:
 -Internet connection (on first run),
 -Some hardware where this OS will run,
 -Electricity for hardware and internet connection,
 -Some basic physical laws,

MaxymVlasov:
@VictoriaMOtus  Why are you so serious? :)

>The text for the sub point is unreadable on the README.md, tested on Firefox and Chrome.

They shouldn't be readable by default, because these deps are more _easter egg_ than something that really needs attention.

Let me explain: this repo is named `pre-commit-terraform`.

* Which, by default means, that for its works it needs `pre-commit` and `terraform` installed. 
* `pre-commit` by itself needs `git`, because its git-hooks framework
) POSIX compatible shell - it is basically any SHELL in MacOS or Linux. And we don't support Windows-native shells as described here:
    image

* Internet connection (on first run) - you see this repo => you have an internet connection most of time. Ofc, if you not in a very boring corporation
*  x86_64 or arm64 compatible operation system - any other arch now used for personal use/in CI? Maybe in 1% of cases, which we can ignore the same as boring corps fro prev point.
* Some hardware where this OS will run, Electricity for hardware and internet connection, Some basic physical laws, Hope that it all will work - no comments :D



> Also these might be better under a section about "Requirements" rather than "Install deps" as these are not installable:
> -Internet connection (on first run),
> -Some hardware where this OS will run,
> -Electricity for hardware and internet connection,
> -Some basic physical laws,

I don't think so, see the explanation above. Also, ` as these are not installable` is not true, because:

* `Internet connection` can be installed by ISP to your location, so it's installable 
* `Some hardware where this OS will run` - hardware can be bought and installed too to your workplace
* `Electricity for hardware and internet connection` - there is more installed by local government and ISP than by users, but all these cables, transformations, and so on are physical stuff that can be installed and can be described as "Electricity"  in the first approximation
* Some basic physical laws - well, those dependencies are definitely preinstalled on our planet, so there should not be problems at all

By the way, have a nice day :)

antonbabenko:
This PR is included in version 1.79.0 :tada:

Issue #526, opened 2023-05-25
feat(`terraform_providers_lock`): Run only when non-cross-platform lockfile found


MaxymVlasov:
### What problem are you facing?

Running `terraform_providers_lock` for each provider for each Architecture tooks about 2 seconds.
We have 4 arches (`linux amd64`, `linux arm64`, `mac amd64`, `mac arm64`) and up to 14 providers.
That took about 2-3 minutes to check the lockfile, which it too much.

I found that during `terraform init`, lockfile not touched in case all providers already here. Each provider block looks like:

 ```hcl
provider "registry.terraform.io/okta/okta" {
  version     = "3.46.0"
  constraints = ">= 3.0.0, ~> 3.38"
  hashes = [
    "h1:8AQNT9HErs8LdGbkU+X7U7ggJtbZuJ15hLIfG+tNxzs=",
    "h1:Ab6Zjnf0jzF8GAk11QLRvOTp37DaavzNLQEpVePZVKg=",
    "h1:TtoJ+D96IIxr4660CGZeTd3VQQCkGZHHgSylhsy/B2A=",
    "h1:efq7Hv5KDfuIIffVSi541msYTcv4HmID5xzVO8ZJzdc=",
    "zh:033e9c282bfb1a47be6701b2d56da14c9cb214079f445adf0a3cf2be3bf546e4",
    "zh:0a49da96635c42cc118e507c171e2fb228d8789a9eeaf3de4b01cd47a624751a",
    "zh:1103e29b144b6f5ec607a32988bc490b661909b86ed2c37d09a08e8cff32a99c",
    "zh:824cdf825f7bf6b367213969f167d39bc0b27fcd5bff56031ee720f89df79714",
    "zh:a8f1f19698c6e0d4b53655d5c55c821398805549047c493e264cdfdf58d9b053",
    "zh:aefd9fc1d88ad2e4990dc52ed470f5d61f115b071897bc91fb5e54311d3599ea",
    "zh:c501aad6dbeeb262757a235ae386d59f758dabc609d545d2c766b4b5ce951859",
    "zh:c67a663e796f425fb1daa6281e726c025aa7199d50c7e93603d4a853ca56282d",
    "zh:d0e2af10e773063b587a2d3acf2317d1eacb68f091b85b124c05dab88a6966bf",
    "zh:de06df4e14817049acbb800c09de702708557cae1a0727f4248b18ff927df664",
[...]

PR #527, opened 2023-05-26
docs: Simplify lists maintenance and fix language

MaxymVlasov:

Quick improvements, which were done during work on #526, but were separated to this PR as not related changes

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -105,7 +105,7 @@ You can use [this PR](https://github.com/antonbabenko/pre-commit-terraform/pull/
 ### Before write code
 
 1. Try to figure out future hook usage.
-2. Confirm the concept with [Anton Babenko](https://github.com/antonbabenko).
+1. Confirm the concept with [Anton Babenko](https://github.com/antonbabenko).
 
 ### Prepare basic documentation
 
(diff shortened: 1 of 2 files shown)


.github/CONTRIBUTING.md
```diff
@@ -105,7 +105,7 @@ You can use [this PR](https://github.com/antonbabenko/pre-commit-terraform/pull/
 ### Before write code
 
 1. Try to figure out future hook usage.
-2. Confirm the concept with [Anton Babenko](https://github.com/antonbabenko).
+1. Confirm the concept with [Anton Babenko](https://github.com/antonbabenko).
```
yermulnik, line 108: Any **sensible** reason to use this type of how numbered lists are rendered? If I wanted to edit markdown file as plain text in terminal editor, I'd expect it to use natural numbering rather than what markdown viewer would take as a numbered list. We even don't need to bother about human mistakes: markdown renders into numbered list even lists with random numbering and the only requirement is that first item in the list should be "number one with trailing dot".

MaxymVlasov: > I got tired of changing numbers every time something is added or rearranged somewhere

> and in general, I don’t care what the number is when you read the source code, because as a docs you will still watch the rendered md

So better use `1.` to be consistent, than have `1. 5. 2. 3. 7.` list

yermulnik: > I got tired of changing numbers every time something is added or rearranged somewhere

It's a bit odd your IDE can't handle lists renumbering automatically. Poor you, though it is not the reason for odd lists with the same number for all items 🤷🏻

> I don’t care what the number is when you read the source code, because as a docs you will still watch the rendered md

Well, I might be missing something, though what I do for living is not just reading docs, though writing code too.

> So better use 1. to be consistent, than have 1. 5. 2. 3. 7. list

Given your attitude and use case, I'd suggest you just use unordered lists rather than neglect common sense 😺

yermulnik:
> @MaxymVlasov MaxymVlasov (Maksym Vlasov) requested a review from yermulnik 1 hour ago

Spelling/language fixes look good to me. Though changes to lists — do not. Sorry 🤷🏻

.github/CONTRIBUTING.md
MaxymVlasov: An unordered list can't be mentioned when you point some to README. Now I start thinking about adding anchors to each point to be able to point folks exactly to what they need to read, and then we can change many lists to unordered. But I'm not sure it's worth it. In any case, if it will be done - it out of scope of this PR

yermulnik (approved):

antonbabenko:
This PR is included in version 1.80.0 :tada:

PR #552, opened 2023-08-10
fix: Fix the terraform_wrapper_module_for_each hook for modules without outputs or variables

ajax-ryzhyi-r:
…do not have outputs or variables

Put an `x` into the box if that apply:

- [ ] This PR introduces breaking change.
- [x] This PR fixes a bug.
- [ ] This PR adds new functionality.
- [ ] This PR enhances existing functionality.

### Description of your changes
The `terraform_wrapper_module_for_each` hook always exits with `1` code for modules that do not have outputs or variables defined. This is because `grep` returns an exit code of `1` when it does not find a match, and `pipefail` is set for the hook script.

This pull request resolves the issue by setting a `0` exit code for variables and outputs `grep` executions.

### How can we test changes

Run the `terraform_wrapper_module_for_each hook` against the module without outputs or variables. 🙃

PR #554, opened 2023-08-10
fix: Fix terraform_wrapper_module_for_each hook heredoc vars defaults

ajax-ryzhyi-r:

- [x] This PR fixes a bug.

Fixes #552 comments

If default variable values are written in heredoc style, the hook produces an error:

```
Terraform wrapper with for_each in module................................Failed
- hook id: terraform_wrapper_module_for_each
- exit code: 1

failed to build expression at the parse phase: failed to parse input: generated_by_buildExpression:21,7-7: Unterminated template string; No closing marker was found for the string.

```

This error occurs because the heredoc syntax requires the closing marker to be alone on its own line. This pull request adds a new line at the end for heredoc default values.

Also, this pull request addresses @yermulnik's comment in issue #552.

### How can we test changes

Changes can be tested against any module that has heredoc default values for variables. For example, the [terraform-aws-appsync](https://github.com/terraform-aws-modules/terraform-aws-appsync/) module.

### DIFF
--- a/hooks/terraform_wrapper_module_for_each.sh
+++ b/hooks/terraform_wrapper_module_for_each.sh
@@ -323,11 +323,11 @@ EOF
 
     # Get names of module variables in all terraform files
     # shellcheck disable=SC2207
-    module_vars=($(echo "$all_tf_content" | hcledit block list | { grep variable. || true; } | cut -d'.' -f 2))
+    module_vars=($(echo "$all_tf_content" | hcledit block list | { grep variable. | cut -d'.' -f 2 || true; }))
 
     # Get names of module outputs in all terraform files
     # shellcheck disable=SC2207
-    module_outputs=($(echo "$all_tf_content" | hcledit block list | { grep output. || true; } | cut -d'.' -f 2))
+    module_outputs=($(echo "$all_tf_content" | hcledit block list | { grep output. | cut -d'.' -f 2 || true; }))
 
     # Looking for sensitive output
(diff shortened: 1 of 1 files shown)


hooks/terraform_wrapper_module_for_each.sh
```diff
@@ -360,6 +360,13 @@ EOF
         # https://github.com/terraform-aws-modules/terraform-aws-security-group/blob/0bd31aa88339194efff470d3b3f58705bd008db0/rules.tf#L8
         # As a result, wrappers in terraform-aws-security-group module are missing values of the rules variable and is not useful. :(
         var_value="try(each.value.${module_var}, var.defaults.${module_var}, {})"
+      elif [[ $var_default == *EOF || $var_default == *EOT ]]; then
```
antonbabenko, line 363: Let's check for the heredoc market (`<<`) instead of the heredoc identifier (EOF, EOT,...) since market can't be changed.

antonbabenko (changes requested):
Looks pretty good, and it works :) A slight improvement is missing.

hooks/terraform_wrapper_module_for_each.sh
ajax-ryzhyi-r: > Let's check for the heredoc market (`<<`) instead of the heredoc identifier (EOF, EOT,...) since market can't be changed.

Yep, you're right. Thanks

antonbabenko (approved):

antonbabenko:
This PR is included in version 1.81.2 :tada:

PR #560, opened 2023-08-25
feat: Skip legacy modules (with provider block) in terraform_wrapper_module_for_each hook

ajax-ryzhyi-r:

- [x] This PR fixes a bug.

Modules that have `provider` blocks do not support `for_each`, `count`, or `depends_on` meta arguments. Therefore, there is no need to generate wrappers for these modules because they cannot be called.

This PR adds a skip for modules that contain a `provider` block.

### How can we test changes

Run the `terraform_wrapper_module_for_each` against the module with the `provider` block - wrapper should not be created.

### DIFF
--- a/hooks/terraform_wrapper_module_for_each.sh
+++ b/hooks/terraform_wrapper_module_for_each.sh
@@ -291,8 +291,6 @@ EOF
     # Wrappers will be stored in "wrappers/{module_name}"
     output_dir="${root_dir}/${wrapper_dir}/${module_name}"
 
-    [[ ! -d "$output_dir" ]] && mkdir -p "$output_dir"
-
     # Calculate relative depth for module source by number of slashes
     module_depth="${module_dir//[^\/]/}"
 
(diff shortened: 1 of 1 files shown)


antonbabenko:
Thank you for your contribution. Could you please provide a small code snippet of the module with a `provider` block which should not be generated by the hook? And/or link to the related issue in Terraform?

ajax-ryzhyi-r:
> Thank you for your contribution. Could you please provide a small code snippet of the module with a `provider` block which should not be generated by the hook? And/or link to the related issue in Terraform?

This is not actually an issue; it is expected behavior. Modules with local provider definitions are considered legacy modules by HashiCorp and do not support the `for_each` meta argument. However, in rare cases, there may be a need to add some provider configurations in the module. In such cases, successfully generated wrappers could be confusing because they cannot be used.

Terraform docs: https://developer.hashicorp.com/terraform/language/modules/develop/providers#legacy-shared-modules-with-provider-configurations

antonbabenko (approved):

antonbabenko:
Thank you for the explanation! Merging...

antonbabenko:
This PR is included in version 1.83.0 :tada:

PR #565, opened 2023-09-04
fix: Fixed ordering issue in terraform_wrapper_module_for_each hook

ajax-ryzhyi-r:
Put an x into the box if that apply:

- [x] This PR fixes a bug.

I have a module with variables defined in two `.tf` files. When I run the `terraform_wrapper_module_for_each hook` locally on macOS, no changes are shown. However, when the same hooks run against the same module by GitHub Actions on Ubuntu, a difference in the order of the wrapper module input variables is shown. After researching the root cause of this behavior, I discovered that the `find` utility returns files in a different order on macOS and Ubuntu because these OS use different virtual file systems:
https://unix.stackexchange.com/questions/13451/what-is-the-directory-order-of-files-in-a-directory-used-by-ls-u

To solve this issue, I added sorting for the outputs and variables before appending them to the wrapper module.

### How can we test changes

Run the hooks against module with the next structure in MacOS and Ubuntu:
```
main.tf
variables.tf             # Some variables are defined here
additional_variables.tf  # Some variables are defined here

### DIFF
--- a/hooks/terraform_wrapper_module_for_each.sh
+++ b/hooks/terraform_wrapper_module_for_each.sh
@@ -325,7 +325,7 @@ EOF
 
     # Get names of module outputs in all terraform files
     # shellcheck disable=SC2207
-    module_outputs=($(echo "$all_tf_content" | hcledit block list | { grep output. | cut -d'.' -f 2 | sort || true; }))
+    module_outputs=($(echo "$all_tf_content" | hcledit block list | { grep output. | cut -d'.' -f 2 || true; }))
 
     # Get names of module providers in all terraform files
     module_providers=$(echo "$all_tf_content" | hcledit block list | { grep provider. || true; })
--- a/hooks/terraform_wrapper_module_for_each.sh
+++ b/hooks/terraform_wrapper_module_for_each.sh
@@ -321,11 +321,11 @@ EOF
 
     # Get names of module variables in all terraform files
     # shellcheck disable=SC2207
-    module_vars=($(echo "$all_tf_content" | hcledit block list | { grep variable. | cut -d'.' -f 2 || true; }))
+    module_vars=($(echo "$all_tf_content" | hcledit block list | { grep variable. | cut -d'.' -f 2 | sort || true; }))
 
     # Get names of module outputs in all terraform files
     # shellcheck disable=SC2207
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: hooks/terraform_wrapper_module_for_each.sh [hunk 1]
```diff
@@ -321,11 +321,11 @@ EOF
...
-    module_vars=($(echo "$all_tf_content" | hcledit block list | { grep variable. | cut -d'.' -f 2 || true; }))
+    module_vars=($(echo "$all_tf_content" | hcledit block list | { grep variable. | cut -d'.' -f 2 | sort || true; }))
 
     # Get names of module outputs in all terraform files
     # shellcheck disable=SC2207
-    module_outputs=($(echo "$all_tf_content" | hcledit block list | { grep output. | cut -d'.' -f 2 || true; }))
+    module_outputs=($(echo "$all_tf_content" | hcledit block list | { grep output. | cut -d'.' -f 2 | sort || true; }))
 
     # Get names of module providers in all terraform files
     module_providers=$(echo "$all_tf_content" | hcledit block list | { grep provider. || true; })
```
yermulnik, line 331: @ajax-ryzhyi-r Would it make sense to add sorting to this var definition as well for consistency?

ajax-ryzhyi-r: > @ajax-ryzhyi-r Would it make sense to add sorting to this var definition as well for consistency?

This variable is only used to check if the module contains provider definitions, so the ordering does not matter. However, for the sake of code consistency, sorting can be added here as well.

antonbabenko: Good idea to add `sort`, but it is enough to have it just in `module_vars` as it is the only variable used in a loop when making a file.

Also, `tflint` may help with this. It would complain if `variable` blocks were defined in various places.

ajax-ryzhyi-r: > Good idea to add `sort`, but it is enough to have it just in `module_vars` as it is the only variable used in a loop when making a file.
> 
> Also, `tflint` may help with this. It would complain if `variable` blocks were defined in various places.

Yes, you're right. The order of outputs does not matter too. Thank you for pointing that out.

I am aware that `tflint` can detect such cases, but we still have some legacy modules where this structure is expected 😢

antonbabenko (approved):

PR #573, opened 2023-09-21
fix: Fix terraform_wrapper_module_for_each for when resource name contains 'variable'

nshenry03:

- [x] This PR fixes a bug.

There are some providers with resources and data sources that have "variable" in their name, for example:

- [env0_configuration_variable](https://registry.terraform.io/providers/env0/env0/latest/docs/resources/configuration_variable)
- [env0_configuration_variable](https://registry.terraform.io/providers/env0/env0/latest/docs/data-sources/configuration_variable) data source

When using such resources, `terraform_wrapper_module_for_each` was adding these as variables to pass in `wrappers/main.tf` which breaks the wrapper 

> ╷
> │ Error: Unsupported argument
> │ 
> │   on main.tf line 14, in module "wrapper":
> │   14:   env0_configuration_variable              = try(each.value.env0_configuration_variable, var.defaults.env0_configuration_variable)
> │ 
> │ An argument named "env0_configuration_variable" is not expected here.
> ╵
> Validation failed: wrappers

My fix simply improves `grep` to ensure that `module_vars` (and `module_outputs` and `module_providers`) only find actual variables, outputs, and providers and NOT resources.

### How can we test changes

Use the examples provided in the documentation for [env0_configuration_variable](https://registry.terraform.io/providers/env0/env0/latest/docs/resources/configuration_variable) and try `terraform_wrapper_module_for_each` as-is today.  You'll see that it adds these as variables to pass to the module when it shouldn't.

Now test with my fix and you'll see that it fixes the issue.

### DIFF
--- a/hooks/terraform_wrapper_module_for_each.sh
+++ b/hooks/terraform_wrapper_module_for_each.sh
@@ -321,14 +321,14 @@ EOF
 
     # Get names of module variables in all terraform files
     # shellcheck disable=SC2207
-    module_vars=($(echo "$all_tf_content" | hcledit block list | { grep variable. | cut -d'.' -f 2 | sort || true; }))
+    module_vars=($(echo "$all_tf_content" | hcledit block list | { grep "^variable\." | cut -d'.' -f 2 | sort || true; }))
 
     # Get names of module outputs in all terraform files
     # shellcheck disable=SC2207
-    module_outputs=($(echo "$all_tf_content" | hcledit block list | { grep output. | cut -d'.' -f 2 || true; }))
+    module_outputs=($(echo "$all_tf_content" | hcledit block list | { grep "^output\." | cut -d'.' -f 2 || true; }))
 
     # Get names of module providers in all terraform files
-    module_providers=$(echo "$all_tf_content" | hcledit block list | { grep provider. || true; })
+    module_providers=$(echo "$all_tf_content" | hcledit block list | { grep "^provider\." || true; })
 
     if [[ $module_providers ]]; then
       common::colorify "yellow" "Skipping ${full_module_dir} because it is a legacy module which contains its own local provider configurations and so calls to it may not use the for_each argument."


hooks/terraform_wrapper_module_for_each.sh
```diff
@@ -321,14 +321,14 @@ EOF
 
     # Get names of module variables in all terraform files
     # shellcheck disable=SC2207
-    module_vars=($(echo "$all_tf_content" | hcledit block list | { grep variable. | cut -d'.' -f 2 | sort || true; }))
+    module_vars=($(echo "$all_tf_content" | hcledit block list | { grep "^\s*variable." | cut -d'.' -f 2 | sort || true; }))
```
antonbabenko, line 324:
```suggestion
    module_vars=($(echo "$all_tf_content" | hcledit block list | { grep "^variable." | cut -d'.' -f 2 | sort || true; }))
```
This should be enough. I don't think there can/should be any space before the "variable".

Please apply the same fix on other lines in this PR.

antonbabenko (changes requested):

hooks/terraform_wrapper_module_for_each.sh
yermulnik: > I don't think there can/should be any space before the "variable".

Leading spaces do not produce errors, trigger failures and are not otherwise disallowed explicitly. Hence should be considered allowed imho.
On the other end `hcledit block list` seems to strip leading spaces, which means Anton's recommendation is correct =)

yermulnik: I also would like to recommend to escape dot to not mix it with "any char" in this use case:
```suggestion
    module_vars=($(echo "$all_tf_content" | hcledit block list | { grep "^variable\." | cut -d'.' -f 2 | sort || true; }))
```

yermulnik: And as Anton mentioned: please apply the same to other code bits in this PR.

nshenry03: Done! Please resolve if this looks sufficient. Thanks!

yermulnik (approved):

antonbabenko (approved):

antonbabenko:
This PR is included in version 1.83.4 :tada:

PR #581, opened 2023-10-29
Add terraform and terragrunt testing

### LINKED ISSUE #549: Support for `terraform test`

I want to run [`terraform test`](https://developer.hashicorp.com/terraform/language/modules/testing-experiment) in a pre-commit hook if either the module or one of the module's tests changed.

`terraform test` has been around since at least 2021 and became quite usable with the `precondition` and `postcondition` blocks introduced in terraform 1.2.0, but is nevertheless still a "experimental" feature of the terraform CLI.

In principle `terraform test` is "just the same" as the `terraform validate` hook, except that it should be only triggered if either `dir_path` has a parent directory that is called `tests` (i.e. there is a changed test in the commit) or `dir_path` has a child directory that is called `tests` (i.e. there is a change in the module). Either way, `terraform test` would be run from the terraform module root directory and it should fail the commit if the tests do not pass.

hoangelos:

- [ ] This PR fixes a bug.
- [X ] This PR adds new functionality.

This runs terraform test or terra grunt test if tf files are updated.

Fixes #549

### How can we test changes

Run this on terraform repo that uses testing framework GA in 1.6.x

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -90,7 +90,11 @@
   description: Runs all terraform tests
   entry: hooks/terragrunt_test.sh
   language: script
+<<<<<<< HEAD
   files: (\.tf)$d
+=======
+  files: (\.tf)$
+>>>>>>> c943481 (Add terraform and terragrunt testing)
   exclude: \.terraform\/.*$
 
 - id: terragrunt_validate
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -77,6 +77,22 @@
   files: (\.hcl)$
   exclude: \.terraform\/.*$
 
+- id: terraform_test
+  name: Terraform test
+  description: Runs all terraform tests
+  entry: hooks/terraform_test.sh
+  language: script
+  files: (\.tf)$
+  exclude: \.terraform\/.*$
+
+- id: terragrunt_test
+  name: Terragrunt test
+  description: Runs all terraform tests
+  entry: hooks/terragrunt_test.sh
+  language: script
+  files: (\.tf)$d
+  exclude: \.terraform\/.*$
+
 - id: terragrunt_validate
   name: Terragrunt validate
   description: Validates all Terragrunt configuration files.
--- a/hooks/terragrunt_test.sh
+++ b/hooks/terragrunt_test.sh
@@ -1,0 +1,67 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+# globals variables
+# shellcheck disable=SC2155 # No way to assign to readonly variable in separate lines
+readonly SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)"
+# shellcheck source=_common.sh
+. "$SCRIPT_DIR/_common.sh"
+
+function main {
+  common::initialize "$SCRIPT_DIR"
+  common::parse_cmdline "$@"
+  common::export_provided_env_vars "${ENV_VARS[@]}"
+  common::parse_and_export_env_vars
+  # JFYI: terragrunt test color already suppressed via PRE_COMMIT_COLOR=never
+
+  # shellcheck disable=SC2153 # False positive
+  common::per_dir_hook "$HOOK_ID" "${#ARGS[@]}" "${ARGS[@]}" "${FILES[@]}"
+}
+
+#######################################################################
+# Unique part of `common::per_dir_hook`. The function is executed in loop
+# on each provided dir path. Run wrapped tool with specified arguments
+# Arguments:
+#   dir_path (string) PATH to dir relative to git repo root.
+#     Can be used in error logging
+#   change_dir_in_unique_part (string/false) Modifier which creates
+#     possibilities to use non-common chdir strategies.
+#     Availability depends on hook.
+#   args (array) arguments that configure wrapped tool behavior
+# Outputs:
+#   If failed - print out hook checks status
+#######################################################################
+function per_dir_hook_unique_part {
+  # shellcheck disable=SC2034 # Unused var.
+  local -r dir_path="$1"
+  # shellcheck disable=SC2034 # Unused var.
+  local -r change_dir_in_unique_part="$2"
+  shift 2
+  local -a -r args=("$@")
+
+  # pass the arguments to hook
+  terragrunt test "${args[@]}"
+
+  # return exit code to common::per_dir_hook
+  local exit_code=$?
+  return $exit_code
+}
+
+#######################################################################
+# Unique part of `common::per_dir_hook`. The function is executed one time
+# in the root git repo
+# Arguments:
+#   args (array) arguments that configure wrapped tool behavior
+#######################################################################
+function run_hook_on_whole_repo {
+  local -a -r args=("$@")
+
+  # pass the arguments to hook
+  terragrunt test "$(pwd)" "${args[@]}"
+
+  # return exit code to common::per_dir_hook
+  local exit_code=$?
+  return $exit_code
+}
+
+[ "${BASH_SOURCE[0]}" != "$0" ] || main "$@"
--- a/hooks/terraform_test.sh
+++ b/hooks/terraform_test.sh
@@ -1,0 +1,54 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+# globals variables
+# shellcheck disable=SC2155 # No way to assign to readonly variable in separate lines
+readonly SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)"
+# shellcheck source=_common.sh
+. "$SCRIPT_DIR/_common.sh"
+
+function main {
+  common::initialize "$SCRIPT_DIR"
+  common::parse_cmdline "$@"
+  common::export_provided_env_vars "${ENV_VARS[@]}"
+  common::parse_and_export_env_vars
+
+  # Suppress terraform test color
+  if [ "$PRE_COMMIT_COLOR" = "never" ]; then
+    ARGS+=("-no-color")
+  fi
+
+  # shellcheck disable=SC2153 # False positive
+  common::per_dir_hook "$HOOK_ID" "${#ARGS[@]}" "${ARGS[@]}" "${FILES[@]}"
+}
+
+#######################################################################
+# Unique part of `common::per_dir_hook`. The function is executed in loop
+# on each provided dir path. Run wrapped tool with specified arguments
+# Arguments:
+#   dir_path (string) PATH to dir relative to git repo root.
+#     Can be used in error logging
+#   change_dir_in_unique_part (string/false) Modifier which creates
+#     possibilities to use non-common chdir strategies.
+#     Availability depends on hook.
+#   args (array) arguments that configure wrapped tool behavior
+# Outputs:
+#   If failed - print out hook checks status
+#######################################################################
+function per_dir_hook_unique_part {
+  # shellcheck disable=SC2034 # Unused var.
+  local -r dir_path="$1"
+  # shellcheck disable=SC2034 # Unused var.
+  local -r change_dir_in_unique_part="$2"
+  shift 2
+  local -a -r args=("$@")
+
+  # pass the arguments to hook
+  terraform test "${args[@]}"
+
+  # return exit code to common::per_dir_hook
+  local exit_code=$?
+  return $exit_code
+}
+
+[ "${BASH_SOURCE[0]}" != "$0" ] || main "$@"


.pre-commit-hooks.yaml
```diff
@@ -77,6 +77,26 @@
...
+- id: terragrunt_test
+  name: Terragrunt test
+  description: Runs all terraform tests
+  entry: hooks/terragrunt_test.sh
+  language: script
+<<<<<<< HEAD
+  files: (\.tf)$d
+=======
+  files: (\.tf)$
+>>>>>>> c943481 (Add terraform and terragrunt testing)
```
yermulnik, line 97: Please resolve this merge conflict.

hooks/terragrunt_test.sh
```diff
@@ -0,0 +1,67 @@
...
+  return $exit_code
+}
+
+#######################################################################
+# Unique part of `common::per_dir_hook`. The function is executed one time
+# in the root git repo
+# Arguments:
+#   args (array) arguments that configure wrapped tool behavior
+#######################################################################
+function run_hook_on_whole_repo {
```
yermulnik, line 56: I'm wondering why the Terraform counterpart has no `run_hook_on_whole_repo` function defined — is this intentional?

yermulnik:
@hoangelos Thanks for the contribution 👍🏻
I have several considerations re this change:
1. I think running `terraform test` command makes sense only when there are TF Test files in the directory: https://developer.hashicorp.com/terraform/language/tests#syntax — «_Each Terraform test lives in a test file. Terraform discovers test files are based on their file extension: `.tftest.hcl` or `.tftest.json`_». This way we'd also ensure support for TF versions prior to Terraform v1.6.0
2. I'm not sure about having Terragrunt specific hook for the TF tests since `test` command doesn't belong in Terragrunt and hence gets forwarded by Terragrunt to Terraform: https://terragrunt.gruntwork.io/docs/reference/cli-options/#all-terraform-built-in-commands — «_Terragrunt is a thin wrapper for Terraform, so except for a few of the special commands defined in these docs, Terragrunt forwards all other commands to Terraform_». Having standalone Terragrunt Tests hook makes only sense if there might be some use cases when Terrafrunt-specific command line options need to be passed to the hook (I'm not Terragrunt user, thus I'm can say for sure if there are any use cases of such kind).

hooks/terragrunt_test.sh
yermulnik: This difference only makes sense if `terragrunt test` can run recursively. I'm not TG user and thus I don't know whether it can or not.

.pre-commit-hooks.yaml
```diff
@@ -77,6 +77,26 @@
   files: (\.hcl)$
   exclude: \.terraform\/.*$
 
+- id: terraform_test
+  name: Terraform test
+  description: Runs all terraform tests
+  entry: hooks/terraform_test.sh
+  language: script
+  files: (\.tf)$
```
yermulnik, line 85: It's probably worth of triggering TF Tests on `tfvars` files change also. And maybe even on TF Lock file changes if present.

MaxymVlasov:
Plus add usage examples to README as it done for other hooks

.pre-commit-hooks.yaml
MaxymVlasov: Yeah
```suggestion
  files: (\.tf(vars)?(\.json)?|\.terraform\.lock\.hcl)$
```

hooks/terragrunt_test.sh
MaxymVlasov: That's defenetly copypasted from 
https://github.com/antonbabenko/pre-commit-terraform/blob/7122298824a9530e8f9c48e5de4281cca6301d67/hooks/terragrunt_validate.sh#L60

But then this command should be `terragrunt run-all test` not `terragrunt test`

hoangelos:
> @hoangelos Thanks for the contribution 👍🏻 I have several considerations re this change:
> 
> 1. I think running `terraform test` command makes sense only when there are TF Test files in the directory: https://developer.hashicorp.com/terraform/language/tests#syntax — «_Each Terraform test lives in a test file. Terraform discovers test files are based on their file extension: `.tftest.hcl` or `.tftest.json`_». This way we'd also ensure support for TF versions prior to Terraform v1.6.0



> 2. I'm not sure about having Terragrunt specific hook for the TF tests since `test` command doesn't belong in Terragrunt and hence gets forwarded by Terragrunt to Terraform: https://terragrunt.gruntwork.io/docs/reference/cli-options/#all-terraform-built-in-commands — «_Terragrunt is a thin wrapper for Terraform, so except for a few of the special commands defined in these docs, Terragrunt forwards all other commands to Terraform_». Having standalone Terragrunt Tests hook makes only sense if there might be some use cases when Terrafrunt-specific command line options need to be passed to the hook (I'm not Terragrunt user, thus I'm can say for sure if there are any use cases of such kind).



> @hoangelos Thanks for the contribution 👍🏻 I have several considerations re this change:
> 
> 1. I think running `terraform test` command makes sense only when there are TF Test files in the directory: https://developer.hashicorp.com/terraform/language/tests#syntax — «_Each Terraform test lives in a test file. Terraform discovers test files are based on their file extension: `.tftest.hcl` or `.tftest.json`_». This way we'd also ensure support for TF versions prior to Terraform v1.6.0

terraform test looks for`.tftest.hcl` or `.tftest.json`_». in the test directory, which by default is `tests`.    I do agree that it's likely we don't want to do this per changed file, but per repo.  


> 2. I'm not sure about having Terragrunt specific hook for the TF tests since `test` command doesn't belong in Terragrunt and hence gets forwarded by Terragrunt to Terraform: https://terragrunt.gruntwork.io/docs/reference/cli-options/#all-terraform-built-in-commands — «_Terragrunt is a thin wrapper for Terraform, so except for a few of the special commands defined in these docs, Terragrunt forwards all other commands to Terraform_». Having standalone Terragrunt Tests hook makes only sense if there might be some use cases when Terrafrunt-specific command line options need to be passed to the hook (I'm not Terragrunt user, thus I'm can say for sure if there are any use cases of such kind).

So, your ignorance of Terragrunt has already created a gap in the hooks.  Wrapper is a correct, but not fully accurate.  As a wrapper, it doesn't "otters things" before it passes things along to terraform.  So for example, you can "generate" files that get created the same for different environments, like providers or versions or backends.  And then you might run a fmt on everything, which would catch the formatting issues in the generated files too.... However, terra grunt fmt hook only does hclfmt, which doesn't actually run any terraform fmt, which leaves these generated files as a gap.  That's a long way to say, technically most of the terraform commands will need separate terragrunt commands too.

hoangelos:
I removed the terra grunt test, I do see the need, but this issue was about terraform test.  And I'm not actually sure if terra grunt supports the test arguments and even how that would work.  I'll research it and open another issue for this once my research is complete.

hoangelos:
Also, should I add a disclaimer in the `terraform_test` section of the README that this may affect the commit time.  Because if the test suite is long.... I'd think that's a give, but you know what they say about assumptions.

yermulnik:
> > 1. I think running `terraform test` command makes sense only when there are TF Test files in the directory: [developer.hashicorp.com/terraform/language/tests#syntax](https://developer.hashicorp.com/terraform/language/tests#syntax) — «_Each Terraform test lives in a test file. Terraform discovers test files are based on their file extension: `.tftest.hcl` or `.tftest.json`_». This way we'd also ensure support for TF versions prior to Terraform v1.6.0
> 
> terraform test looks for`.tftest.hcl` or `.tftest.json`_». in the test directory, which by default is `tests`. I do agree that it's likely we don't want to do this per changed file, but per repo.

What I mean is that running `terraform test` makes only sense if the commit comprised changes to either TF config files or TF test files only when there are TF test files in current directory, rather than there are only TF config files changed, as there's no sense running `terraform test` command of there are no TF test files in current directory, including TF versions less than 1.6.0, which have no `test` command for Terraform CLI. 

> > 2. I'm not sure about having Terragrunt specific hook for the TF tests since `test` command doesn't belong in Terragrunt and hence gets forwarded by Terragrunt to Terraform: [terragrunt.gruntwork.io/docs/reference/cli-options/#all-terraform-built-in-commands](https://terragrunt.gruntwork.io/docs/reference/cli-options/#all-terraform-built-in-commands) — «_Terragrunt is a thin wrapper for Terraform, so except for a few of the special commands defined in these docs, Terragrunt forwards all other commands to Terraform_». Having standalone Terragrunt Tests hook makes only sense if there might be some use cases when Terrafrunt-specific command line options need to be passed to the hook (I'm not Terragrunt user, thus I'm can say for sure if there are any use cases of such kind).
> 
> So, your ignorance of Terragrunt has already created a gap in the hooks.

Please don't try and make unreasonable assumptions.

>  Wrapper is a correct, but not fully accurate.

Yes, I agree, and this is why I commented on this PR.
It looked like there's no sense to require Terragrunt CLI and to relay `test` command via Terragrunt CLI if `terraform test` suffices the task. I might be wrong though as I'm not a TG user (which doesn't imply ignorance though).

> However, terra grunt fmt hook only does hclfmt, which doesn't actually run any terraform fmt, which leaves these generated files as a gap. That's a long way to say, technically most of the terraform commands will need separate terragrunt commands too.

The TG `hclfmt` is not a TF command, whilst TG does relay `test` command directly to TF since it's TF-specific command as opposite to `hclfmt` being a TG-specific command, which doesn't belong in TF, and hence this means what you're comparing is not of the same kind with what I was talking about.

yermulnik:
> Also, should I add a disclaimer in the `terraform_test` section of the README that this may affect the commit time. Because if the test suite is long.... I'd think that's a give, but you know what they say about assumptions.

Yes, you probably should. It makes sense to let consumers know this hook it compatible with Terraform version `>=1.6.`0 only and will produce failure for all the folks with Terraform version `<1.6.0`. Along with that it makes sense to remove this hook from the default `pre-commit-terraform` to maintain backward compatibility by avoiding failure that requires human interaction to revolve the failure by adjusting default `pre-commit-terraform` config.
And yes, I'd think that's a give, but you know what they that about OSS breaking backward compatibility w/o argued reason and prior (line prior) notification 😏

yermulnik:
Please don't take my PR comments as ignorance or criticism. I do appreciate the contribution. Though what we'd want to achieve is the tool that can be used by many, which implies a need to maintain backward compatibility and reasonability of changes that are not backward compatible.

yermulnik:
Simplicity vs flexibility, you know…

hoangelos:
So are you suggesting we gate the terraform tests with a check to see if there are  files that look like test files in the test directory which defaults to not the current directory but a directory called tests.

hoangelos:
I do want to point out that terraform test doesn’t fail if it’s run in a directory that doesn’t have tests.

yermulnik:
> So are you suggesting we gate the terraform tests with a check to see if there are files that look like test files in the test directory which defaults to not the current directory but a directory called tests.

I'm suggesting to run `terraform test` command only in those directories that either have "TF config files changed AND there are TF test files in the same directory" or have "TF test files changed".
There's no "look like test files" from what I see in the documentation. There's pretty definitive explanation of which exact files are considered TF test files to be honest.

yermulnik:
> I do want to point out that terraform test doesn’t fail if it’s run in a directory that doesn’t have tests.

I was talking more of a TF version since TF `test` command was officially introduced with version `1.6.0` (and was not guaranteed to be supported prior to `1.6.0`) rather than about TF failing or not.
But... but I must admit I didn't know TF supports `test` command since version `0.15.0`.
So what I was concerned about doesn't make sense with TF versions `>=0.15.0`.
To be honest I'm not quire sure about which most lowest version of Terraform `pre-commit-terraform` supports officially. Hence it may (or may) not support earlier versions of Terraform.
@MaxymVlasov @antonbabenko What you folks think?

.pre-commit-hooks.yaml
```diff
@@ -77,6 +77,14 @@
   files: (\.hcl)$
   exclude: \.terraform\/.*$
 
+- id: terraform_test
+  name: Terraform test
+  description: Runs all terraform tests
+  entry: hooks/terraform_test.sh
+  language: script
+  files: (\.tf|\.tfvars|\.terraform\.lock\.hcl||\.terraform\.lock\.hcl||\.terraform\.lock\.json)$
```
yermulnik, line 85:
```suggestion
  files: \.(tf(vars|test\.(hcl|json))?|terraform\.lock\.hcl)$
```
Should

README.md
```diff
@@ -271,9 +272,10 @@ There are several [pre-commit](https://pre-commit.com/) hooks to keep Terraform
...
 | `terraform_tfsec`                                      | [TFSec](https://github.com/aquasecurity/tfsec) static analysis of terraform templates to spot potential security issues. [Hook notes](#terraform_tfsec)                                                                                      | `tfsec`                                                                              |
-| `terraform_validate`                                   | Validates all Terraform configuration files. [Hook notes](#terraform_validate)                                                                                                                                                               | `jq`, only for `--retry-once-with-cleanup` flag                                      |
+| `terraform_validate`        sd                           | Validates all Terraform configuration files. [Hook notes](#terraform_validate)                                                                                                                                                               | `jq`, only for `--retry-once-with-cleanup` flag                                      |
```
yermulnik, line 278: `sd`?

README.md
```diff
@@ -645,6 +647,24 @@ To replicate functionality in `terraform_docs` hook:
...
+      args:
+        - --args=-test-directory=path
+        - --arg=-filter=testfile
+        - --args=-json
+        - --arg=-verbose
+        - --arg=-var-file=filename
+        - --arg=-var=variable
+    ```
+
+2. `terraform_test` only runs per repository.
```
yermulnik, line 667: I beg your pardon once again, though is it not possible to run `terraform test` per directory? 🤔

hoangelos:
> The TG `hclfmt` is not a TF command, whilst TG does relay `test` command directly to TF since it's TF-specific command as opposite to `hclfmt` being a TG-specific command, which doesn't belong in TF, and hence this means what you're comparing is not of the same kind with what I was talking about.

My comment is wn by actual terragrunt usage problem that arose with your hooks.  I didn’t look into the hook so I assumed it did both fmt and hclfmt.  I used terragrunt_fmt assuming it would at least alert if there were failures in formatting these generated terraform files generated by terragrunt.  It didn’t.  My CICD caught it.  But there should be a terragrunt fmt thank a separate terragrunt_hclfmt.

README.md
hoangelos: > I beg your pardon once again, though is it not possible to run `terraform test` per directory? 🤔

not sure what that would mean.  Is it possible that people litter tests in directories throughout?  Yes.  I can put the per directory back if you want to support that.  Although they’ll have to use the same convention for test-directories for it to work.  That makes sense although I can imagine why someone would do that but I’m not a mono repo fan so that’s one use case.

yermulnik:
> My comment is wn by actual terragrunt usage problem that arose with your hooks. I didn’t look into the hook so I assumed it did both fmt and hclfmt. I used terragrunt_fmt assuming it would at least alert if there were failures in formatting these generated terraform files generated by terragrunt. It didn’t. My CICD caught it. But there should be a terragrunt fmt thank a separate terragrunt_hclfmt.

I'm a bit confused with your attitude to be honest 😢 It's an open source project and we're welcoming and we do appreciate any contribution. And I'm very much sorry you haven't gained as much satisfaction with `pre-commit-terraform` as you expected from it. And I hope you'll become a valuable part of this community by bringing in solution(s) for the use cases you've been hitting and that might be a matter for others.

README.md
yermulnik: What I mean is that we should deliberately and explicitly support both monorepo and multirepo consumers of `pre-commit-terraform` instead of making assumptions on how others should better Terraform their infrastructure.

hoangelos:
> > My comment is wn by actual terragrunt usage problem that arose with your hooks. I didn’t look into the hook so I assumed it did both fmt and hclfmt. I used terragrunt_fmt assuming it would at least alert if there were failures in formatting these generated terraform files generated by terragrunt. It didn’t. My CICD caught it. But there should be a terragrunt fmt thank a separate terragrunt_hclfmt.
> 
> I'm a bit confused with your attitude to be honest 😢 It's an open source project and we're welcoming and we do appreciate any contribution. And I'm very much sorry you haven't gained as much satisfaction with `pre-commit-terraform` as you expected from it. And I hope you'll become a valuable part of this community by bringing in solution(s) for the use cases you've been hitting and that might be a matter for others.

I’m sorry if you are reading in an attitude.  I’m just pointing out the philosophy of skipping some terragrunt commands contributes to some side effects that aren’t good.  Happy to file a bug and contribute to that.

README.md
hoangelos: > What I mean is that we should deliberately and explicitly support both monorepo and multirepo consumers of `pre-commit-terraform` instead of making assumptions on how others should better Terraform their infrastructure.

I understand why you’d want to do that as your project can enforce opinions.  I’m trying to balance the little time to contribute and wanted to contribute and not fork.  But in a
Big organization money repo is very unwieldy so we disallow it.  I’ll try to find some time to build in support that make sense per dir.

MaxymVlasov:
> > I do want to point out that terraform test doesn’t fail if it’s run in a directory that doesn’t have tests.
> 
> I was talking more of a TF version since TF `test` command was officially introduced with version `1.6.0` (and was not guaranteed to be supported prior to `1.6.0`) rather than about TF failing or not. But... but I must admit I didn't know TF supports `test` command since version `0.15.0`. So what I was concerned about doesn't make sense with TF versions `>=0.15.0`. To be honest I'm not quire sure about which most lowest version of Terraform `pre-commit-terraform` supports officially. Hence it may (or may not) make sense to support earlier versions of Terraform. @MaxymVlasov @antonbabenko What do you folks think?

Tests in 1.6.0 have breaking changes compared to tests that were before. So there is no reason to support `test` before 1.6.0 until someone opens a feature request because of opentofu etc.

hoangelos:
> Tests in 1.6.0 have breaking changes compared to tests that were before. So there is no reason to support `test` before 1.6.0 until someone opens a feature request because of opentofu etc.

So, to only have this run again 1.6.0 and above how would you recommend me add that.  Is there any other hook that does that?

README.md
MaxymVlasov: >But in a Big organization money repo is very unwieldy so we disallow it.

Well, I have ~320 unique components (modules not included), which are used by about ~2000 stacks. In one monorepo. We switched to monorepo, because it is much better manageable than hundreds of mini-repos.

I previously had "luck" to work with about ~50 repos which were setup on a per-account basic, and that was hell on Earth.

Probably, you'll start to see some issues when achieving thousands of unique components, mostly from a git perspective, which will need some split-up or additional tuning as it done in big companies like Google or Microsoft, but it is still better than trying to maintain thousands of repos at once.

In any case, the spirit of pre-commit hooks by definition - runs only on changed files a.k.a minimum subset to chack that all is OK.
Otherwise, you'll just waste time and resources. 
Also, `function run_hook_on_whole_repo {` is just an optimization option basically for `pre-commit run -a`, because usually you don't change every file when you run `git commit` (and invoking `pre-commit run`)

https://github.com/antonbabenko/pre-commit-terraform/blob/95fc56f84c6c0d2363c47dfec6c6e931d26e9f36/hooks/_common.sh#L201-L207

At the same time, not providing the `per_dir_hook_unique_part` function means a broken hook.

MaxymVlasov:
> > Tests in 1.6.0 have breaking changes compared to tests that were before. So there is no reason to support `test` before 1.6.0 until someone opens a feature request because of opentofu etc.
> 
> So, to only have this run again 1.6.0 and above how would you recommend me add that. Is there any other hook that does that?

Add info notice to README like:

> [!IMPORTANT]
> That hook supports only Terraform 1.6.0+. 
> Reason: there are [significant changes](https://github.com/hashicorp/terraform/releases/tag/v1.6.0) to how tests are written and executed, based on feedback from the experimental phase

```
> [!IMPORTANT]
> That hook supports only Terraform 1.6.0+.
> Reason: there are [significant changes](https://github.com/hashicorp/terraform/releases/tag/v1.6.0) to how tests are written and executed, based on feedback from the experimental phase"
```

Also, you could add a check to hook like this, but for `terraform --version -json | grep -E  -o '[0-9]+\.[0-9]+'` 
and compare it with `1.6`
https://github.com/antonbabenko/pre-commit-terraform/blob/95fc56f84c6c0d2363c47dfec6c6e931d26e9f36/hooks/infracost_breakdown.sh#L43-L51

MaxymVlasov:
@hoangelos howdy. Do you plan to continue working on this PR? Do you need any help from our side?

jmreicha:
Also interested in this one. Any updates or anything we can do to help land this?

MaxymVlasov:
@jmreicha feel free to fork @hoangelos branch to implement requested changes in this PR and send a new PR.
Thank you. https://github.com/antonbabenko/pre-commit-terraform/blob/master/.github/CONTRIBUTING.md

jmreicha:
Had a little bit of time to play around with this and got something working. I'd like to get some feedback on how folks imagine this should work before I open a PR for cases that I'm not thinking about.

jmreicha:
So I have the second use case accounted for currently. I suppose it makes sense to account for both, I’m imagining some extra flag to toggle the second case on/off and default the behavior to first case. Does that sound reasonable?

MaxymVlasov:
> I’m imagining some extra flag to toggle the second case on/off and default the behavior to first case. Does that sound reasonable?

Actually, that already done for you

You just need to create `per_dir_hook_unique_part` for 1 and `run_hook_on_whole_repo` for 2

https://github.com/antonbabenko/pre-commit-terraform/blob/e0322737dccad136b751416f9e45d9980fe18435/hooks/_common.sh#L289-L295

https://github.com/antonbabenko/pre-commit-terraform/blob/e0322737dccad136b751416f9e45d9980fe18435/hooks/_common.sh#L389

wyardley:
Presumably the ideal case is to run `terraform test` once for each directory containing affected files / dirs? There are a couple layouts possible, either having `xyz.tftest.hcl` (where `xyz` might or might not line up directly with the filename(s) changed), or a subdirectory, configurable, but defaulting to `tests/`.

Presumably tests should also run if the tests are changed, but figuring out what directory to run from if only tests are changed might be challenging? This is a weird corner case... for example, you're using the subdirectory approach, and have a test-only change (some external expectation changed, or you're adding a new test with no code changes; the changed file is in the subdirectory, but the place you'd want to run it would be a directory up. So, for example, with this structure:
```
├── README.md
├── examples
│   └── examples.tf
├── main.tf
├── outputs.tf
├── tests
│   └── main.tftest.hcl
├── variables.tf
└── versions.tf
```
where `tests` only is modified, you'd actually want to run `terraform test` one directory level up, and if both tests and the main terraform code is modified, you'd only want to run it in the base directory of the module.

Long term, hopefully they'll add some kind of way to label unit or plan-based tests / integration or acceptance tests separately so that you could also choose to only run a subset.

MaxymVlasov:
```
main.tf
foo.tftest.hcl
```
Easy one, can be added as is, without any problems


For
```
main.tf
test/foo.tftest.hcl
```
and
```
main.tf
anyfolder/bar.tftest.hcl
```

Current `common::per_dir_hook` could be problematic.
It's not a big deal to read env var (or where non-default test folder path set) and set it by default to `test/` if not founded.


So basically, we need to deduplicate `<unique_path>/` and `<unique_path>/<test_folder_name>`
and always run `terraform test` on  `<unique_path>/` (if I rightly understand how it works)

To achieve that we can drop `<test_folder_name>` here
https://github.com/antonbabenko/pre-commit-terraform/blob/e0322737dccad136b751416f9e45d9980fe18435/hooks/_common.sh#L297-L306

<details><summary>In case if there no ENV VAR to specify `test/` directory or there to many options to set it...</summary>

This will require to move section above below 
https://github.com/antonbabenko/pre-commit-terraform/blob/e0322737dccad136b751416f9e45d9980fe18435/hooks/_common.sh#L308-L339

and add to last one `--hook-config` for manually set test dir name to hook

... and after adding hook-specific functional to global scope, we will need to try to not cry.

</details>

MaxymVlasov:
Note, that all what I wrote above will work only in case if tests ALWAYS located or in same folder where TF code located or in subfolder (no matter how deep).

But if there exist any possibility to specify path to test in parent folder, or in same level-folder like:

```
module/main.tf
tests/foo.tftest.hcl
```
```
module/main.tf
bar.tftest.hcl
```

then proposal above will not cover it.

jmreicha:
I have discovered that there are some quirks with how terraform test behaves. For example, the tests directory must be located within the module that is initialized, so the first example above would be invalid.

@wyardley there is a filter filter option to limit execution to a subset of the tests. The convention I have adopted is to split tests into unit (plan) and integration (apply) which influenced my original implementation.

I’ll take the feedback and hopefully should be able to come up with something workable. Thanks for the input and suggestions.

PR #620, opened 2024-02-09
feat: Add parallelism to major chunk of hooks. Check `Parallelism` section in README

MaxymVlasov:

- [ ] This PR fixes a bug.

### Reasoning 

We have a GH workflow that runs lockflies updates every week (implementation and reasoning [here](https://grem1.in/post/terraform-lockfiles-maxymvlasov/)). 
It usually takes from 2h 30min to  3h 15min. That was fine for us, till we found that our GH runners, based on AWS EC2s, started silently failing after 30min "without recent logs", and that was fixed by crutch which sends a dummy log every 10min.

However, during the debugging, I spent some time describing why hooks were not utilizing all the provided resources. 
And that means a waste of time and money, not only for that corner case but for every huge commit, which can cause opting out by `git commit -n` of using hooks locally for changes that affect many directories.

* Add per-hook `--parallelism-limit` setting to `--hook-config`. Defaults to `number of logical CPUs - 1`
* <s>Main implementation uses `flock`, which not pre-installed in macOS and some ancient BSD/Linux systems.</s>
  * <s>To maintain backward compatibility, for cases where no `flock` - fallback to "bash-native locking mechanism" was added.</s>
* As quick tests show, ~5% of stacks face race condition problem, no matter if any locking mechanism exists or dirs try to init in parallel. I suppose the lock failed as it uses disk when hooks run in memory, so the creation of the lock can take some time as there bunch of caches between Mem and Disk. These milliseconds are enough to allow running a few `t init` in parallel.
* Final implementation uses a retry mechanism for cases when race condition failed to `t init` directory. 

In quick tests, I can say that on big changes:
* <s>With `flock`, there 300-500% speed increase for most of the hooks.</s>
* <s>Without `flock`, there 250-450% speed increase for most of the hooks.</s>
* Up to 2000% speed increase for `terraform_validate`, and up to 500% - for other affected hooks. 
* When `--parallelism-limit=1` I observed an insignificant increase in time (about 5-10%) compared to v1.86.0 which has no parallelism at all. This may be the cost of maintaining parallelism or the result of external factors since the tests were not conducted in a vacuum.

For small changes, improvements are less significant.

### How can we test changes

<details><summary>There is no easy way to test it, but that could be used as a testing template</summary>

```yaml
repos:
  rev: f8d26f85b054e2842ef1b0c5bcef82d8c8cf1832
  hooks:
    # - id: infracost_breakdown
    #   args:
    #     # - --args=--config-file=test/infracost.yml
    #     - --args=--path=./environment/qa
    #     - --args=--terraform-plan-flags="-var-file=../../test.tfvars"t
    #     - --args=--terraform-plan-flags="-var-file=../../test2.tfvars"
    #     - --hook-config='.totalHourlyCost >= 0.1'
    #     - --hook-config=".totalHourlyCost|tonumber > 1"
    #     - -h ".projects[].diff.totalMonthlyCost|tonumber!=10000"
    #     - -h [.projects[].diff.totalMonthlyCost | select (.!=null) | tonumber] | add > 1000
    #   verbose: true # Always show costs

    # - id: terraform_checkov
    #   args:
    #   #   - --args=--quiet
    #     - --args=--skip-check CKV_SECRET_14
    #     # - --env-vars=ANSI_COLORS_DISABLED=true
    #     # - --hook-config=--parallelism-limit=1


    # - id: terraform_docs
    #   args:
    #   # - --hook-config=--path-to-file=README.md        # Valid UNIX path. I.e. ../TFDOC.md or docs/README.md etc.
    #   # - --hook-config=--add-to-existing-file=true     # Boolean. true or false
    #   - --hook-config=--create-file-if-not-exist=true # Boolean. true or false
    #   # - --args=--config=__GIT_WORKING_DIR__/.terraform-docs.yml
    #   - --args=--config=.terraform-docs.yml
    #   # - --args=--lockfile=false
    #   # - --args=--max-depth=2


    # - id: terraform_fmt
    #   exclude: environment\/.*$


    # - id: terraform_providers_lock
    #   args:
    #   # - --args=-platform=windows_amd64
    #   # - --args=-platform=darwin_amd64
    #   - --args=-platform=linux_amd64

    #   - --hook-config=--mode=always-regenerate-lockfile
    #   # - --hook-config=--parallelism-limit=CPU_NUM*2
    #   verbose: true

    # - id: terraform_tflint
    #   args:
    #     - --args=--config=__GIT_WORKING_DIR__/.config/tflint.hcl
    #     # # - '--args=--config=__GIT_WORKING_DIR__/${CONFIG_FILE}' # export CONFIG_FILE=.tflint.hcl
    #     # - --args=--config=__GIT_WORKING_DIR__/${CONFIG_NAME}.${CONFIG_EXT} # export CONFIG_NAME=.tflint; export CONFIG_EXT=hcl
    #     - --args=--module

    # - id: terraform_tfsec
    #   args:
    #       - --args=--config-file=.tfsec.json
    #       - --args=--force-all-dirs
    #       - --args=--exclude-downloaded-modules
    #       - --args=--concise-output
    #       - --args=--config-file=__GIT_WORKING_DIR__/.tfsec.json

    # - id: terraform_trivy
    #   args:
    #     - --hook-config=--parallelism-limit=CPU_NUM*4


    # - id: terraform_validate
    #   args:
    #     - --tf-init-args=-upgrade
    #     - --tf-init-args=-get=true
    #     - --env-vars=AWS_DEFAULT_REGION="us-west-2"
    #     - --env-vars=AWS_ACCESS_KEY_ID="anaccesskey"
    #     - --env-vars=AWS_SECRET_ACCESS_KEY="asecretkey"
    #     - --hook-config=--retry-once-with-cleanup=true     # Boolean. true or false
    # #    - --hook-config=--parallelism-limit=CPU_NUM*4
    #   verbose: true



    # - id: terraform_validate
    #   args:
    #     - --hook-config=--retry-once-with-cleanup=true


    # - id: terragrunt_fmt


    # - id: terragrunt_validate


    # - id: terrascan
    #   args:
    #     - --args=--non-recursive # avoids scan errors on subdirectories without Terraform config files
    #     - --args=--policy-type aws
    #     # - --hook-config=--parallelism-limit=CPU_NUM*2


    - id: tfupdate
      name: Autoupdate AWS provider versions
      args:

        - --args=provider aws
        - --args=--version '~> 4.2.0'
        - --hook-config=--parallelism-limit=1


    # - id: tfupdate
    #   name: Autoupdate Helm provider versions
    #   args:
    #     - --args=provider helm
    #     - --args=-v 2.5.0
    #     - --args=--ignore-path sdfs/sfdfs

    # - id: tfupdate
    #   name: Update Google provider to specified version
    #   args:
    #     - --args=provider google
    #     - --args=--ignore-path sdfs/sfdfs
    #     - --args=--version ">= 4.18.0, < 5"

    # - id: tfupdate
    #   name: Update Terraform provider to specified version
    #   args:
    #     - --args=terraform
    #     - --args=--version "> 1.1.8"
    #     - --env-vars=AWS_DEFAULT_REGION="us-west-2"
    #     - --env-vars=AWS_ACCESS_KEY_ID="anaccesskey"
    #     - --env-vars=AWS_SECRET_ACCESS_KEY="asecretkey"
```

</details>


hooks/terraform_providers_lock.sh
```diff
@@ -15,6 +15,8 @@ function main {
   common::parse_and_export_env_vars
   # JFYI: suppress color for `terraform providers lock` is N/A`
 
+  # Cleanup in case lockfile was not cleaned in previous run of common::terraform_init function
+  rm -rf "$PARALLELISM_FALLBACK_LOCK_DIR"
```
yermulnik, line 19: I'm wondering why in one case `rmdir` is used while in another — `rm -rf` — is there a reason for this discrepancy in approaches? 🤔

hooks/_common.sh
```diff
@@ -307,15 +345,19 @@ function common::colorify {
...
+#   TF_PLUGIN_CACHE_DIR (string) user defined env var to directory
+#     which can't be R/W concurrently
 # Outputs:
 #   If failed - print out terraform init output
 #######################################################################
 # TODO: v2.0: Move it inside terraform_validate.sh
 function common::terraform_init {
   local -r command_name=$1
   local -r dir_path=$2
+  local -r parallelism_disabled=$3
```
yermulnik, line 360: For the sake of consistency across hooks, it's better to wrap assignments into double quotes (`local -r var_name="$var"`)

yermulnik (commented):
Great, invaluable work! ❤️

.github/ISSUE_TEMPLATE/bug_report_local_install.md
```diff
@@ -89,6 +89,7 @@ tflint --version                          2>/dev/null || echo "tflint SKIPPED"
 echo -n "tfsec " && tfsec --version       2>/dev/null || echo "tfsec SKIPPED"
 echo -n "tfupdate " && tfupdate --version 2>/dev/null || echo "tfupdate SKIPPED"
 echo -n "hcledit " && hcledit version     2>/dev/null || echo "hcledit SKIPPED"
+echo -n "flock " && flock --versiob       2>/dev/null || echo "flock SKIPPED"
```
yermulnik, line 92:
```suggestion
echo -n "flock " && flock --version 2>/dev/null || echo "SKIPPED"
```

Dockerfile
```diff
@@ -173,6 +173,7 @@ RUN . /.env && \
 RUN . /.env && \
     F=tools_versions_info && \
     pre-commit --version >> $F && \
+    echo "flock $(flock 2>&1 | head -n 1)" >> $F && \
```
yermulnik, line 176:
```suggestion
flock --version 2>&1 >> $F && \
```

MaxymVlasov: Nope

```
➜ docker run -ti --entrypoint=bash ghcr.io/antonbabenko/pre-commit-terraform
de8f525d1c67:/# flock --version
flock: unrecognized option: version
BusyBox v1.35.0 (2022-11-19 10:13:10 UTC) multi-call binary.

Usage: flock [-sxun] FD | { FILE [-c] PROG ARGS }

[Un]lock file descriptor, or lock FILE, run PROG

        -s      Shared lock
        -x      Exclusive lock (default)
        -u      Unlock FD
        -n      Fail rather than wait
de8f525d1c67:/# echo "flock $(flock 2>&1 | head -n 1)"
flock BusyBox v1.35.0 (2022-11-19 10:13:10 UTC) multi-call binary.
de8f525d1c67:/# 
```

yermulnik: Ah, my bad, missed this is in docker.

README.md
```diff
@@ -341,6 +343,60 @@ To disable color output for all hooks, set `PRE_COMMIT_COLOR=never` var. Eg:
...
+
+> [!TIP]
+> <details><summary>Info useful for parallelism fine-tunning</summary>
+>
+> <br>
+> Tests below was run on repo with 45 terraform dirs on laptop with 16 CPU, SSD and 1Gbit/s network. Laptop was slightly used in the process.
+>
+> Observed results may vary greatly depending on your repo structure, machine characteristics and their usage.
+>
+> If during fine-tuning you'll find that your results are very different from provided below and you thonk that this data could help someone else - feel free to send PR.
```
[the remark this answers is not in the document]
MaxymVlasov: `thonk` is a new `bonk` :D

README.md
```diff
@@ -341,6 +343,60 @@ To disable color output for all hooks, set `PRE_COMMIT_COLOR=never` var. Eg:
...
+>
+>
+> | Hook                                     | Most used resource                 | Optimization runs comparison / Notes                            |
```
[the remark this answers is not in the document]
MaxymVlasov: I mean something like "Comparison of optimization runs"

hooks/_common.sh
```diff
@@ -326,8 +368,36 @@ function common::terraform_init {
   fi
 
   if [ ! -d .terraform/modules ] || [ ! -d .terraform/providers ]; then
-    init_output=$(terraform init -backend=false "${TF_INIT_ARGS[@]}" 2>&1)
-    exit_code=$?
+    # Plugin cache dir can't be write concurrently or read during writing
+    # https://github.com/hashicorp/terraform/issues/31964
+    if [ -z "$TF_PLUGIN_CACHE_DIR" ] || $parallelism_disabled; then
```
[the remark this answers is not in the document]
yermulnik: Just had a thought that I don't remember whether `[[` is supported in Bash v3 =(
So maybe it is not a good idea to switch from `[` to `[[` 😢 Along with that everything inside single brackets has to be wrapped in double quotes (I seem to think I met this in this PR, where some of the expressions had LHS or RHS of the expression being not wrapped into double quotes).

ps: who the hell use Bash v3 nowadays and why they don't `brew install bash`...? 🤪

hooks/terraform_providers_lock.sh
MaxymVlasov: Good catch. Switched to `rm -rf` as more robust solution (in case something will wrote anything inside dir)

README.md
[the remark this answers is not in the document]
yermulnik: Maybe "Comparison of optimization results" then?

hooks/terraform_providers_lock.sh
yermulnik: Fingers crossed `$PARALLELISM_FALLBACK_LOCK_DIR` won't resolve to `/` on systems that have not safety mechanism to prevent removal for root dir 🤞🏻

hooks/_common.sh
```diff
@@ -233,38 +248,61 @@ function common::per_dir_hook {
...
-    local exit_code=$?
-    if [ $exit_code -ne 0 ]; then
-      final_exit_code=$exit_code
-    fi
-
-    if [[ $change_dir_in_unique_part == false ]]; then
-      popd > /dev/null
+        if [ $exit_code -ne 0 ]; then
+          final_exit_code=$exit_code
+        fi
```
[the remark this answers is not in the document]
MaxymVlasov: I prefer the explicit variant

hooks/_common.sh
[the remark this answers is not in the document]
MaxymVlasov: It supported in bash 3.2.57, but not in sh/POSIX AFAIK

hooks/_common.sh
```diff
@@ -326,8 +368,36 @@ function common::terraform_init {
...
+      if command -v flock &> /dev/null; then
+        init_output=$(
+          flock --exclusive "$TF_PLUGIN_CACHE_DIR" \
+            terraform init -backend=false "${TF_INIT_ARGS[@]}" 2>&1
+        )
+        exit_code=$?
+      # Fall back "simple-lock" mechanizm if `flock` is not available
+      else
+
+        while true; do
```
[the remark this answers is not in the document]
MaxymVlasov: Can't mind such situation

hooks/_common.sh
```diff
@@ -233,38 +248,61 @@ function common::per_dir_hook {
...
   # preserve errexit status
-  shopt -qo errexit && ERREXIT_IS_SET=true
-  # allow hook to continue if exit_code is greater than 0
-  set +e
+  shopt -qo errexit
+
   local final_exit_code=0
+  local pids=()
+
+  mapfile -t dir_paths_unique < <(echo "${dir_paths[*]}" | tr ' ' '\n' | sort -u)
```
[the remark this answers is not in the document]
MaxymVlasov: Because I use IFS only for built-in functions, global IFS not changed
(yes, read code from question, not answers :))
https://unix.stackexchange.com/questions/92187/setting-ifs-for-a-single-statement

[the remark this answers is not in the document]
MaxymVlasov: But, as long as shellcheck OK with `*` -> `@` change, I'll do it

hooks/_common.sh
```diff
@@ -233,38 +248,61 @@ function common::per_dir_hook {
...
-    dir_path="${dir_path//__REPLACED__SPACE__/ }"
+      per_dir_hook_unique_part "$dir_path" "$change_dir_in_unique_part" "$parallelism_disabled" "${args[@]}"
+    } &
+    pid=$!
+    pids+=("$pid")
 
-    if [[ $change_dir_in_unique_part == false ]]; then
-      pushd "$dir_path" > /dev/null || continue
-    fi
+    if [ $parallelism_disabled ] ||
```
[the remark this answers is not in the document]
MaxymVlasov: I think that can be done in simple way

```suggestion
    if $parallelism_disabled ||
```

hooks/terraform_providers_lock.sh
MaxymVlasov: eeeeh. Okay, rmdir then

hooks/_common.sh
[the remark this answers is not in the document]
robinbowes: This will try to execute the contents of `$parallelism_disabled` as a command.

@yermulnik's answer is the accepted way to do this:

https://stackoverflow.com/questions/2953646/how-can-i-declare-and-use-boolean-variables-in-a-shell-script

(I prefer to use `[[ ]]`, but that's mostly a personal consistency/style thing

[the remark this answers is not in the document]
MaxymVlasov: And `$parallelism_disabled` can be only `true` or `false`, as it is set only in code.

And as I see from the link which points to https://stackoverflow.com/questions/2953646/how-can-i-declare-and-use-boolean-variables-in-a-shell-script/21210966#21210966

It's OK till it can be only boolean, right?

Or still better to add extra safety for future "just in case"?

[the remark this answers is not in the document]
robinbowes: It's always better to be explicit in my opinion.

You want to test if the variable contains a value you have set: "true".

So test for the variable being equal to "true".

hooks/_common.sh
```diff
@@ -237,7 +237,7 @@ function common::per_dir_hook {
     # It gets split by `;` into array, which we're parsing here ('--foo=bar' ' --baz=4')
     # Next line removes leading spaces, to support >1 `--hook-config` args
     # shellcheck disable=SC2001 # Rule exception
-    key=$(echo "${config[0]}" | sed 's/^[[:space:]]*//')
+    key=$(sed 's/^[[:space:]]*//' <<< "${config[0]}")
```
robinbowes, line 240: You can do this with bash parameter substitution:

```
key="${config[0]## }"
```

See here for details: https://tldp.org/LDP/abs/html/parameter-substitution.html

Search for `${var##Pattern}`

.github/ISSUE_TEMPLATE/bug_report_local_install.md
```diff
@@ -91,6 +91,7 @@ echo -n "tfsec " && tfsec --version       2>/dev/null || echo "tfsec SKIPPED"
 echo -n "trivy " && trivy --version       2>/dev/null || echo "tfsec SKIPPED"
 echo -n "tfupdate " && tfupdate --version 2>/dev/null || echo "tfupdate SKIPPED"
 echo -n "hcledit " && hcledit version     2>/dev/null || echo "hcledit SKIPPED"
+echo -n "flock " && flock --version       2>/dev/null || echo "flock SKIPPED"
```
robinbowes, line 94: This construct does not work as you expect it to. If flock is not installed, this is the output:

```
echo -n "flock " && flock --version       2>/dev/null || echo "flock SKIPPED"
flock flock SKIPPED
```

You could remove the 2nd "flock", ie:

```
echo -n "flock " && flock --version       2>/dev/null || echo "SKIPPED"
```

The same applies to many of the other commands too.

hooks/_common.sh
```diff
@@ -233,38 +248,61 @@ function common::per_dir_hook {
...
-    fi
-
-    if [[ $change_dir_in_unique_part == false ]]; then
-      popd > /dev/null
+        if [ $exit_code -ne 0 ]; then
+          final_exit_code=$exit_code
+        fi
+      done
+      # Reset pids for next iteration
+      pids=()
```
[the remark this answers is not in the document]
yermulnik: The `{}` is the grouping operator. The subshell is `()`.
Example:
```shell
> cat /tmp/yz/zzz.sh
#!/usr/bin/env bash

pwd
{ pushd /tmp/ >/dev/null; pwd; }
pwd

> /tmp/yz/zzz.sh
/home/yz
/tmp
/tmp
```

hooks/_common.sh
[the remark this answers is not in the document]
yermulnik: > but not in sh/POSIX AFAIK

Yep, this is not supported in POSIX shell. It's Bash-specific I guess. Though since we're running this code with Bash we should only care of whether it is supported with Bash v3. And since it is supported there, we're safe 👍🏻

hooks/_common.sh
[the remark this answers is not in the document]
yermulnik: E.g. somehow happened that pre-commit was run twice from different terminals and the first instance successfully created `$PARALLELISM_FALLBACK_LOCK_DIR` dir, but subsequent `terraform init` is e.g. killed by OOM killer and hence `$PARALLELISM_FALLBACK_LOCK_DIR` dir is not cleaned up — then the second instance is going to loop forever... until the dev comes back from the vacation =)
I might be overlooking something, though it's best to put safety measure here "just in case" something goes wrong. IMHO.

hooks/_common.sh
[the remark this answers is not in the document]
yermulnik: > Because I use IFS only for built-in functions, global IFS not changed

It's not about global IFS. It's more about an inadvertent stuff, like e.g. for any reason you set `IFS` to a non-standard value somewhere before this block of code and forget to set it back to a standard value (or unset it to reset to default value). In the command above the `"${dir_paths[*]}"` expands to single word of array elements separated by the first char of `IFS` while `tr` tool replaces **spaces** — and hence `tr` cannot do what it is expected to. In the same time the `"${dir_paths[@]}"` expands to words separated by space (not sure whether there are any situations when it is not a space though).

> But, as long as shellcheck OK with `*` -> `@` change, I'll do it

👍🏻

hooks/_common.sh
yermulnik: Great catch @robinbowes. I was looking at the original expression, was feeling there's something not right with it, but my attention caught the extra `echo` while the better way is param substitution that you suggested 👍🏻

.github/ISSUE_TEMPLATE/bug_report_local_install.md
yermulnik: Hmm, I was sure I commented the same, though... I missed to =) Thanks @robinbowes for catching this.

README.md
```diff
@@ -265,24 +267,24 @@ docker run --rm --entrypoint cat ghcr.io/antonbabenko/pre-commit-terraform:$TAG
...
+| `terraform_docs_without_`<br>`aggregate_type_defaults` | Inserts input and output documentation into `README.md` without aggregate type defaults. Hook notes same as for [terraform_docs](#terraform_docs)                                                                                            | `terraform-docs`                                                                            |
+| `terraform_fmt`                                        | Reformat all Terraform configuration files to a canonical format. [Hook notes](#terraform_fmt)                                                                                                                                               | -                                                                                           |
+| `terraform_providers_lock`                             | Updates provider signatures in [dependency lock files](https://www.terraform.io/docs/cli/commands/providers/lock.html). [Hook notes](#terraform_providers_lock)                                                                              | `flock` recommended (have fallback to bash)                                                 |
```
yermulnik, line 278:
```suggestion
| `terraform_providers_lock`                             | Updates provider signatures in [dependency lock files](https://www.terraform.io/docs/cli/commands/providers/lock.html). [Hook notes](#terraform_providers_lock)                                                                              | `flock` recommended (fallbacks to pure shell implementation if `flock` isn't found)                                                 |
```

README.md
```diff
@@ -265,24 +267,24 @@ docker run --rm --entrypoint cat ghcr.io/antonbabenko/pre-commit-terraform:$TAG
...
+| `terraform_tfsec`                                      | [TFSec](https://github.com/aquasecurity/tfsec) static analysis of terraform templates to spot potential security issues. **DEPRECATED**, use `terraform_trivy`. [Hook notes](#terraform_tfsec-deprecated)                                    | `tfsec`                                                                                     |
+| `terraform_trivy`                                      | [Trivy](https://github.com/aquasecurity/trivy) static analysis of terraform templates to spot potential security issues. [Hook notes](#terraform_trivy)                                                                                      | `trivy`                                                                                     |
+| `terraform_validate`                                   | Validates all Terraform configuration files. [Hook notes](#terraform_validate)                                                                                                                                                               | `jq` only for `--retry-once-with-cleanup` flag; `flock` recommended (have fallback to bash) |
```
yermulnik, line 282:
```suggestion
| `terraform_validate`                                   | Validates all Terraform configuration files. [Hook notes](#terraform_validate)                                                                                                                                                               | `jq` only for `--retry-once-with-cleanup` flag; `flock` recommended (fallbacks to pure shell implementation if `flock` isn't found) |
```

hooks/_common.sh
```diff
@@ -219,10 +222,22 @@ function common::per_dir_hook {
...
   IFS=";" read -r -a configs <<< "${HOOK_CONFIG[*]}"
   for c in "${configs[@]}"; do
     IFS="=" read -r -a config <<< "$c"
-    key=${config[0]}
+
+    # $hook_config receives string like '--foo=bar; --baz=4;' etc.
+    # It gets split by `;` into array, which we're parsing here ('--foo=bar' ' --baz=4')
+    # Next line removes leading spaces, to support >1 `--hook-config` args
+    # shellcheck disable=SC2001 # Rule exception
+    key=$(sed 's/^[[:space:]]*//' <<< "${config[0]}")
```
yermulnik, line 240: For Max's convenience:
```suggestion
    key="${config[0]## }"
```

hooks/_common.sh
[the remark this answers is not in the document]
yermulnik: The commit suggestion is still valid as the `if [ -z "$TF_PLUGIN_CACHE_DIR" ] || $parallelism_disabled; then` is incorrect:
```shell
# var is not set
> unset parallelism_disabled
> if $parallelism_disabled; then echo DISABLED; fi
DISABLED
```
Also the quirk of setting var to `true` or `false` and testing its value w/o using `test` or `[[` commands is effectively a... well a quirk since it is based on a variable expansion and the fact that value expands to respective executables (either shell builtins or binaries under `/bin/`). This is pretty much unreliable as once you wrap value in quotes inside the conditional expression, it won't work as "boolean" any more, but will become an ordinary string w/o no meaning of `true of `false`. Using this quirk is subject to unexpectedness and prone to errors.

hooks/_common.sh
```diff
@@ -326,8 +364,36 @@ function common::terraform_init {
...
+          flock --exclusive "$TF_PLUGIN_CACHE_DIR" \
+            terraform init -backend=false "${TF_INIT_ARGS[@]}" 2>&1
+        )
+        exit_code=$?
+      # Fallback to a "simple-lock" mechanism if `flock` is not available
+      else
+
+        while true; do
+          if mkdir "$PARALLELISM_FALLBACK_LOCK_DIR" 2> /dev/null; then
+            init_output=$(terraform init -backend=false "${TF_INIT_ARGS[@]}" 2>&1)
```
yermulnik, line 385: Since the `terraform init -backend=false "${TF_INIT_ARGS[@]}" 2>&1` expression is repeated at least three times in a row in this `if/else` block of code, it makes sense to define it as var before this block so that it conforms with DRY principle and simplifies management of this expression.

hooks/terraform_providers_lock.sh
```diff
@@ -15,6 +15,8 @@ function main {
   common::parse_and_export_env_vars
   # JFYI: suppress color for `terraform providers lock` is N/A`
 
+  # Cleanup in case lockfile was not cleaned in previous run of common::terraform_init function
+  rmdir "$PARALLELISM_FALLBACK_LOCK_DIR"
```
yermulnik, line 19: Maybe check if it exists to avoid `No such file or directory` error?
```suggestion
  [[ -d "$PARALLELISM_FALLBACK_LOCK_DIR" ]] && rmdir "$PARALLELISM_FALLBACK_LOCK_DIR"
```

hooks/terraform_validate.sh
```diff
@@ -20,6 +20,9 @@ function main {
   if [ "$PRE_COMMIT_COLOR" = "never" ]; then
     ARGS+=("-no-color")
   fi
+
+  # Cleanup in case lockfile was not cleaned in previous run of common::terraform_init function
+  rmdir "$PARALLELISM_FALLBACK_LOCK_DIR"
```
yermulnik, line 25:
```suggestion
  [[ -d "$PARALLELISM_FALLBACK_LOCK_DIR" ]] && rmdir "$PARALLELISM_FALLBACK_LOCK_DIR"
```

hooks/_common.sh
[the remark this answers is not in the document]
MaxymVlasov: & -  is a subshell by definition, as you move the command to a standalone process

hooks/_common.sh
MaxymVlasov: That's not work. 
https://github.com/antonbabenko/pre-commit-terraform/pull/620/commits/a1cac640cb848326e171ed7833493292c12d7bef

```
Too many command line arguments. Did you mean to use -chdir?
```

Reverted

hooks/_common.sh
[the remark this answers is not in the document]
yermulnik: Oops, I tend to sometimes be inattentive — the `&` slipped out of my attention 🤦🏻 Apologies for the noise.

hooks/_common.sh
yermulnik: Executing variables is a bit tricky and it's where `eval` comes in handy: try `init_output=$(eval $init_command)`

MaxymVlasov: No thanks, I prefer to avoid eval as not safe construction. 
And not only me
https://google.github.io/styleguide/shellguide.html#s6.6-eval

yermulnik: In this particular case I think the only thing we need to care about is command injection by means of `TF_INIT_ARGS` array, which apparently is remediated like the below I guess (by quoting `TF_INIT_ARGS` elements):
```
init_command="terraform init -backend=false $(printf "'%s' " ${TF_INIT_ARGS[@]}) 2>&1"
init_output=$(eval $init_command)
```
Ref: https://earthly.dev/blog/safely-using-bash-eval/

ps: up to you to leave it as is (multiple same tf init defs)

hooks/_common.sh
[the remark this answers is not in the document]
MaxymVlasov: Okay, let's imagine that  subshell `init_output=$(terraform init  ..)` in `L385` will be killed.
So it will exit with bad exit code and the dir will be removed in `L387`.

If the main pre-commit run which run cycle will be killed - then everything should stop, and if not and become a zombie - that's not what we can handle here..


Or I miss something?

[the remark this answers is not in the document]
MaxymVlasov: Aha, I found what I miss

pre-commit main -> many parallel runs in background -> each run have init subshell

If parallel run will be killed after it will create a dir, but before remove it, we get infinite loop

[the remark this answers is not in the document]
yermulnik: Yep, effectively such a safety measure is to try and secure us from any kind of unexpectedness whether we can or we cannot predict it.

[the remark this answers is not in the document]
MaxymVlasov: Done

hooks/_common.sh
```diff
@@ -326,8 +360,56 @@ function common::terraform_init {
...
+          flock --exclusive "$TF_PLUGIN_CACHE_DIR" \
+            terraform init -backend=false "${TF_INIT_ARGS[@]}" 2>&1
+        )
+        exit_code=$?
+      # Fallback to a "simple-lock" mechanism if `flock` is not available
+      else
+
+        # Get lockfile name unique to the current pre-commit process
+        local -r lockfile="/tmp/TF_PLUGIN_CACHE_DIR_lock_$(grep -a -P $PPID | cut -d' ' -f1)"
+        local timeout=600 # 10min in seconds
```
MaxymVlasov, line 381: Why 10min

I checked that the cold-init of one of the biggest tf components (basically, a module of modules) took 58s and 240MB.

In case there are any potato internet receivers over exist which has only 10Mb/s, that would mean they need 

240MB×8Mb÷10Mb÷60s = 3.2min just for download.

A I'm not confident enough about 5min

.github/ISSUE_TEMPLATE/bug_report_local_install.md
```diff
@@ -81,16 +81,17 @@ pre-commit --version                      2>/dev/null || echo "pre-commit SKIPPE
...
 tflint --version                          2>/dev/null || echo "tflint SKIPPED"
-echo -n "tfsec " && tfsec --version       2>/dev/null || echo "tfsec SKIPPED"
-echo -n "trivy " && trivy --version       2>/dev/null || echo "tfsec SKIPPED"
-echo -n "tfupdate " && tfupdate --version 2>/dev/null || echo "tfupdate SKIPPED"
-echo -n "hcledit " && hcledit version     2>/dev/null || echo "hcledit SKIPPED"
+echo -n "tfsec " && tfsec --version       2>/dev/null || echo "SKIPPED"
+echo -n "trivy " && trivy --version       2>/dev/null || echo "SKIPPED"
+echo -n "tfupdate " && tfupdate --version 2>/dev/null || echo "SKIPPED"
+echo -n "hcledit " && hcledit version     2>/dev/null || echo "SKIPPED"
+echo -n "flock " && flock --version       2>/dev/null || echo "SKIPPED"
```
yermulnik, line 94: I've got a `flock` from `util-linux` bundle and it reports its name when querying its version so I'm having messed output:
```shell
> echo -n "flock " && flock --version       2>/dev/null || echo "SKIPPED"
flock flock from util-linux 2.37.2
```
Does `flock` installed by other means not provide leading `flock` in the output?

hooks/_common.sh
```diff
@@ -219,10 +219,21 @@ function common::per_dir_hook {
 
   # Lookup hook-config for modifiers that impact common behavior
   local change_dir_in_unique_part=false
+  # Limit the number of parallel processes to the number of CPU cores -1
+  # `nproc` - linux, `sysctl -n hw.ncpu` - macOS, `echo 1` - fallback
+  local CPU
+  CPU=$(nproc || sysctl -n hw.ncpu || echo 1)
```
yermulnik, line 225: I seem to recall there was something why declaration need to be split off from definition, but I can't remember what exactly, and hence:
```suggestion
  local CPU=$(nproc || sysctl -n hw.ncpu || echo 1)
```
ps: works for me locally =)

hooks/_common.sh
```diff
@@ -233,38 +244,57 @@ function common::per_dir_hook {
...
+      per_dir_hook_unique_part "$dir_path" "$change_dir_in_unique_part" "$parallelism_disabled" "${args[@]}"
+    } &
+    pids+=("$!")
 
-    if [[ $change_dir_in_unique_part == false ]]; then
-      pushd "$dir_path" > /dev/null || continue
-    fi
+    if [[ $parallelism_disabled == true ]] ||
+      [ "$i" -ne 0 ] && [ $((i % parallelism_limit)) -eq 0 ] || # don't stop on first iteration when parallelism_limit>1
+      [ "$i" -eq $last_index ]; then
```
yermulnik, line 281: I'm not entirely sure about the details of conditional, but I feel like conditional expressions on line 280 (`[ "$i" -ne 0 ] && [ $((i % parallelism_limit)) -eq 0 ]`) should be grouped (and if yes, then a change to third conditional (`[` -> `[[`) for consistency with previous two; also removing double quotes as they are not needed inside `[[ ... ]]`) 🤔 
```suggestion
    if [[ $parallelism_disabled == true ]] ||
      [[ $i -ne 0 && $((i % parallelism_limit)) -eq 0 ]] || # don't stop on first iteration when parallelism_limit>1
      [[ $i -eq $last_index ]]; then
```

hooks/_common.sh
```diff
@@ -326,8 +360,56 @@ function common::terraform_init {
...
+        init_output=$(
+          flock --exclusive "$TF_PLUGIN_CACHE_DIR" \
+            terraform init -backend=false "${TF_INIT_ARGS[@]}" 2>&1
+        )
+        exit_code=$?
+      # Fallback to a "simple-lock" mechanism if `flock` is not available
+      else
+
+        # Get lockfile name unique to the current pre-commit process
+        local -r lockfile="/tmp/TF_PLUGIN_CACHE_DIR_lock_$(grep -a -P $PPID | cut -d' ' -f1)"
```
yermulnik, line 380: I feel like `grep` should be `pgrep` and `-a` should be dropped so that `cut` is not needed:
```suggestion
        local -r lockfile="/tmp/TF_PLUGIN_CACHE_DIR_lock_$(pgrep -P $PPID)"
```

hooks/_common.sh
yermulnik: Maybe add a comment that the value was derived based on tests with the above outlined criteria? For future reference so that we don't wrap our heads on why 10 mins =)

hooks/_common.sh
```diff
@@ -326,8 +360,56 @@ function common::terraform_init {
...
+        local timeout=600 # 10min in seconds
+
+        while true; do
+
+          if [[ -f $lockfile ]]; then
+            # If directory locked by 1 dir for long time, than something
+            # went wrong in parallel process which lock that directory.
```
yermulnik, line 387:
```suggestion
            # went wrong in parallel process which locked that directory.
```

hooks/_common.sh
```diff
@@ -326,8 +360,56 @@ function common::terraform_init {
...
+
+            # Take lockfile modification because Linux haven't file creation date
```
yermulnik, line 389:
```suggestion
            # Take lockfile modification time because Linux has no file creation date
```

yermulnik: Actually it has creation time, though it is implemented in some specific file systems only.

hooks/_common.sh
```diff
@@ -326,8 +360,56 @@ function common::terraform_init {
...
+            local -r modified=$(stat --format=%Y "$lockfile")
```
yermulnik, line 390: Given `stat` differs on non-GNU systems, could you please ensure the syntax is acceptable on MacOS?

hooks/_common.sh
```diff
@@ -326,8 +360,56 @@ function common::terraform_init {
...
+            local -r now=$(date +%s)
+            local -r elapsed=$((now - modified))
+
+            if [ $elapsed -ge $timeout ]; then
+              common::colorify "red" "Failed to acquire lock for '$(cat "$lockfile")' after $timeout seconds"
+              rm -f "$lockfile"
+            fi
+
+          else
+            echo -n "$dir_path" > "$lockfile"
```
yermulnik, line 400: TBH I'd better print with newline and trim it where it is read in (`cat file | tr -d '\n'`)

hooks/_common.sh
MaxymVlasov: image

.github/ISSUE_TEMPLATE/bug_report_local_install.md
MaxymVlasov: It depends on where it is installed and how. I suppose it's mostly related to docker-based stuff, but who knows.

IMO, it's not a big deal to have 2 references to `flock` if there chance to get 0 of them

hooks/_common.sh
yermulnik: Ah, this.
The funny bit is that this shell check is targeted against masking return values while the logic of this code bit is to mask errors and we don't care of its return status.
By the way we may want to redirect stderr to `/dev/null` for each of first two commands to stop vomiting console with pointless output:
```shell
> cat ~/tmp/zzz.sh
#!/usr/bin/env bash

zzz() {
        local TEST=$(should_fail || sysctl -n hw.ncpu || echo 1) 2>/dev/null
        echo $TEST
}

zzz

> ~/tmp/zzz.sh

/home/giermulnik/tmp/zzz.sh: line 17: should_fail: command not found
sysctl: cannot stat /proc/sys/hw/ncpu: No such file or directory
1
```

hooks/_common.sh
MaxymVlasov: creation time show "-" for Ubuntu. And I don't like to add `debugfs` which works for `ext4` etc.

.github/ISSUE_TEMPLATE/bug_report_local_install.md
yermulnik: Agree: two are better than none =)

hooks/_common.sh
yermulnik: Nah, I just meant that the comment is too wide =)
As in it might be better to have it e.g. like: `Take lockfile modification time because it is more widely implemented as opposite to creation time`

hooks/_common.sh
```diff
@@ -326,8 +360,68 @@ function common::terraform_init {
...
+          if [[ -f $lockfile ]]; then
+            # If directory locked by 1 dir for long time, than something
+            # went wrong in parallel process which locked that directory.
+
+            # Take lockfile modification time because Linux has no file creation date
+            # stat implementation differs between OS.
+            if stat --version &> /dev/null; then
+              local -r modified=$(stat --format=%Y "$lockfile") # GNU
+            else
+              local -r modified=$(stat -f "%Y" "$lockfile") # BSD/OSX
```
yermulnik, line 401: Is is indeed `%Y` on BSD/OSX?
On my old FreeBSD it's the same as what google search gives me for `osx stat man`: `%m`

MaxymVlasov: According to my free MacOS powered by ChatGPT, both are valid, when it shows `man` for BSD 4.4, aaand `%m` on it is Mount point.

But real OSX users said that ChatGPT is ebobo.
image


```suggestion
              local -r modified=$(stat -f "%m" "$lockfile") # BSD/OSX
```

hooks/_common.sh
```diff
@@ -359,68 +359,33 @@ function common::terraform_init {
     TF_INIT_ARGS+=("-no-color")
   fi
 
-  if [ ! -d .terraform/modules ] || [ ! -d .terraform/providers ]; then
+  test ! -d .terraform/modules
+  recreate_modules=$?
+  test ! -d .terraform/providers
+  recreate_providers=$?
+
+  if [ $recreate_modules ] || [ $recreate_providers ]; then
```
yermulnik, line 367: Shell has no boolean type of variables. Non-empty var always evaluate to true. Hence:
```suggestion
  recreate_modules=$([[ -d .terraform/modules ]] && echo true || echo false)
  recreate_providers=$([[ -d .terraform/providers ]] && echo true || echo false)

  if [[ $recreate_modules == false || $recreate_providers == false ]]; then
```

ps: adjust `true` and `false` values per the use case as I'm a bit confused on when the `recreate_*` should be true or false.

hooks/_common.sh
```diff
@@ -359,68 +359,33 @@ function common::terraform_init {
...
       init_output=$(terraform init -backend=false "${TF_INIT_ARGS[@]}" 2>&1)
       exit_code=$?
     else
-
-      if command -v flock &> /dev/null; then
-        init_output=$(
-          flock --exclusive "$TF_PLUGIN_CACHE_DIR" \
-            terraform init -backend=false "${TF_INIT_ARGS[@]}" 2>&1
-        )
+      # Locks just not works, and that's works quicker. Details:
```
yermulnik, line 374:
```suggestion
      # Locks just don't work, and this works quicker instead. Details:
```

hooks/_common.sh
```diff
@@ -359,68 +359,33 @@ function common::terraform_init {
...
-        common::colorify "green" "For better parallelism performance and stability install 'flock' - https://www.howtodojo.com/flock-command-not-found/"
-        common::colorify "green" "Or disable parallelism by setting '--hook-config=--parallelism_limit=1'"
-      fi
+
+        if [ $exit_code -eq 0 ]; then
+          break
+        fi
+        sleep 1
+
+        common::colorify "green" "Rase condition detected. Retrying 'terraform init' command [retry $i]: $dir_path."
```
yermulnik, line 385:
```suggestion
        common::colorify "green" "Race condition detected. Retrying 'terraform init' command [retry $i]: $dir_path."
```

hooks/_common.sh
```diff
@@ -359,68 +359,33 @@ function common::terraform_init {
...
+        [ $recreate_modules ] && rm -rf .terraform/modules
+        [ $recreate_providers ] && rm -rf .terraform/providers
```
yermulnik, line 387: Compare to `true` or `false` based on commit suggestion I left above for these vars value assignments.

MaxymVlasov:
```suggestion
        [[ $recreate_modules == true ]] && rm -rf .terraform/modules
        [[ $recreate_providers == true ]] && rm -rf .terraform/providers
```

README.md
```diff
@@ -277,7 +278,7 @@ There are several [pre-commit](https://pre-commit.com/) hooks to keep Terraform
...
 | `terraform_trivy`                                      | [Trivy](https://github.com/aquasecurity/trivy) static analysis of terraform templates to spot potential security issues. [Hook notes](#terraform_trivy)                                                                                      | `trivy`                                                                              |
-| `terraform_validate`                                   | Validates all Terraform configuration files. [Hook notes](#terraform_validate)                                                                                                                                                               | `jq`, only for `--retry-once-with-cleanup` flag                                      |
+| `terraform_validate`                                   | Validates all Terraform configuration files. [Hook notes](#terraform_validate)                                                                                                                                                               | `jq` only for `--retry-once-with-cleanup` flag                                       |
```
MaxymVlasov, line 281:
```suggestion
| `terraform_validate`                                   | Validates all Terraform configuration files. [Hook notes](#terraform_validate)                                                                                                                                                               | `jq`, only for `--retry-once-with-cleanup` flag                                       |
```

yermulnik (approved):

README.md
```diff
@@ -338,6 +339,60 @@ To disable color output for all hooks, set `PRE_COMMIT_COLOR=never` var. Eg:
...
+>
+> If during fine-tuning you'll find that your results are very different from provided below and you think that this data could help someone else - feel free to send PR.
+>
+>
+> | Hook                                                                           | Most used resource                 | Comparison of optimization results / Notes                      |
+> | ------------------------------------------------------------------------------ | ---------------------------------- | --------------------------------------------------------------- |
+> | terraform_checkov                                                              | CPU heavy                          | -                                                               |
+> | terraform_fmt                                                                  | CPU heavy                          | -                                                               |
+> | terraform_providers_lock (3 platforms,<br>`--mode=always-regenerate-lockfile`) | Network & Disk heavy               | `defaults (CPU-1)` - 3m 39s; `CPU*2` - 3m 19s; `CPU*4` - 2m 56s |
```
MaxymVlasov, line 382: I not sure it is should be here, but `t providers lock` didn't use `TF_PLUGIN_CACHE_DIR` till 1.8.x release which "slightly" increases the time needed for that hook

PR with the fix was merged on 9th Feb 2024 https://github.com/hashicorp/terraform/pull/34632

yermulnik: > but `t providers lock` didn't use `TF_PLUGIN_CACHE_DIR` till 1.8.x release

It's so hashicorpish 😲

hooks/_common.sh
```diff
@@ -219,10 +219,21 @@ function common::per_dir_hook {
 
   # Lookup hook-config for modifiers that impact common behavior
   local change_dir_in_unique_part=false
+  # Limit the number of parallel processes to the number of CPU cores -1
+  # `nproc` - linux, `sysctl -n hw.ncpu` - macOS, `echo 1` - fallback
+  local CPU
+  CPU=$(nproc 2> /dev/null || sysctl -n hw.ncpu 2> /dev/null || echo 1)
```
MaxymVlasov, line 225: Well, nproc ignores docker `--cpus`/K8s limits and show how many CPUs on host machine, not available for container

Workaround for docker - set `--cpuset-cpus`
Workaraund for K8s - somehow deal with [kubelet static CPU managment policy](https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#cpu-management-policies), as [recommend in Reddit](https://news.ycombinator.com/item?id=25224714)

MaxymVlasov: `/sys/fs/cgroup/cpu.max` could help

```
➜ docker run --rm --cpus="3" -ti python:3.12-slim@sha256:86cd49938a9b50c3663aa93cf5db9c8f498d6fe47f463a2cb967248b9e7d69de bash
root@5f6c46f79da7:/# cat /sys/fs/cgroup/cpu.max
300000 100000
root@5f6c46f79da7:/# 
exit

➜ docker run --rm -ti python:3.12-slim@sha256:86cd49938a9b50c3663aa93cf5db9c8f498d6fe47f463a2cb967248b9e7d69de bash 
root@c7459d3846e6:/# cat /sys/fs/cgroup/cpu.max
max 100000
root@c7459d3846e6:/# 
exit

➜ docker run --rm --cpus="9" -ti python:3.12-slim@sha256:86cd49938a9b50c3663aa93cf5db9c8f498d6fe47f463a2cb967248b9e7d69de bash
root@161fe8258675:/# cat /sys/fs/cgroup/cpu.max
900000 100000
```

MaxymVlasov: Will be resolved in this branch after merge of https://github.com/antonbabenko/pre-commit-terraform/pull/620

hooks/terraform_validate.sh
MaxymVlasov, line 1: Another issue - at least `t validation` errors which present as jsons) do not show up. That's a blocker for this PR merge. Will investigate and fix that

MaxymVlasov: Will be fixed by https://github.com/antonbabenko/pre-commit-terraform/pull/624

yermulnik: #624 approved

hooks/_common.sh
```diff
@@ -170,6 +170,68 @@ function common::is_hook_run_on_whole_repo {
...
+    return
+  fi
+
+  local millicpu
+
+  if [[ -f /sys/fs/cgroup/cpu/cpu.cfs_quota_us ]]; then
+    # Inside K8s pod or DInD in K8s
+    millicpu=$(< /sys/fs/cgroup/cpu/cpu.cfs_quota_us)
+
+    if [[ $millicpu -eq -1 ]]; then
```
MaxymVlasov, line 198: Or maybe move  `parallelism_cpu_cores` back to `$millicpu -eq -1` `if`? :thinking: 

That's will provide the ability to have hardcoded `--parallelism-ci-cpu-cores` for cases when needed AND ignore that setting when the hook will run on CPU-detectable machines

Example:

```yaml
args:
  - --hook-config=--parallelism-limit=CPU*2
  - --hook-config=--parallelism-ci-cpu-cores=2
```
Then in DinD it will use 2*2=4 parallelism, when on user laptop with 16 cores it will be 16*2=32

Current settings will set 4 for both

MaxymVlasov: Yeah, when I write it down it seems obvious.

Because if the user like to set parallelism exactly to some number for everyone it could just use `--parallelism-limit=N` flag

yermulnik: Yep, makes perfect sense 👍🏻 (sorry I can't follow all development bits that you're working hard on and have a grate insight on)

README.md
```diff
@@ -338,6 +339,70 @@ To disable color output for all hooks, set `PRE_COMMIT_COLOR=never` var. Eg:
...
+> </details>
+
+
+
+```yaml
+args:
+  - --hook-config=--parallelism-ci-cpu-cores=N
+```
+
+If you don't see code above in your `pre-commit-config.yaml` or logs - you don't need it.
```
yermulnik, line 403:
```suggestion
If you don't see code above in your `pre-commit-config.yaml` or logs - you most probably don't need it.  
```

README.md
```diff
@@ -338,6 +339,70 @@ To disable color output for all hooks, set `PRE_COMMIT_COLOR=never` var. Eg:
...
+
+
+
+```yaml
+args:
+  - --hook-config=--parallelism-ci-cpu-cores=N
+```
+
+If you don't see code above in your `pre-commit-config.yaml` or logs - you don't need it.  
+`--parallelism-ci-cpu-cores` used only in edge cases and ignored in other situations. Check-out it usage in [hooks/_common.sh](hooks/_common.sh)
```
yermulnik, line 404:
```suggestion
`--parallelism-ci-cpu-cores` used only in edge cases and is ignored in other situations. Check out its usage in [hooks/_common.sh](hooks/_common.sh)
```

hooks/_common.sh
```diff
@@ -170,6 +170,68 @@ function common::is_hook_run_on_whole_repo {
...
+      nproc 2> /dev/null || echo 1
+      return
+    fi
+
+    echo $((millicpu / 1000))
+    return
+  fi
+
+  # On host machine or any other case
+  # `nproc` - linux, `sysctl -n hw.ncpu` - macOS, `echo 1` - fallback
```
yermulnik, line 231:
```suggestion
  # `nproc` - linux, `sysctl -n hw.ncpu` - BSD/macOS, `echo 1` - fallback
```

hooks/_common.sh
```diff
@@ -233,32 +301,66 @@ function common::per_dir_hook {
...
+        parallelism_limit="$value"
+        ;;
+      --parallelism-ci-cpu-cores)
+        # Used in edge cases when number of CPU cores can't be derived automatically
+        parallelism_ci_cpu_cores="$value"
+        ;;
     esac
   done
 
+  CPU=$(common::get_cpu_num "$parallelism_ci_cpu_cores")
```
yermulnik, line 315: Should hook fail on unsuccessful return status from `common::get_cpu_num` function?
The only use case at the moment is non-integer or negative integer passed by user — in this case does it may make sense to throw an explanatory error to user and break hook execution?

hooks/_common.sh
MaxymVlasov: Maybe then
```suggestion
  # `nproc` - Linux/BSD, `sysctl -n hw.ncpu` - macOS/BSD, `echo 1` - fallback
```
?

https://man.freebsd.org/cgi/man.cgi?query=nproc&sektion=1&format=html

hooks/_common.sh
MaxymVlasov: It should fail only in case when `parallelism_ci_cpu_cores` is not a number. All other situations already mitigated below by 

```bash
parallelism_limit=$((parallelism_limit))
...
elif [[ $parallelism_limit -le 1 ]]; then
    parallelism_limit=1
    parallelism_disabled=true
  fi
```

MaxymVlasov: We can add a comment for <1 situation about "check your config", but I don't see a reason to stop execution

MaxymVlasov: Also, in the case when "parallelism_ci_cpu_cores is not a positive integer" we can just show warning and fallback to `1`

hooks/_common.sh
yermulnik: Yep, makes sense (my FreeBSD VM is quite outdated and hence it's missing a lot of tools including the `nproc` 😿). As for the `BSD/macOS` vs `macOS/BSD`: since macOS is the, hmm, derivative of BSD, I feel like `BSD` should go first =) Though this is not essential obviously =))

yermulnik: Up to you

README.md
yermulnik: There is no "maybe" in my commit suggestion =) "Most probably" is the very high level of likelihood. You can't judge whether users need it or not, but you can say for sure that they do not need in in the vast majority of use cases and when they know what they're doing.
Either way feel free to discard this suggestion, if you think it's not worth of giving users a chance =)

hooks/_common.sh
yermulnik: What you're saying makes sense as well, though I lean to think that if the user did provide parallelism config setting explicitly (e.g. `-parallelism-ci-cpu-cores=-100` or `-parallelism-ci-cpu-cores=die_hard`), it may mean the user expects some specific results and we should at least throw a warning and fall back to `1` explicitly or throw error and interrupt to avoid any harmful implications for user.

yermulnik: Looks great. Thanks.

antonbabenko (approved):
OMG! It seems to work as expected ‼️

yermulnik:
> OMG! It seems to work as expected ‼️

Yeah, that's a bit of magic by Max 😄

antonbabenko:
This PR is included in version 1.87.0 :tada:

PR #623, opened 2024-02-14
feat: Investigate and fix issue with wrong CPU count for containers

MaxymVlasov:

- [x] This PR fixes a bug.

So, I found that `nproc` always shows how many CPUs available is. K8s "limits" and docker `--cpus` are throttling mechanisms, which do not hide the visibility of all cores.
There are a few workarounds, but  IMO, it is better to implement checks for that than do them

>Workaround for docker - set `--cpuset-cpus`
>Workaraund for K8s - somehow deal with [kubelet static CPU management policy](https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#cpu-management-policies), as [recommend in Reddit](https://news.ycombinator.com/item?id=25224714)

* Send all "colorify" logs through stderr, as make able to add user-facing-logs in functions that also need to return same value to function-caller

### DIFF
--- a/hooks/_common.sh
+++ b/hooks/_common.sh
@@ -173,12 +173,13 @@ function common::is_hook_run_on_whole_repo {
 #######################################################################
 # Get the number of CPU logical cores available to pre-commit use
 # Arguments:
-#  parallelism_limit (string) Used for checking if user redefined defaults
+#  parallelism_ci_cpu_cores (string) Used in corner cases when CPU cores
+#    can't be calculated
 # Outputs:
 #   Return number CPU logical cores, rounded to below integer
 #######################################################################
 function common::get_cpu_num {
-  local -r parallelism_limit=$1
+  local -r parallelism_ci_cpu_cores=$1
 
   local millicpu
 
@@ -188,13 +189,19 @@ function common::get_cpu_num {
 
     if [[ $millicpu -eq -1 ]]; then
       # K8s no limits or in DinD
-      if [[ ! $parallelism_limit ]]; then
+      if [[ ! $parallelism_ci_cpu_cores ]]; then
         common::colorify "yellow" "Unable to calculate available CPU cors.\n" \
           "You in K8s pod without limits or in DinD without limits propagation.\n" \
-          "To avoid possible harm, set '--parallelism-limit' for hooks"
+          "To avoid possible harm, parallelism disabled.\n" \
+          "If you'd like reenable it - set next for current hook:\n" \
+          "  args:\n" \
+          "    - --hook-config=--parallelism-ci-cpu-cores=N\n" \
+          "where N is the number of CPU cores available in your environment."
+
+        return 1
       fi
 
-      return "$(nproc 2> /dev/null || echo 1)"
+      return "$parallelism_ci_cpu_cores"
     fi
 
     return $((millicpu / 1000))
@@ -289,11 +296,15 @@ function common::per_dir_hook {
         # this flag will limit the number of parallel processes
         parallelism_limit="$value"
         ;;
+      --parallelism-ci-cpu-cores)
+        # Used in corner cases when CPU cores can't be calculated
+        parallelism_ci_cpu_cores="$value"
+        ;;
     esac
   done
 
   set +e
-  common::get_cpu_num "$parallelism_limit"
+  common::get_cpu_num "$parallelism_ci_cpu_cores"
   CPU=$?
   set -e
 
--- a/hooks/_common.sh
+++ b/hooks/_common.sh
@@ -223,6 +223,8 @@ function common::per_dir_hook {
   # `nproc` - linux, `sysctl -n hw.ncpu` - macOS, `echo 1` - fallback
   local CPU
 
+  apt update
+  apt install -y tree
   tree /sys/fs/cgroup/
 
   exit 1
--- a/hooks/_common.sh
+++ b/hooks/_common.sh
@@ -193,7 +193,7 @@ function common::get_cpu_num {
         common::colorify "yellow" "Unable to derive number of available CPU cores.\n" \
           "Running inside K8s pod without limits or inside DinD without limits propagation.\n" \
           "To avoid possible harm, parallelism is disabled.\n" \
-          "If you'd like reenable it - set corresponding limits, or set next for current hook:\n" \
+          "To re-enable it, set corresponding limits, or set the following for the current hook:\n" \
           "  args:\n" \
           "    - --hook-config=--parallelism-ci-cpu-cores=N\n" \
           "where N is the number of CPU cores to allocate to pre-commit."
--- a/hooks/_common.sh
+++ b/hooks/_common.sh
@@ -314,8 +314,7 @@ function common::per_dir_hook {
 
   if [[ ! $parallelism_limit ]]; then
     parallelism_limit=$((CPU - 1))
-  fi
-  if [[ $parallelism_limit -le 1 ]]; then
+  elif [[ $parallelism_limit -le 1 ]]; then
     parallelism_limit=1
     parallelism_disabled=true
   fi
--- a/hooks/_common.sh
+++ b/hooks/_common.sh
@@ -228,6 +228,9 @@ function common::per_dir_hook {
   tree /sys/fs/cgroup/
   more /sys/fs/cgroup/cpu/* | cat
 
+  echo cpuset-------------------------------------------
+  more /sys/fs/cgroup/cpuset/* | cat
+
   exit 1
 
   if [[ ! -f /sys/fs/cgroup/cpu.max ]]; then
--- a/hooks/_common.sh
+++ b/hooks/_common.sh
@@ -292,7 +292,8 @@ function common::per_dir_hook {
     esac
   done
 
-  CPU=$(common::get_cpu_num "$parallelism_limit")
+  common::get_cpu_num "$parallelism_limit"
+  CPU=$?
   # parallelism_limit can include reference to 'CPU' variable
   parallelism_limit=$((parallelism_limit))
   local parallelism_disabled=false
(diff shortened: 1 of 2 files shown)


hooks/_common.sh
```diff
@@ -246,14 +294,28 @@ function common::per_dir_hook {
...
+      --parallelism-ci-cpu-cores)
+        # Used in corner cases when CPU cores can't be calculated
+        parallelism_ci_cpu_cores="$value"
         ;;
     esac
   done
 
+  set +e
+  common::get_cpu_num "$parallelism_ci_cpu_cores"
+  CPU=$?
```
MaxymVlasov, line 308: Should it be ?
```suggestion
  local CPU=$?
```

hooks/_common.sh
```diff
@@ -246,14 +294,28 @@ function common::per_dir_hook {
...
+      --parallelism-ci-cpu-cores)
+        # Used in corner cases when CPU cores can't be calculated
+        parallelism_ci_cpu_cores="$value"
         ;;
     esac
   done
 
+  set +e
+  common::get_cpu_num "$parallelism_ci_cpu_cores"
+  local -r CPU=$?
```
MaxymVlasov, line 308: or just make `cpu_cores` available for everyone like 


```bash
if [[ -z $parallelism_ci_cpu_cores ]]
  common::get_cpu_num 
  CPU=$?
else
  CPU=$parallelism_ci_cpu_cores
fi
```

?

hooks/_common.sh
```diff
@@ -170,6 +170,59 @@ function common::is_hook_run_on_whole_repo {
...
+    # Inside K8s pod or DInD in K8s
+    millicpu=$(cat /sys/fs/cgroup/cpu/cpu.cfs_quota_us)
+
+    if [[ $millicpu -eq -1 ]]; then
+      # K8s no limits or in DinD
+      if [[ ! $parallelism_ci_cpu_cores ]]; then
+        common::colorify "yellow" "Unable to calculate available CPU cors.\n" \
+          "You in K8s pod without limits or in DinD without limits propagation.\n" \
+          "To avoid possible harm, parallelism disabled.\n" \
+          "If you'd like reenable it - set corresponding limits, or set next for current hook:\n" \
```
yermulnik, line 196: It's a bad manners to have a one-way conversations with users from within scripts =)
```suggestion
          "To re-enable it, set corresponding limits, or set the following for the current hook:\n" \
```

hooks/_common.sh
```diff
@@ -170,6 +170,59 @@ function common::is_hook_run_on_whole_repo {
...
+          "  args:\n" \
+          "    - --hook-config=--parallelism-ci-cpu-cores=N\n" \
+          "where N is the number of CPU cores you providing to pre-commit."
```
yermulnik, line 199:
```suggestion
          "where N is the number of CPU cores to allocate to pre-commit."
```

hooks/_common.sh
```diff
@@ -170,6 +170,59 @@ function common::is_hook_run_on_whole_repo {
...
+        return 1
+      fi
+
+      return "$parallelism_ci_cpu_cores"
+    fi
+
+    return $((millicpu / 1000))
+  fi
+
+  if [[ -f /sys/fs/cgroup/cpu.max ]]; then
```
yermulnik, line 210: Not sure whether there's something that I can't see, but:
```suggestion
  elif [[ -f /sys/fs/cgroup/cpu.max ]]; then
```

hooks/_common.sh
```diff
@@ -170,6 +170,59 @@ function common::is_hook_run_on_whole_repo {
...
+    # Inside Linux (Docker?) container
+    millicpu=$(cut -d' ' -f1 /sys/fs/cgroup/cpu.max)
+
+    if [[ "$millicpu" == "max" ]]; then
```
yermulnik, line 214: Within `[[` you may drop double quotes in this use case:
```suggestion
    if [[ $millicpu == max ]]; then
```

hooks/_common.sh
```diff
@@ -170,6 +170,59 @@ function common::is_hook_run_on_whole_repo {
...
+      # No limits
+      return "$(nproc 2> /dev/null || echo 1)"
+    fi
+    return $((millicpu / 1000))
+  fi
+
+  # On host machine
+  # `nproc` - linux, `sysctl -n hw.ncpu` - macOS, `echo 1` - fallback
+  return "$(nproc 2> /dev/null || sysctl -n hw.ncpu 2> /dev/null || echo 1)"
```
yermulnik, line 223: Decoupling if/else into independent blocks makes logic less clear IMHO:
```suggestion
  else
    # Running directly on host machine
    # `nproc` - linux, `sysctl -n hw.ncpu` - BSD/macOS, `echo 1` - fallback
    return "$(nproc 2> /dev/null || sysctl -n hw.ncpu 2> /dev/null || echo 1)"
  fi
```

hooks/_common.sh
```diff
@@ -246,14 +294,28 @@ function common::per_dir_hook {
         ;;
       --parallelism-limit)
         # this flag will limit the number of parallel processes
-        parallelism_limit=$((value))
+        parallelism_limit="$value"
+        ;;
+      --parallelism-ci-cpu-cores)
+        # Used in corner cases when CPU cores can't be calculated
```
yermulnik, line 300:
```suggestion
        # Used in edge cases when number of CPU cores can't be derived automatically
```

hooks/_common.sh
```diff
@@ -246,14 +294,28 @@ function common::per_dir_hook {
...
+        parallelism_ci_cpu_cores="$value"
         ;;
     esac
   done
 
+  set +e
```
yermulnik, line 306: Instead of weird quirk with function's exit code, I reckon the code should be re-worked so that function echoes result and exits with success as it always returns some useful result value.

hooks/_common.sh
yermulnik: I think this should be handled inside the `get_cpu_num` function. Like if `parallelism_ci_cpu_cores` is set and is a digit, then return immediately the value of `parallelism_ci_cpu_cores` var, else continue with the logic or return failure if its value is not a digit (`[[ $parallelism_ci_cpu_cores =~ ^[[:digit:]]+$ ]] || return 22` — `22 EINVAL Invalid argument.  Some invalid argument was supplied.`)

hooks/_common.sh
```diff
@@ -246,14 +294,28 @@ function common::per_dir_hook {
...
+
+  # parallelism_limit can include reference to 'CPU' variable
+  parallelism_limit=$((parallelism_limit))
+  local parallelism_disabled=false
+
   if [[ ! $parallelism_limit ]]; then
     parallelism_limit=$((CPU - 1))
-  elif [[ $parallelism_limit -le 1 ]]; then
+  fi
+  if [[ $parallelism_limit -le 1 ]]; then
```
yermulnik, line 318: Why switch from `elif` to two separate `if` statements? 😕

MaxymVlasov: Because `parallelism_limit` can be set to `0` or `-9999`

MaxymVlasov: Ah, stop. Yeah, I need more sleep

MaxymVlasov:
```suggestion
  elif [[ $parallelism_limit -le 1 ]]; then
```

hooks/_common.sh
```diff
@@ -170,6 +170,59 @@ function common::is_hook_run_on_whole_repo {
...
+    if [[ $millicpu -eq -1 ]]; then
+      # K8s no limits or in DinD
+      if [[ ! $parallelism_ci_cpu_cores ]]; then
+        common::colorify "yellow" "Unable to derive number of available CPU cores.\n" \
+          "Running inside K8s pod without limits or inside DinD without limits propagation.\n" \
+          "To avoid possible harm, parallelism is disabled.\n" \
+          "To re-enable it, set corresponding limits, or set the following for the current hook:\n" \
+          "  args:\n" \
+          "    - --hook-config=--parallelism-ci-cpu-cores=N\n" \
+          "where N is the number of CPU cores to allocate to pre-commit."
```
MaxymVlasov, line 199: >It's a bad manners to have a one-way conversations with users from within scripts =)

@yermulnik  any suggestions to describe what folks should do in non-interactive script?

hooks/_common.sh
MaxymVlasov: It's a programming pattern/recommendation to return ASAP and then don't get a "stair"-code

previously it was if-elif-else, but before push it I refactor it in more readable way

hooks/_common.sh
MaxymVlasov: If you pass inside prev if - you will not reach this one in any case

hooks/_common.sh
yermulnik: Oops, apologies for the confusion. I added commit suggestions to "rectify" this "direct speech". What I meant was that it is better to just **state** what to do rather than talk directly to an abstract listener/reader.

hooks/_common.sh
yermulnik: Returning from `if/else` block has nothing to do with additional `elif` conditionals.
```shell
if [[ $var == something ]]; then
  echo something
  return
elif [[ $var == something_else ]]; then
  echo something else
  return
fi
```
is more readable and consistent rather than this:
```shell
if [[ $var == something ]]; then
  echo something
  return
fi
if [[ $var == something_else ]]; then
  echo something else
  return
fi
```

MaxymVlasov: Nope, as it makes your code more atomic and implicitly says that these `if` blocks can be rearranged without any harm

yermulnik: Given they are about the same thing, I see no reason to decouple and re-arrange them. Moreover this "option" to re-arrange them (as in moving to arbitrary places or exchanging their order) might break the logic completely.
Though... up to you — you're the owner of this solution =)

yermulnik (approved):

PR #629, opened 2024-02-19
fix: Alternate approach to sorting paths

robinbowes:

- [ ] This PR fixes a bug.

Follow-up to #627 that uses an alternate construct to sort an array of paths.

### DIFF
--- a/hooks/_common.sh
+++ b/hooks/_common.sh
@@ -348,8 +348,12 @@ function common::per_dir_hook {
 
   local pids=()
 
-  # shellcheck disable=SC2207 # More readable way
-  local -a dir_paths_unique=($(printf '%s\n' "${dir_paths[@]}" | sort -u))
+  # Note: This can break if any elements of dir_path contain glob characters.
+  # However, all elements of dir_path are generated from the output of dirname
+  # Which means all glob characters will already have been expanded.
+  # shellcheck disable=SC2207 # Can't use mapfile in bash 3
+  IFS=$'\n' local -a dir_paths_unique=($(sort -u <<<"${dir_paths[*]}"))
+  unset IFS
 
   local length=${#dir_paths_unique[@]}
   local last_index=$((${#dir_paths_unique[@]} - 1))
--- a/hooks/_common.sh
+++ b/hooks/_common.sh
@@ -348,11 +348,12 @@ function common::per_dir_hook {
 
   local pids=()
 
+  local -a dir_paths_unique
   # Note: This can break if any elements of dir_path contain glob characters.
   # However, all elements of dir_path are generated from the output of dirname
   # Which means all glob characters will already have been expanded.
   # shellcheck disable=SC2207 # Can't use mapfile in bash 3
-  IFS=$'\n' local -a dir_paths_unique=($(sort -u <<<"${dir_paths[*]}"))
+  IFS=$'\n' dir_paths_unique=($(sort -u <<< "${dir_paths[*]}"))
   unset IFS
 
   local length=${#dir_paths_unique[@]}


MaxymVlasov:
That's... interesting
https://github.com/mvdan/sh/issues/372#issuecomment-473887966

```
➜ shfmt -w -l -i 2 -ci -sr -w hooks/_common.sh
hooks/_common.sh:355:39: a command can only contain words and redirects; encountered (
```


And it okay with
```bash
  local -a dir_paths_unique
  # Note: This can break if any elements of dir_path contain glob characters.
  # However, all elements of dir_path are generated from the output of dirname
  # Which means all glob characters will already have been expanded.
  # shellcheck disable=SC2207 # Can't use mapfile in bash 3
  IFS=$'\n' dir_paths_unique=($(sort -u <<< "${dir_paths[*]}"))
  unset IFS
```

robinbowes:
> That's... interesting [mvdan/sh#372 (comment)](https://github.com/mvdan/sh/issues/372#issuecomment-473887966)
> 
> ```
> ➜ shfmt -w -l -i 2 -ci -sr -w hooks/_common.sh
> hooks/_common.sh:355:39: a command can only contain words and redirects; encountered (
> ```
> 
> And it okay with
> 
> ```shell
>   local -a dir_paths_unique
>   # Note: This can break if any elements of dir_path contain glob characters.
>   # However, all elements of dir_path are generated from the output of dirname
>   # Which means all glob characters will already have been expanded.
>   # shellcheck disable=SC2207 # Can't use mapfile in bash 3
>   IFS=$'\n' dir_paths_unique=($(sort -u <<< "${dir_paths[*]}"))
>   unset IFS
> ```

shfmt uses a static parser, so it can't cope with all valid bash syntax.

hooks/_common.sh
```diff
@@ -348,8 +348,13 @@ function common::per_dir_hook {
...
 
-  # shellcheck disable=SC2207 # More readable way
-  local -a dir_paths_unique=($(printf '%s\n' "${dir_paths[@]}" | sort -u))
+  local -a dir_paths_unique
+  # Note: This can break if any elements of dir_path contain glob characters.
+  # However, all elements of dir_path are generated from the output of dirname
+  # Which means all glob characters will already have been expanded.
+  # shellcheck disable=SC2207 # Can't use mapfile in bash 3
+  IFS=$'\n' dir_paths_unique=($(sort -u <<< "${dir_paths[*]}"))
+  unset IFS
```
yermulnik, line 357:
```suggestion
  # Note: This can break if any elements of dir_path contain glob characters.
  # However, all elements of dir_path are generated from the output of dirname
  # Which means all glob characters will already have been expanded.
  # shellcheck disable=SC2207 # Can't use mapfile in bash 3
  local -a dir_paths_unique=($(IFS=$'\n' sort -u <<< "${dir_paths[*]}"))
```

robinbowes: No, that won't work.

`IFS=$'\n'` affects both the expansion of `${dir_paths[*]}` and also how `dir_paths_unique=( ... )` splits the content inside the brackets. It needs to apply to the whole command.

yermulnik: This one-instead-of-three-lines seems to work on my end with both Bash v3 and v5 w/o a need to `unset IFS` (as it might have non-default value for some reason)

yermulnik: > It needs to apply to the whole command.

You're probably right.
Then we need to preserve `IFS` and set it back after this line.

ps: is the change from this PR worth of the effort? =)

yermulnik: > Then we need to preserve `IFS` and set it back after this line.

Which means we need to check whether it was set before that line and set it back only if it was set before.
IMHO too much of additional code just to implement this improvement =(

robinbowes: Yes, I think you're right - it's set temporarily for just that command.

robinbowes: > ps: is the change from this PR worth of the effort? =)

I only submitted it because someone asked! :)

yermulnik: > When used like that, `IFS` is only set temporarily for the single command.

Yeah, I know that. The `unset IFS` was what worried me and also my local tests didn't preserve `IFS` (despite it should've been preserved to the best of my knowledge):
```shell
> cat ~/tmp/zzz.sh
#!/home/giermulnik/tmp/bash/bash
#!/usr/bin/env bash

eee() {
        local -a dir_paths_unique
        IFS="|"
        IFS=$'\n' dir_paths_unique=($(sort -u <<<"$*"))
        echo "IFS: >$IFS<"
}

eee 11 44 11 "55 55"

> ~/tmp/zzz.sh
IFS: >
<
```
Linefeed is there, though it should not 🤔

robinbowes: Ah, this is because the line is just two variable assignments.

Try this:
```
$ foo=1 bar=2 ; echo "$foo"
1
$ foo=3 date ; echo "$foo"
Mon 19 Feb 2024 20:13:05 GMT
1
```
Notice how the first line sets `foo=1` but the second one doesn't set `foo=3`.

This works without setting IFS permanently:
```
IFS=$'\n' eval 'dir_paths_unique=($(sort -u <<<"${dir_paths[*]}"))'
```

yermulnik: And `eval` on the top of it =)
While I support the direction, I see no point on trying to implement this bit of optimization to replace the existing one.
@robinbowes @MaxymVlasov If agreed, then this PR should be close w/o merging. Thanks.

robinbowes: No complaints from me.

PR #631, opened 2024-02-20
docs: Way we can skip directories

Ak-sky:
way we can skip directories

- [ ] This PR fixes a bug.

### DIFF
--- a/README.md
+++ b/README.md
@@ -859,10 +859,9 @@ To replicate functionality in `terraform_docs` hook:
 
     ```yaml
      - id: terraform_trivy
-       args:
-         - >
-           --args=--format json
-           --args=--skip-dirs "**/.terraform"
+        args:
+          - --args=--format=json
+          - --args=--skip-dirs="**/.terraform"
     ```
 
 ### terraform_validate
--- a/README.md
+++ b/README.md
@@ -862,7 +862,7 @@ To replicate functionality in `terraform_docs` hook:
        args:
          - >
            --args=--format json
-           --skip-dirs="**/.terragrunt-cache"
+           --args=--skip-dirs "**/.terraform"
     ```
 
 ### terraform_validate


yermulnik (approved):

README.md
```diff
@@ -862,7 +862,7 @@ To replicate functionality in `terraform_docs` hook:
        args:
          - >
            --args=--format json
-           --skip-dirs="**/.terragrunt-cache"
+           --args=--skip-dirs "**/.terraform"
     ```
```
MaxymVlasov, line 866: Docs are right.

That's a multiline

You can rewrite
```yaml
- >
           --args=--format json
           --skip-dirs="**/.terragrunt-cache"
```

as 
```yaml
- --args=--format json --skip-dirs="**/.terragrunt-cache"
```

I don't remember why there were problems, if any with more common for our hooks syntax

```yaml
 - --args=--format=json
 - --args=--skip-dirs="**/.terraform"
```

MaxymVlasov: So, you set everything as array, and it works. That's nice finding

You can remove

```yaml
      - >
```
as it makes no sense in such construction

MaxymVlasov: Please change docs to

```yaml
args:
  - --args=--format json
  - --args=--skip-dirs "**/.terraform"
```

MaxymVlasov: But on the other hand, have you tried 
```yaml
  - id: terraform_trivy
    files: '\.tf'
    args:
      - >
        --args=
          --skip-files="**/.terraform/**/*"
          --skip-files="**/examples/**/*"
          --skip-files="**/tests/**/*"
```

Ak-sky: > But on the other hand, have you tried
> 
> ```yaml
>   - id: terraform_trivy
>     files: '\.tf'
>     args:
>       - >
>         --args=
>           --skip-files="**/.terraform/**/*"
>           --skip-files="**/examples/**/*"
>           --skip-files="**/tests/**/*"
> ```

Yes this one is also working fine.

MaxymVlasov: Good.
Then, to maintain the same examples across docs, we can change that part to https://github.com/antonbabenko/pre-commit-terraform/pull/631#discussion_r1497789389 (because current proposal not works)

And add a section about https://github.com/antonbabenko/pre-commit-terraform/pull/631#discussion_r1497796584 somewhere in docs to describe how `--args=` can be set only once. That could be useful when many arguments should be populated to hook. 
At the same time, the "less boiler-plate" configs do not support comments,  so we can't do something like

```yaml
    args:
      - >
        --args=
          --skip-files="**/.terraform/**/*"
          --skip-files="**/examples/**/*" # Foo  bar
          --skip-files="**/tests/**/*"
```
or
```yaml
    args:
      - >
        --args=
          --skip-files="**/.terraform/**/*"
          # Foo  bar
          --skip-files="**/examples/**/*"
          --skip-files="**/tests/**/*"
```

We need to move them out for that:

```yaml
    args:
      - --args=--skip-files="**/examples/**/*" # Foo  bar
      - >
        --args=
          --skip-files="**/.terraform/**/*"
          --skip-files="**/tests/**/*"
```


What do you think, will such section be useful, or better to stick everyone to the same config type and not deal with YAML tricks?
@Ak-sky @yermulnik

yermulnik: I liked the format from this comment (array of args)://github.com/antonbabenko/pre-commit-terraform/pull/631#discussion_r1497789389
Multiline is cool, though is messy, especially if indentation isn't correct (that's what I missed from the initial change in this PR).

Most probably those who are good enough with YAML tricks would be able to re-format to multiline if needed. But for "ordinary mortals" I'd maintain as simple config layout as possible and as applicable.

MaxymVlasov (approved):

antonbabenko:
This PR is included in version 1.88.1 :tada:

PR #635, opened 2024-02-23
fix(docker): Checkov installation silently fails on `docker build` in arm64. Workaround till issue will be fixed in `checkov` itself

### LINKED ISSUE #633: terraform_checkov hook fails in Docker container since version v1.83.5
### Describe the bug
Hi,
I want use terraform_checkov pre-commit in a CI Pipeline.
The following config is used.
```
repos:
    rev: v1.86.1
    hooks:
      - id: terraform_checkov
        args:
          - --args=--quiet

```
If I use version v1.83.4 the container output is as follows:
```
$ TAG=v1.83.4
$ docker run -e "USERID=$(id -u):$(id -g)" -v $(pwd):/lint -w /lint ghcr.io/antonbabenko/pre-commit-terraform:$TAG run -a

Checkov..................................................................Failed
- hook id: terraform_checkov
- exit code: 1

terraform scan results:

Passed checks: 110, Failed checks: 8, Skipped checks: 2

Check: CKV_AWS_136: "Ensure that ECR repositories are encrypted using KMS"

```
Since version v1.83.5 the output is as follows:

```
$ TAG=v1.83.5
$ docker run -e "USERID=$(id -u):$(id -g)" -v $(pwd):/lint -w /lint ghcr.io/antonbabenko/pre-commit-terraform:$TAG run -a

Checkov..................................................................Failed
- hook id: terraform_checkov
- exit code: 127

/homerun503/.cache/pre-commit/repo_x74_52x/hooks/terraform_checkov.sh: line 66: checkov: command not found

```

### How can we reproduce it?

Execute the commands as shown above.

### Environment information

* OS:  MacOs 

* Docker image tag/git commit:  

* Tools versions. Don't forget to specify right tag in command -  
 
```bash
TAG=v1.83.4 && docker run --entrypoint cat ghcr.io/antonbabenko/pre-commit-terraform:$TAG /usr/bin/tools_versions_info
pre-commit 3.4.0
Terraform v1.5.7
checkov 2.4.48
Infracost v0.10.29
terraform-docs version v0.16.0 1f686b1 linux/arm64
terragrunt version v0.51.4
terrascan version: v1.18.3
TFLint version 0.48.0
+ ruleset.terraform (0.4.0-bundled)
tfsec v1.28.4
tfupdate 0.7.2
hcledit 0.2.10

## Checkov version is missing
TAG=v1.83.5 && docker run --entrypoint cat ghcr.io/antonbabenko/pre-commit-terraform:$TAG /usr/bin/tools_versions_info
pre-commit 3.4.0
Terraform v1.6.1
checkov
Infracost v0.10.29
terraform-docs version v0.16.0 1f686b1 linux/arm64
terragrunt version v0.52.1
terrascan version: v1.18.3
TFLint version 0.48.0
+ ruleset.terraform (0.4.0-bundled)
tfsec v1.28.4
tfupdate 0.8.0
hcledit 0.2.10

TAG=latest && docker run --entrypoint cat ghcr.io/antonbabenko/pre-commit-terraform:$TAG /usr/bin/tools_versions_info
pre-commit 3.6.2
Terraform v1.7.3
checkov
Infracost v0.10.33
terraform-docs version v0.17.0 795d369 linux/arm64
terragrunt version v0.55.3
terrascan version: v1.18.11
TFLint version 0.50.3
+ ruleset.terraform (0.5.0-bundled)
tfsec v1.28.5
trivy Version: 0.49.1
tfupdate 0.8.1
hcledit 0.2.10
```

* `.pre-commit-config.yaml`:

```bash
repos:
    rev: v1.86.1
    hooks:
      - id: terraform_checkov
        args:
          - --args=--quiet
```

</details>

### LINKED ISSUE #634: Checkov install silently fails during image build (and is also missing when pulling from ghcr.io)
### Describe the bug

On macos M2 Max (arm/aarch64)
Using docker image dhcr.io/atonbabenko/pre-commit-terraform:latest (since Oct 23, but up to and including a local build of main (v1.88.0).
When executing terraform_checkov during pre-commit the 'checkov command is not found'

### Cause

Rather than all the logs for diagnosis I can supply root cause.

pip install of checkov requires 0.13.0<rustworkx<0.14.0

Similar to the current limitation in dockerfile where cffi requires to build from source due to lack of musl wheel for aarch64 (update to this has been requested), rustworkx only had musl wheel for aarch64 added in 0.14.0, and current version is 0.14.1.

A request to checkov to bump rustworkx to at least 0.14.0 has been requested, although I'm not sure if that may have breaking changes or require further testing from them, so don't know the delivery time.

In the meantime adding/removing rust and cargo to the apk before checkov install allows rustworkx@0.13.2 to be compiled meeting the checkov dependency (albeit very slowly) until such time as they resolve it.  

I have raised a PR to this effect.

There is also some confusion in the error handling in the dockerfile, where because the pip install checkov fails in the first block it then tries to run the pip install checkov==latest in the or block.  Obviously happy path this isn't won't occur, but confuses the error-handling slightly when the first install fails.
### Environment information

* OS:  MacOS

antm-pp:

- [X] This PR fixes a bug.

Pending a new version of checkov (which has been requested), which bumps rustworkx >0.14.0.

This temporarily adds rust, cargo during checkov to allow rustworks@0.13.2 to compile, similar to how gcc etc are already added to compile cffi for similar reason (lack of musl aarch64).

Testing removal of all items after compile highlighted checkov exception due to missing gcc lib and therefore PR keeps gcc installed.  It's possible that only the specific lib is needed and could be much smaller.  In hindsight this was in the builder image, not in the final image, so not sure the reasoning behind trying to tidy up.

The impact of the change:

Installing packages takes about 55 seconds with or without rust/cargo so they're not adding significant time.
Compiling rustworkx takes about 330 seconds (only affecting aarch64 where not pre-compiled, and failing otherwise

The final image is inflated from ~980MB to ~1.68GB presumably by Checkov now being present.  Not sure how this compares to the image size on x86_64.

Fixes #634
Fixes #633 
https://github.com/bridgecrewio/checkov/issues/5608
https://github.com/Qiskit/rustworkx/issues/992#issuecomment-1746078224
https://github.com/Qiskit/rustworkx/pull/1008

### How can we test changes

I have built this docker image locally on MacOS M2 Max (aarch64) and am successfully able to get Checkov installed.  Nothing in these minor changes would be expected to impact the behaviour of a non-arm architecture build.

### DIFF
--- a/.github/.container-structure-test-config.yaml
+++ b/.github/.container-structure-test-config.yaml
@@ -15,6 +15,11 @@ commandTests:
     args: ["-version"]
     expectedOutput: ["^Terraform v([0-9]+\\.){2}[0-9]+\\non linux_amd64\\n$"]
 
+  - name: "gcc"
+    command: "gcc"
+    args: ["--version"]
+    expectedOutput: ["^gcc \(Alpine 12\..*$"]
+
   - name: "checkov"
     command: "checkov"
     args: ["--version"]
--- a/.github/.container-structure-test-config.yaml
+++ b/.github/.container-structure-test-config.yaml
@@ -18,7 +18,7 @@ commandTests:
   - name: "gcc"
     command: "gcc"
     args: ["--version"]
-    expectedOutput: ["^gcc \(Alpine 12\..*$"]
+    expectedOutput: ["^gcc \\(Alpine 12\\..*$"]
 
   - name: "checkov"
     command: "checkov"
--- a/.github/.container-structure-test-config.yaml
+++ b/.github/.container-structure-test-config.yaml
@@ -18,7 +18,7 @@ commandTests:
   - name: "gcc"
     command: "gcc"
     args: ["--version"]
-    expectedOutput: ["^gcc \\(Alpine 12\\..*$"]
+    expectedOutput: ["^gcc \\(Alpine 12\\."]
 
   - name: "checkov"
     command: "checkov"
--- a/.github/workflows/build-image-test.yaml
+++ b/.github/workflows/build-image-test.yaml
@@ -1,6 +1,6 @@
 name: "Build Dockerfile if changed and run smoke tests"
 
-on: push
+on: [pull_request]
 
 env:
   IMAGE_TAG: pr-test
--- a/.github/workflows/build-image-test.yaml
+++ b/.github/workflows/build-image-test.yaml
@@ -7,7 +7,12 @@ env:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        arch: [amd64, arm64]
+
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
         with:
@@ -61,7 +66,7 @@ jobs:
       # Can't build both platforms and use --load at the same time
       # https://github.com/docker/buildx/issues/59#issuecomment-1433097926
       - name: Build Multi-arch docker-image
-        if: steps.changed-files-specific.outputs.any_changed == 'true'
+        if: steps.changed-files-specific.outputs.any_changed == 'true' && matrix.os == 'ubuntu-latest' && matrix.arch == 'amd64'
         uses: docker/build-push-action@4a13e500e55cf31b7a5d59a38ab2040ab0f42f56 # v5.1.0
         with:
           context: .
--- a/.github/workflows/build-image-test.yaml
+++ b/.github/workflows/build-image-test.yaml
@@ -1,6 +1,6 @@
 name: "Build Dockerfile if changed and run smoke tests"
 
-on: [pull_request]
+on: push
 
 env:
   IMAGE_TAG: pr-test
--- a/Dockerfile
+++ b/Dockerfile
@@ -66,7 +66,9 @@ RUN if [ "$INSTALL_ALL" != "false" ]; then \
 RUN . /.env && \
     if [ "$CHECKOV_VERSION" != "false" ]; then \
     ( \
-        apk add --no-cache gcc=~12 libffi-dev=~3 musl-dev=~1 libgcc=~12 rust=~1 cargo=~1; \
+        # gcc rust cargo musl-dev required for compilation of rustworkx@0.13.2, no longer required once checkov version depends on rustworkx >0.14.0
+        # gcc libffi-dev musl-dev libffi-dev required for compilation of cffi, until it contains musl aarch64 
+        apk add --no-cache cargo=~1 gcc=~12 libffi-dev=~3 libgcc=~12 musl-dev=~1 rust=~1 ; \
         [ "$CHECKOV_VERSION" = "latest" ] && pip3 install --no-cache-dir checkov \
         || pip3 install --no-cache-dir checkov==${CHECKOV_VERSION}; \
         apk del gcc libffi-dev musl-dev rust cargo \
--- a/Dockerfile
+++ b/Dockerfile
@@ -72,8 +72,10 @@ RUN . /.env && \
         # gcc libffi-dev musl-dev required for compilation of cffi, until it contains musl aarch64
         export CARGO_NET_GIT_FETCH_WITH_CLI=true && \
         apk add --no-cache cargo=~1 gcc=~12 git=~2 libffi-dev=~3 libgcc=~12 musl-dev=~1 rust=~1 ; \
-        [ "$CHECKOV_VERSION" = "latest" ] && pip3 install --no-cache-dir checkov \
-        || pip3 install --no-cache-dir checkov==${CHECKOV_VERSION}; \
+        if [ "$CHECKOV_VERSION" = "latest" ]; \
+            then pip3 install --no-cache-dir checkov || exit 1; \
+            else pip3 install --no-cache-dir checkov==${CHECKOV_VERSION} || exit 1; \
+        fi; \
         apk del cargo gcc git libffi-dev musl-dev rust \
     ) \
     ; fi
--- a/Dockerfile
+++ b/Dockerfile
@@ -66,10 +66,10 @@ RUN if [ "$INSTALL_ALL" != "false" ]; then \
 RUN . /.env && \
     if [ "$CHECKOV_VERSION" != "false" ]; then \
     ( \
-        apk add --no-cache gcc=~12 libffi-dev=~3 musl-dev=~1; \
+        apk add --no-cache gcc=~12 libffi-dev=~3 musl-dev=~1 rust cargo; \
         [ "$CHECKOV_VERSION" = "latest" ] && pip3 install --no-cache-dir checkov \
         || pip3 install --no-cache-dir checkov==${CHECKOV_VERSION}; \
-        apk del gcc libffi-dev musl-dev \
+        apk del libffi-dev musl-dev rust cargo \
     ) \
     ; fi
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -66,10 +66,10 @@ RUN if [ "$INSTALL_ALL" != "false" ]; then \
 RUN . /.env && \
     if [ "$CHECKOV_VERSION" != "false" ]; then \
     ( \
-        apk add --no-cache gcc=~12 libffi-dev=~3 musl-dev=~1 rust cargo; \
+        apk add --no-cache gcc=~12 libffi-dev=~3 musl-dev=~1 libgcc rust cargo; \
         [ "$CHECKOV_VERSION" = "latest" ] && pip3 install --no-cache-dir checkov \
         || pip3 install --no-cache-dir checkov==${CHECKOV_VERSION}; \
-        apk del libffi-dev musl-dev rust cargo \
+        apk del gcc libffi-dev musl-dev rust cargo \
     ) \
     ; fi
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -66,12 +66,13 @@ RUN if [ "$INSTALL_ALL" != "false" ]; then \
 RUN . /.env && \
     if [ "$CHECKOV_VERSION" != "false" ]; then \
     ( \
-        # cargo, gcc, git, musl-dev, and rust required for compilation of rustworkx@0.13.2, no longer required once checkov version depends on rustworkx >0.14.0
+        # cargo, gcc, git, musl-dev, rust and CARGO envvar required for compilation of rustworkx@0.13.2, no longer required once checkov version depends on rustworkx >0.14.0
         # gcc libffi-dev musl-dev required for compilation of cffi, until it contains musl aarch64
-        apk add --no-cache cargo=~1 gcc=~12 libffi-dev=~3 libgcc=~12 musl-dev=~1 rust=~1 ; \
+        export CARGO_NET_GIT_FETCH_WITH_CLI=true && \
+        apk add --no-cache cargo=~1 gcc=~12 git=~2 libffi-dev=~3 libgcc=~12 musl-dev=~1 rust=~1 ; \
         [ "$CHECKOV_VERSION" = "latest" ] && pip3 install --no-cache-dir checkov \
         || pip3 install --no-cache-dir checkov==${CHECKOV_VERSION}; \
-        apk del cargo gcc libffi-dev musl-dev rust \
+        apk del cargo gcc git libffi-dev musl-dev rust \
     ) \
     ; fi
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -66,12 +66,12 @@ RUN if [ "$INSTALL_ALL" != "false" ]; then \
 RUN . /.env && \
     if [ "$CHECKOV_VERSION" != "false" ]; then \
     ( \
-        # gcc rust cargo musl-dev required for compilation of rustworkx@0.13.2, no longer required once checkov version depends on rustworkx >0.14.0
-        # gcc libffi-dev musl-dev libffi-dev required for compilation of cffi, until it contains musl aarch64 
+        # cargo, gcc, git, musl-dev, and rust required for compilation of rustworkx@0.13.2, no longer required once checkov version depends on rustworkx >0.14.0
+        # gcc libffi-dev musl-dev required for compilation of cffi, until it contains musl aarch64
         apk add --no-cache cargo=~1 gcc=~12 libffi-dev=~3 libgcc=~12 musl-dev=~1 rust=~1 ; \
         [ "$CHECKOV_VERSION" = "latest" ] && pip3 install --no-cache-dir checkov \
         || pip3 install --no-cache-dir checkov==${CHECKOV_VERSION}; \
-        apk del gcc libffi-dev musl-dev rust cargo \
+        apk del cargo gcc libffi-dev musl-dev rust \
     ) \
     ; fi
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -173,7 +173,6 @@ RUN . /.env && \
 RUN . /.env && \
     F=tools_versions_info && \
     pre-commit --version >> $F && \
-    echo "flock $(flock 2>&1 | head -n 1)" >> $F && \
     ./terraform --version | head -n 1 >> $F && \
     (if [ "$CHECKOV_VERSION"        != "false" ]; then echo "checkov $(checkov --version)" >> $F;     else echo "checkov SKIPPED" >> $F        ; fi) && \
     (if [ "$INFRACOST_VERSION"      != "false" ]; then echo "$(./infracost --version)" >> $F;         else echo "infracost SKIPPED" >> $F      ; fi) && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -69,6 +69,7 @@ RUN . /.env && \
     if [ "$CHECKOV_VERSION" != "false" ]; then \
     ( \
         # cargo, gcc, git, musl-dev, rust and CARGO envvar required for compilation of rustworkx@0.13.2, no longer required once checkov version depends on rustworkx >0.14.0
+        # https://github.com/bridgecrewio/checkov/pull/6045
         # gcc libffi-dev musl-dev required for compilation of cffi, until it contains musl aarch64
         export CARGO_NET_GIT_FETCH_WITH_CLI=true && \
         apk add --no-cache cargo=~1 gcc=~12 git=~2 libffi-dev=~3 libgcc=~12 musl-dev=~1 rust=~1 ; \
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,8 +17,10 @@ ARG PRE_COMMIT_VERSION=${PRE_COMMIT_VERSION:-latest}
 ARG TERRAFORM_VERSION=${TERRAFORM_VERSION:-latest}
 
 # Install pre-commit
-RUN [ ${PRE_COMMIT_VERSION} = "latest" ] && pip3 install --no-cache-dir pre-commit \
-    || pip3 install --no-cache-dir pre-commit==${PRE_COMMIT_VERSION}
+RUN if [ ${PRE_COMMIT_VERSION} = "latest" ]; \
+        then pip3 install --no-cache-dir pre-commit; \
+        else pip3 install --no-cache-dir pre-commit==${PRE_COMMIT_VERSION}; \
+    fi
 
 # Install terraform because pre-commit needs it
 RUN if [ "${TERRAFORM_VERSION}" = "latest" ]; then \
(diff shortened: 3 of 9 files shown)


yermulnik:
> This temporarily adds rust, cargo during checkov to allow rustworks@0.13.2 to compile, similar to how gcc etc are already added to compile cffi for similar reason (lack of musl aarch64).

@antm-pp Would it make sense 1) to leave corresponding comments in the Dockerfile for tracking and 2) to try and employ `TARGETARCH` (and `TARGETOS`?) to only add those to the platforms that indeed need them? 🤔

antm-pp:
@yermulnik  I am open to direction.  I'm not an expert in contributing, just found a solution and wanted to post it.

It looked like the cffi library was doing the same in the Dockerfile, so I adopted the same approach, without TARGETARCH. I can only comment for arm64 on Mac, don't have the facilities to test other variations, although I can see the pipeline builds multi-arch so could verify both paths if I update the PR I guess.

gcc is installed in the final image anyway, it's only the builder that installs then removes it.  Likewise the significant time imapct during build only applies if your arch doesn't have a pre-compiled rustworkx, so I don't think we're causing any issues in amd64 with the current approach.  As I said though, happy to take direction.

yermulnik:
Oh, I did misread the 2nd part of the «_The final image is inflated from ~980MB to ~1.68GB presumably by Checkov now being present_» sentence and thought the size change was related to rust/cargo being added 🤦🏻 Still is odd that Checkov almost doubles the size of the final image 🤔 
So, I'm fine with the changes (still need @MaxymVlasov to have his eyes on it though).

yermulnik (commented):
@antm-pp Please take a look into Dockerfile linter notice and see whether you can have container image built with Checkov if you follow what linter suggests. Thanks for the contribution.

yermulnik (approved):
I feel like I'm looking like a nerd now, though as of look of it the list of packages was alphabetically sorted before this change. @antm-pp Sorry I didn't pay attention to this before.

The PR looks good to me though. Hence approved, though awaiting for Max to review.

antm-pp:
@yermulnik 

Apologies, obviously the repo has pre-commit for itself, I didn't employ it locally first.  Now passing!  The instructions for contributing were quite hook heavy and not too much about PR process.

I noticed in the GHA multi-arch build actually the compile of rust failed due to a cargo issue.  This didn't happen to me locally. I'm just running a linux/arm64 build explicit now to see if that fails for me like it did in the pipeline.

Dockerfile
```diff
@@ -66,10 +66,10 @@ RUN if [ "$INSTALL_ALL" != "false" ]; then \
 RUN . /.env && \
     if [ "$CHECKOV_VERSION" != "false" ]; then \
     ( \
-        apk add --no-cache gcc=~12 libffi-dev=~3 musl-dev=~1; \
+        apk add --no-cache gcc=~12 libffi-dev=~3 musl-dev=~1 libgcc=~12 rust=~1 cargo=~1; \
```
MaxymVlasov, line 69: A few questions: Did you try to remove each of these dependencies, build an image, and confirm that is a minimal setup?

Please add check for GCC in https://github.com/antonbabenko/pre-commit-terraform/blob/master/.github/.container-structure-test-config.yaml

Also, I think that we need to find a way to run these tests on arm64 too.

antm-pp:
@MaxymVlasov re PR process : Still not sure what the right way is, only which bits I was told were wrong 🤣

MaxymVlasov:
>I presume forking and then PR back to here was right approach, as you don't allow local branch contribution.

@antm-pp  That's the default GitHub workflow. We can't provide Write access to everyone who has a GH account. That's technically impossible (at least, was). Branch creation requires write access, AFAIK. If I miss some new GH feature - please point me to docs - I can't able to quickly find such info. Also, there is a probability, that such a feature is available only for Enterprise customers or in private beta.

antm-pp:
@MaxymVlasov no problem. My lack of knowledge then, literally my first public contribution. Only worked in private repos before, wasn't aware of the limitation. Assumed wrongly that it was a setting rather than a system limitation.  Thanks for the correction.

Dockerfile
MaxymVlasov: >Happy to add a gcc check as requested, although I've not added the gcc dependency in the final image.

Ah, really? I see removal of rust and cargo, but not libgcc. How it works then :thinking: 

If there no package at the end - then there nothing to test

yermulnik: > > Happy to add a gcc check as requested, although I've not added the gcc dependency in the final image.
> 
> Ah, really? I see removal of rust and cargo, but not libgcc. How it works then 🤔

Apologies if I'm missing some bits of context: `gcc` != `libcc`. The former is the compiler collection and the latter is runtime libs only — https://pkgs.alpinelinux.org/package/edge/main/x86/gcc vs https://pkgs.alpinelinux.org/package/edge/main/x86/libgcc

yermulnik:
>     * I presume forking and then PR back to here was right approach, as you don't allow local branch contribution.

Yep, forking and PR'ing back is the common approach that allows arbitrary people to **not** have a **write** permission in the target repo (which allows a way too much). As Max already mentioned this is common approach with public repos. This is the contribution howto by GH: https://docs.github.com/en/get-started/exploring-projects-on-github/contributing-to-a-project

>     * Obviously I should have used the repos own pre-commit locally first.

You've hit a drawback of pre-commit (as a whole) framework: one can skip it =( So this is more of a **trust** to contributors to not bypass this soft requirement =)

ps: hope you're not suffering from all this stuff given it's your first public contribution (which — your contribution — is a way much more better than a dozens of others I've seen across different repos lately). thanks for your effort and time. we do appreciate this 👍🏻

Dockerfile
MaxymVlasov: Btw, we can't just add tests for macos, as Structure Test currently not support macos 
https://github.com/antonbabenko/pre-commit-terraform/pull/636

antm-pp:
So I am seeing some quite strange behaviour.

When I docker buildx (I'm running colima which runs ubuntu in qemu):

default (no target args, no --platform) or --platform linux/aarch64; then my original build (just adding rust, cargo) runs fine and completes locally.

However, with --platform linux/arm64 the build fails saying that during the cargo process for compiling it can't pull the crates.io index (it reports as a network/proxy issue)

In all 3 cases the container reports uname==aarch64, $TARGETOS==linux, $TARGETARCH==arm64

My limited understanding being that aarch64 is an alias of arm64 as a docker platform type.  The only guess I can make is that in some way colima or the docker process on my machine creates the container in a more 'native' environment as aarch64 but seems to think it needs a custom virtual host with some odd networking for building arm64.  Certainly the inside of the container seems to be consistent for both.

Whatever that issue is locally, was also present in the GHA/workflow build of linux/arm64.  Adding the git dependency and the ENV VAR to use cli git fetch for cargo seems to give a consistent success for me locally.  So will commit that to this PR to valid in the GHA workflow.

I've added a simple gcc check.  There isn't much detail in alpine's gcc --verison, but I've regexed on the 'gcc (Alpine 12.' it gives version verification to the major we've pinned, and is string locked to the version output (rather than only matching gcc which would also match a 'gcc not found' type stdout too.  I did look to test locally, but don't have a personally image repo I can send an image of that size too and recall with the container test action. I verified the regex with a grep -E inside a build container though.

Also added some notes around the package dependencies (sort of assumed the cffi ones from my read of that compilation.  This allows them to be alphabetical, but still clear which dependency is for which compilation.

Dockerfile
antm-pp: ## In BUILDER container
- So gcc was added and removed for the cffi requirement (pre-existed).
- gcc is also used/needed by the rustworkx requirement (plus rust,cargo)
- When the pre-existing gcc purge/removal happened it caused an exception in checkov (which had compiled correctly, but executes after gcc is removed in the 'builder' container to get its version number).  The exception was due to a missing ggc library (supplied by libgcc).
- Instead of keeping the whole of gcc (including the libgcc dependency), I decided to add libgcc seperately ratehr than only as an implicit dependency of gcc, so it was retained after the purge, so the version check worked.

## In RUN container / Final Image
However after all of that the important point to note is we were talking about the builder container, not the final image.  In the final image gcc is installed anyway as a dependency for pre-commit hooks, I haven't changed that.  So it is there, checkov does need it, and we can test against it, but nothing's changed for it in my PR.

I can add a comment [where gcc is installed in final image](https://github.com/antonbabenko/pre-commit-terraform/pull/635/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557L200) to say checkov is explicitly dependent upon it, rather than the current more general 'pre-commit hooks deps'

## MacOS

I don't think we're seeing MacOS issues here (although it is my OS and that of the other person who's raised the issue).  I think we're seeing only arm vs amd for inside the container (rather than the docker host).  Maybe there's just not many linux/arm users, but if you look at the multi-arch docker builds in actions they've been failing to compile rustworkx or install checkov since Oct23.  So it's not an issue limited to Mac or Darwin, it's impacting linux/arm64 too.

antm-pp:
After hours of re-running containers taking 15min at a time I think the whole aarch64 vs arm64 thing is nonsense (as it should be they're the same thing).  I think crates.io had an outage, and pulling from their github via the env var just changed the source.

Although they've reported no issues today, they had an extended outage with symptoms like I saw here (index unavailable) on Feb 14th. https://status.crates.io/

I'm now able to build this current PR successfully with no additions.  Could one of you rerun the existing docker build job action? [Run 2](https://github.com/antonbabenko/pre-commit-terraform/actions/runs/8024515032/job/21923526869) as I think it'll compile fine, and demonstrate the same issue has cleared inside the github runner with no further code change.

.github/.container-structure-test-config.yaml
```diff
@@ -18,7 +18,7 @@ commandTests:
   - name: "gcc"
     command: "gcc"
     args: ["--version"]
-    expectedOutput: ["^gcc \(Alpine 12\..*$"]
+    expectedOutput: ["^gcc \\(Alpine 12\\..*$"]
```
yermulnik, line 21: That's not essential though:
```suggestion
    expectedOutput: ["^gcc \\(Alpine 12\\."]
```

antm-pp:
🤦‍♂️. It'd be really handy if this pre-commit thing could have something like a 'hook' so it had to run before you could commit, not 5min later when you remember and the jobs have already been kicked off.  I can't believe I'm so out of my depth with such a minor change, and wasting your time kicking off jobs.

yermulnik:
> and wasting your time kicking off jobs.

No worries. It's the easiest I can do 😺

yermulnik:
> Appreciate the heavy involvement and responsiveness.

Appreciate the dedication 👍🏻

IvanIsCoding:
FYI if https://github.com/bridgecrewio/checkov/pull/6045 gets merged and released your failures will go away.

Dockerfile
MaxymVlasov: >I can add a comment [where gcc is installed in final image](https://github.com/antonbabenko/pre-commit-terraform/pull/635/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557L200) to say checkov is explicitly dependent upon it

Yes please, something like `# pre-commit-hooks & checkov deps`

antm-pp: So I think the issue is the structure of the '||' command for the install. When the rust compiler call fails it generates a false allowing the other part of the command to run (intended for managing checkov==latest vs checkov==<version>).  The 2nd part has its own failure mode that doesn't actually generated an exit1. So the error gets buried and the build passes.

MaxymVlasov: So all these combinations should be rewritten to vanilla if-then-else statements? @yermulnik 

https://github.com/antonbabenko/pre-commit-terraform/blob/c29bdb12779331c9fc9c3e61846d914db4219126/Dockerfile#L70-L71

yermulnik: Oh, yep, that sort of "ternary" in Bash is not vanilla `if/else` and has this kind of "discomfortable" hiccups 😿 
And, yes, these need to be re-written 😢 Either using vanilla `if/else`, or like below:
```shell
 [ "$CHECKOV_VERSION" = "latest" ] && pip3 install --no-cache-dir checkov; \ 
 [ "$CHECKOV_VERSION" != "latest" ] && pip3 install --no-cache-dir checkov==${CHECKOV_VERSION}; \
```

ps: I probably can try and do that, though I will need help building it and testing resulting images (@MaxymVlasov, that would be super great if you already had that automation so that I can push changes and you test build/run 🤪).

yermulnik: or like below 🤔 
```shell
 [ "$CHECKOV_VERSION" = "latest" ] && CHECKOV_VERSION="" || CHECKOV_VERSION="==${CHECKOV_VERSION}"; \ 
 pip3 install --no-cache-dir checkov${CHECKOV_VERSION}; \ 
```

.github/workflows/build-image-test.yaml
```diff
@@ -1,6 +1,6 @@
 name: "Build Dockerfile if changed and run smoke tests"
 
-on: [pull_request]
+on: push
```
MaxymVlasov, line 3:
```suggestion
on: [pull_request]
```

Dockerfile
```diff
@@ -66,10 +68,15 @@ RUN if [ "$INSTALL_ALL" != "false" ]; then \
 RUN . /.env && \
     if [ "$CHECKOV_VERSION" != "false" ]; then \
     ( \
-        apk add --no-cache gcc=~12 libffi-dev=~3 musl-dev=~1; \
-        [ "$CHECKOV_VERSION" = "latest" ] && pip3 install --no-cache-dir checkov \
-        || pip3 install --no-cache-dir checkov==${CHECKOV_VERSION}; \
-        apk del gcc libffi-dev musl-dev \
+        # cargo, gcc, git, musl-dev, rust and CARGO envvar required for compilation of rustworkx@0.13.2, no longer required once checkov version depends on rustworkx >0.14.0
```
MaxymVlasov, line 71:
```suggestion
        # cargo, gcc, git, musl-dev, rust and CARGO envvar required for compilation of rustworkx@0.13.2, no longer required once checkov version depends on rustworkx >0.14.0
        # https://github.com/bridgecrewio/checkov/pull/6045
```

Dockerfile
```diff
@@ -66,10 +68,16 @@ RUN if [ "$INSTALL_ALL" != "false" ]; then \
...
+        # https://github.com/bridgecrewio/checkov/pull/6045
+        # gcc libffi-dev musl-dev required for compilation of cffi, until it contains musl aarch64
+        export CARGO_NET_GIT_FETCH_WITH_CLI=true && \
+        apk add --no-cache cargo=~1 gcc=~12 git=~2 libffi-dev=~3 libgcc=~12 musl-dev=~1 rust=~1 ; \
+        if [ "$CHECKOV_VERSION" = "latest" ]; \
+            then pip3 install --no-cache-dir checkov || exit 1; \
+            else pip3 install --no-cache-dir checkov==${CHECKOV_VERSION} || exit 1; \
```
MaxymVlasov, line 78: `exit 1` prevents silent fail of `error: can't find Rust compiler`

Dockerfile
MaxymVlasov: @yermulnik done in this PR. Also, I can confirm that without @antm-pp changes to rust-cargo, it fails in arm64, when with - checks passes.

And now we have prevention of silent fails for checkov - https://github.com/antonbabenko/pre-commit-terraform/pull/635#pullrequestreview-1923825302

Dockerfile
MaxymVlasov: btw, we can't just add comments inside the code about it. Probably, the best place is on L73

yermulnik: Out of interest why at `# Install pre-commit` code block (lines 20-23) `pip3` has no `|| exit 1` bits? Is ii intentional?

yermulnik (approved):
Probs https://github.com/antonbabenko/pre-commit-terraform/pull/635/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R20-R23 needs `|| exit 1` bits too?

ps: approved just in case `|| exit 1` isn't needed there.

MaxymVlasov:
Because pre-commit don't uses rust?

I prefer somehow make structured tests working form arm64, than add `exit 1` in every possible place

Dockerfile
MaxymVlasov: Because pre-commit don't uses rust?

I prefer somehow make structured tests working form arm64, than add exit 1 in every possible place

yermulnik: > Because pre-commit doesn't use rust?

That's kind of weird: if `pip3 install ...` fails (as in returns non-zero code) then exit the shell with non-successful code (`exit 1`). And this is implemented for `checkov` install only, which means it's okay to not exit for the same e.g. with `pre-commit` install by `pip`. Does `pip` behave differently if it fails to install `pre-commit`? 🤔 Aint't Docker's `RUN` imply somewhat `set -e`? Should we try and add `set -e` to each `RUN` to explicitly let Docker `RUN` exit when any downstream command fails within `if/else` statement? 🤔 I'm a bit lost to be honest 😲 We definitely should use consistent solution for all similar expressions for consistency

MaxymVlasov: > When the rust compiler call fails it generates a false

exit code 0 and `false`

I dunno why, and who is responsible for such a great idea, but it's an edge case.

yermulnik: Apologies for... erm... odd suggestion, though should we wrap the shell logic of each `RUN` command into a separate shell scripts and run them instead of shell-scripting directly within `RUN` commands with all the fuss around escaping and trying to keep shell code readable with weird formatting quirks? 🤪
I do understand that this would add more layers, but... but it would ease maintenance of shell code =)
We can write up a single shell script that would handle different scenarios based on cmdline parameters (like `install_deps.sh pre-commit`, `install_deps.sh checkov`, so on) so that we `COPY` it only once.

Half past three a.m. — I might be generating odd ideas though 🤣

MaxymVlasov: >Does pip behave differently if it fails to install pre-commit

First of all, it not fails :)
second, if pip fails to install something, it usually returns exit code 1 => docker build fails

tbh, it's looks like a bug somewhere in pip or in it deps

yermulnik: > > When the rust compiler call fails it generates a false
> 
> exit code 0 and `false`

`false` in common terms is "return non-zero code".

> I dunno why, and who is responsible for such a great idea, but it's an edge case.

The only reasonable explanation for me is "Docker `RUN` runs shell w/o `set -e`" and that is why it doesn't fail when a program returns unsuccessfully and this is why `|| exit 1` makes it fail explicitly since `exit` command forces shell to exit.

MaxymVlasov: Regarding the installation script - why not, but it is out of the scope of this PR + not on my priority list

yermulnik: > > Does pip behave differently if it fails to install pre-commit
> 
> First of all, it not fails :)

If it does **not**, then you wouldn't be able to rely upon `||` (which is "if command om the left side returned non-successfully") or... or we definitely should **re**-RTFM =)))

> tbh, it's looks like a bug somewhere in pip or in it deps

or we're missing something very obvious 🤔

yermulnik: Prepend all `RUN` command with `set -e && ` and drop ` || exit 1` statements 🤪

MaxymVlasov: I'll rewrite it to bash scripts. Firstly I thought it will be a part of this PR, but now I see that better split them

MaxymVlasov (approved):

antonbabenko:
This PR is included in version 1.88.1 :tada:

Dockerfile
yermulnik: > I'll rewrite it to bash scripts.

What do you of a single script that takes args? Like `install_deps.sh pre-commit`, `install_deps.sh checkov`, `install_deps.sh pre-commit checkov ...`, or `install_deps.sh ALL`? Just to keep code in the same place and have re-usable snippets (like shell funcxtions) 🤔

MaxymVlasov: Problem there that they are different

MaxymVlasov: currently, I just moving it out and implementing logic as
```
COPY tools/install/ /install/

WORKDIR /bin_dir

RUN /install/pre-commit.sh
RUN /install/foo.sh
```

We can discuss it in next PR :)

yermulnik: > Problem there that they are different

Some are slightly different, whilst others use the same solutions (like extracting download URLs from GH releases API response). And also I see it odd to have a bunch of almost-the-same ten-lines shell scripts instead of one that can handle installation of all the deps one-by-one or all-at-once.
On the other hand such approach with single script would negatively impact build caching and each `RUN` layer would get rebuilt if the file is updated with a change to the installation steps of a specific dep =(
From this point if view I'd better stay with the current approach =)

> We can discuss it in next PR :)

Makes sense 🤝 (apologies that I already outlined my thoughts — this helps me imprint them in memory 😺)

webknjaz: @MaxymVlasov *do not* install distributions with pip separately because it will not take all things installed in the previous session into account when you run a new install command. Enumerate everything and let the dependency resolver know all your requirements together. Ideally, use pip-compile to produce and commit constraint files (lockfiles) and invoke it via `pip install -r direct-deps.txt -c constraint.txt`.

If you do run separate pip installs, inject a `pip check` invocation at the end to verify integrity.

PR #636, opened 2024-02-23
fix: Add tests for arm64 and for macOS

MaxymVlasov:

- [ ] This PR fixes a bug.

### DIFF
--- a/.github/workflows/build-image-test.yaml
+++ b/.github/workflows/build-image-test.yaml
@@ -7,7 +7,12 @@ env:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+        arch: [amd64, arm64]
+
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
         with:


.github/workflows/build-image-test.yaml
```diff
@@ -23,6 +28,13 @@ jobs:
             tools/entrypoint.sh
             .github/workflows/build-image-test.yaml
 
+      - name: Setup docker (missing on MacOS)
+        if: runner.os == 'macos'
+        run: |
+          brew install docker
+          colima start
```
MaxymVlasov, line 35: Take that from
https://github.com/actions/runner/issues/1456#issuecomment-1676495453

antm-pp:
You may understand this better than me (and appreciate it's a work in progress so not trying to review it).

The [macos-latest](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-private-repositories) github runner is an amd host, and the only arm host is macos-14 (beta)

The [ubuntu-latest](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#cloud-hosts-used-by-github-hosted-runners:~:text=repositories%2C%20GitHub%20uses-,Standard_DS2_v2,-virtual%20machines.%20For) (for private repo) is running in Azure according to [this[(https://learn.microsoft.com/en-us/azure/virtual-machines/dv2-dsv2-series#dsv2-series) spec, and indicates it's an Intel(amd) processor.

I think there's limited ability to run an arm host as a github runner.

In terms of outcomes, (despite the comments on the other PR I've made about seeing differences between aarch64/arm64) isn't the principle test just for a build with buildx arg --platform linux/arm64 as well as the current linux/amd64. Effectively the whole workflow in parallel rather than just the multi-arch build step?

MaxymVlasov:
arm64 exist for macOS 13. GH just not updated their docs
https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners

And matrix is a parallelism by itself, we build multiparty image just to test that this will work, during release

Check run workflows for the last commit
https://github.com/antonbabenko/pre-commit-terraform/actions/runs/8025437914

If there were no possibility to run arm64 runners, jobs would not start at all

MaxymVlasov:
I will take a look at the container structure GHA next week. if it can be quickly fixed, I will use my fork (as it was some time ago), and merge this PR with all 4 checks.

If not - well, 2 arches for Ubuntu are still better than just 1 Ubuntu amd64 build.

Then we will be able to test your PR right away in GHA, just in case.

.github/workflows/build-image-test.yaml
```diff
@@ -30,6 +30,14 @@ jobs:
           os: ubuntu-24.04-arm
           arch: arm64
 
+        - os-name: MacOS x64
+          os: macos-latest
+          arch: amd64
+
+         - os-name: MacOS ARM
```
MaxymVlasov, line 37:
```suggestion
       - os-name: MacOS ARM
```

.github/workflows/build-image-test.yaml
```diff
@@ -56,6 +64,12 @@ jobs:
         echo "IMAGE=ghcr.io/${GITHUB_REPOSITORY@L}:${{ env.IMAGE_TAG }}"
         >> $GITHUB_ENV
 
+    - name: Setup docker (missing on MacOS)
+        if: matrix.os == 'macos'
+        run: |
+          brew install docker
+          colima start
+
```
MaxymVlasov, line 72:
```suggestion
      if: matrix.os == 'macos'
      run: |
        brew install docker
        colima start

```

.github/workflows/build-image-test.yaml
```diff
@@ -56,6 +64,11 @@ jobs:
         echo "IMAGE=ghcr.io/${GITHUB_REPOSITORY@L}:${{ env.IMAGE_TAG }}"
         >> $GITHUB_ENV
 
+    - name: Setup docker (missing on MacOS)
+      if: matrix.os == 'macos'
```
MaxymVlasov, line 68:
```suggestion
      if: runner.os == 'macos'
```

.github/workflows/build-image-test.yaml
```diff
@@ -30,6 +30,14 @@ jobs:
           os: ubuntu-24.04-arm
           arch: arm64
 
+        - os-name: MacOS x64
+          os: macos-latest
+          arch: amd64
+
+        - os-name: MacOS ARM
+          os: macos-latest
+          arch: amd64
```
MaxymVlasov, line 39:
```suggestion
          arch: arm64
```

.github/workflows/build-image-test.yaml
```diff
@@ -53,9 +61,15 @@ jobs:
       # Lowercase the org/repo name to allow for workflow to run in forks,
       # which owners have uppercase letters in username
       run: >-
-        echo "IMAGE=ghcr.io/${GITHUB_REPOSITORY@L}:${{ env.IMAGE_TAG }}"
+        echo "IMAGE=ghcr.io/${{ github.repository | lower }}:${{ env.IMAGE_TAG
```
MaxymVlasov, line 64:
```suggestion
        repo=$(echo $GITHUB_REPOSITORY | tr '[:upper:]' '[:lower:]')
        echo "IMAGE=ghcr.io/${repo}:${{ env.IMAGE_TAG }}"
```

MaxymVlasov:
This works for mac x64.
Mac arm still need to figure out how to fix. Better do as separate PR

.github/workflows/build-image-test.yaml
MaxymVlasov, line 1: so, as Mac docker is still a linux VM which is not desceibed anyhow in metadata => I assume it have no difference with standard linux build, is it make any sense to test Mac-specific docker builds at all?





For context - all that started as issue from https://github.com/antonbabenko/pre-commit-terraform/pull/636. Linux arm tests whete added long time ago.
Mac x64 build works in this PR, Mac arm - still not, and according to build action used here for mac - it near impossible to make it run natively

@yermulnik

.github/workflows/build-image-test.yaml
```diff
@@ -52,9 +62,24 @@ jobs:
...
+    - name: Setup Docker on macOS
+      if: matrix.os-type == 'darwin'
+      uses: douglascamata/setup-docker-macos-action@v1.0.0
+
+    - name: Set DOCKER_HOST to point to Colima's socket
+      if: matrix.os-type == 'darwin'
+      run: |
+        docker context ls
+        # Colima usually creates the socket at ~/.colima/default/docker.sock
+        DOCKER_SOCKET_PATH=$(docker context ls | awk '{print $4}' | grep colima)
```
yermulnik, line 78: * Using `aws` as `cut` is an anti-pattern
* But if you change it to something like `... | awk -F$'\t' '$4 ~ /colima/ {print $3; exit}'`, the it makes sense (the `exit` is to produce only one string if there are more matches). Otherwise it should be something like `... | cut -f 4 | grep colima`.

.github/workflows/build-image-test.yaml
yermulnik: > I assume it have no difference with standard linux build, is it make any sense to test Mac-specific docker builds at all?

Might there be a difference around tool paths, libs, whatever the tests rely upon? Like those common diffs in POSIX (macOS) vs GNU (Linux)?
If not, then, given the info you provided, I agree that we can take macOS build as equal to Linux build.
Maybe call out for someone with macOS to validate tests provide the same outcome in Docker and on native Mac hardware? 🤔

MaxymVlasov: Different tools inside linux VMs? Could be, but then it depends what users use - Docker Desktop, Rancher etc. and specific it version.

I'll close it for now, till we will face a mac-specific issues that are not detected by current tests

yermulnik: It's probably that I didn't get the basic point of the message: did you mean that `os: macos-13 os-type: darwin` in reality is a linux ubuntu-based container? What make it macOS build then? Or did you mean that underlying host is Linux anyways and hence built artifacts (like pre-compiled binaries) should be the same anyways (kinda use the same set of instructions)? 🤔 I'm definitely of context of this PR and thus most probably am mixing up things 🤷🏻

MaxymVlasov: `os: macos-13 os-type: darwin` is Mac, but colima spin-ups Linux VM to run docker

>Colima means Containers on Lima. Since Lima is aka Linux Machines. By transitivity, Colima can also mean Containers on Linux Machines.

yermulnik: So your concern is underlying host system and not container internals, right?

PR #657, opened 2024-04-10
feat: Hook terraform_wrapper_module_for_each should use versions.tf from the module if it exists

nshenry03:

- [x] This PR introduces breaking change.
- [x] This PR fixes a bug.

Rather than having `hooks/terraform_wrapper_module_for_each.sh` always create `versions.tf` for us, if the module has `versions.tf`, use that instead.

We want to do this because if your module uses a non-Hashicorp provider and `versions.tf` doesn't have a the required `required_providers` config, then Terraform will try to download the provider from Hashicorp (i.e., for the [env0 provider](https://registry.terraform.io/providers/env0/env0/latest/docs), it would try to download `hashicorp/env0` which doesn't exist and NOT `env0/env0`)... We COULD set this in the deployment code; however, if you're using Terragrunt and you configure the provider with the `required_providers` config, then if you use a root module (with proper `versions.tf`) with the provider, then you get a conflict because you have `required_providers` configured in the module AND `required_providers` configured in the deployment code.

### How can we test changes

When you run pre-commit, `wrappers/versions.tf` should be copied from / the same as `/versions.tf`.

### DIFF
--- a/hooks/terraform_wrapper_module_for_each.sh
+++ b/hooks/terraform_wrapper_module_for_each.sh
@@ -393,7 +393,13 @@ EOF
       mv "$tmp_file_tf" "${output_dir}/main.tf"
 
       echo "$CONTENT_VARIABLES_TF" > "${output_dir}/variables.tf"
-      echo "$CONTENT_VERSIONS_TF" > "${output_dir}/versions.tf"
+
+      # If the root module has a versions.tf, use that; otherwise, create it
+      if [[ -f "${full_module_dir}/versions.tf" ]]; then
+        cp "${full_module_dir}/versions.tf" "${output_dir}/versions.tf"
+      else
+        echo "$CONTENT_VERSIONS_TF" > "${output_dir}/versions.tf"
+      fi
 
       echo "$CONTENT_OUTPUTS_TF" > "${output_dir}/outputs.tf"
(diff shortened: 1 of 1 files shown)


yermulnik (dismissed):

antonbabenko (commented):
I'm not sure that I agree or fully understand the reasoning for this PR. Let me think a bit and come back to it on Monday.

yermulnik:
> I'm not sure that I agree or fully understand the reasoning for this PR. Let me think a bit and come back to it on Monday.

The context from how I understood author's idea: it either pre-defined quite permissive constraint (see screenshot) or the user can provide it's own configuration.
image

While it looks more or less safe, I'd support @antonbabenko hesitaion as that user-provided file may contain something irrelevant and hence user can inadvertently shoot their leg.
So it's probably make make sense to allow this usage by an explicit argument passed to this hook, so that user sort of confirms they understand what they do.
The arg can rely on a pre-defined file name, or may require an explicit file name (the latter seems safer).

Just thoughts aloud.

ps: I'll take by approval back to prevent accidental merge to allow Anton to has his "veto" take actual effect.

antonbabenko:
This PR is included in version 1.89.0 :tada:

PR #670, opened 2024-05-04
feat: Support set custom TF/OpenTofu binary. | If you use a custom Docker image build, please note that `TERRAFORM_VERSION` now must be provided

### LINKED ISSUE #570: Support for OpenTofu
It would be nice if the hooks could support using opentf in addition to terraform. I'm thinking a simple change to using `_BIN` variables and setting those depending on what can be found in the `PATH`. I'd be willing to author a PR for that and help maintain that functionality in the future, if it is welcome :-)

den-is:

- [ ] This PR fixes a bug.

Allow setting custom TF binary.
- provide per hook setting `--hook-config=--tf-path=<path_to_binary_or_binary_name>`
- export ENV variable `PCT_TFPATH=<path_to_binary_or_binary_name>`
- export ENV variable `TERRAGRUNT_TFPATH=<path_to_binary_or_binary_name>`
- if none of above are set defaults to `terraform` binary found in the user's $PATH
- if none of above are set/available defaults to `tofu` binary found in the user's $PATH
- in the end, if none of above is set/available - `exit 1` with error message

Side feature. Fixes bad README.md

I'm open to any suggestion and contributions.

Fixes #570 
### How can we test changes

Create dummy repository with a dummy terraform module `test`

```sh
mkdir test
cd test
git init
```

Add files to `test/`:
**.pre-commit-config.yaml**
```yaml
repos:
- repo: local
  hooks:
  - id: terraform_fmt
    name: terraform_fmt
    entry: /path/to/pre-commit-terraform/hooks/terraform_fmt.sh
    language: script
    files: \.tf$
    # Uncomment these lines. Or set global ENV variables PCT_TFPATH or TERRAGRUNT_TFPATH
    #args:
    #- --hook-config=--tf-path=tofu

```
**main.tf**
```terraform
# dummy terraform file with ugly formatting
resource "null_resource" "print_time" {
triggers = {
current_time = timestamp()
}

provisioner "local-exec" {
command = "echo Current time is ${triggers.current_time}"
}
}
```

Run in the test module directory:
```sh
pre-commit install
pre-commit run --verbose --files main.tf
```


README.md
yermulnik, line 1: Please revert changes to bulleted lists in this file ("Table of contents" and "Install dependencies").

README.md
```diff
@@ -25,69 +25,81 @@ If you are using `pre-commit-terraform` already or want to support its developme
...
+- [`terrascan`](https://github.com/tenable/terrascan) required for `terrascan` hook
+- [`TFLint`](https://github.com/terraform-linters/tflint) required for `terraform_tflint` hook
+- [`TFSec`](https://github.com/liamg/tfsec) required for `terraform_tfsec` hook
+- [`Trivy`](https://github.com/aquasecurity/trivy) required for `terraform_trivy` hook
+- [`infracost`](https://github.com/infracost/infracost) required for `infracost_breakdown` hook
+- [`jq`](https://github.com/stedolan/jq) required for `terraform_validate` with `--retry-once-with-cleanup` flag, and for `infracost_breakdown` hook
+- [`tfupdate`](https://github.com/minamijoyo/tfupdate) required for `tfupdate` hook
+- [`hcledit`](https://github.com/minamijoyo/hcledit) required for `terraform_wrapper_module_for_each` hook
+
+#### 1.1 Custom terraform binaries and OpenTofu support
```
yermulnik, line 89:
```suggestion
#### 1.1 Custom Terraform binaries and OpenTofu support
```

README.md
```diff
@@ -25,69 +25,81 @@ If you are using `pre-commit-terraform` already or want to support its developme
...
+
+It is possible to set custom path to `terraform` binary.
+
+This makes possible to use [OpenTofu](https://opentofu.org) binary `tofu` instead of terraform.
```
yermulnik, line 93:
```suggestion
This makes it possible to use [OpenTofu](https://opentofu.org) binary `tofu` instead of `terraform`.
```

README.md
```diff
@@ -25,69 +25,81 @@ If you are using `pre-commit-terraform` already or want to support its developme
...
+
+How to:
+
+- For per hook configuration set `--hook-config=tf_binary=<path_to_binary_or_binary_name>`
+- export `PCT_TFPATH=<path_to_binary_or_binary_name>` environment variable
+- or maybe your already have `TERRAGRUNT_TFPATH=<path_to_binary_or_binary_name>` environment variable set
+- if none of above is set/available, hook defaults to `terraform` binary found in the user's $PATH
+- if none of above are set/available, hook try to find and use `tofu` binary in the user's $PATH
+- `--hook-config=tf_binary=xxx` takes precedence, then `PCT_TFPATH` env var, and so on
```
yermulnik, line 102: Please use `*` for bulleted list items instead of `-` for consistency.

README.md
```diff
@@ -25,69 +25,81 @@ If you are using `pre-commit-terraform` already or want to support its developme
...
+
+#### 1.1 Custom terraform binaries and OpenTofu support
+
+It is possible to set custom path to `terraform` binary.
+
+This makes possible to use [OpenTofu](https://opentofu.org) binary `tofu` instead of terraform.
+
+How to:
+
+- For per hook configuration set `--hook-config=tf_binary=<path_to_binary_or_binary_name>`
```
yermulnik, line 97: For the consistency sake, please use dash instead of underscore in hook config parameter name and prefix it with double dash. Also for the consistency's sake let's make parameter name unified/consistent.
```suggestion
* For per hook configuration set `--hook-config=--tf-path=<path_to_binary_or_binary_name>`
```

README.md
yermulnik, line 102:
```suggestion
* Export `PCT_TFPATH=<path_to_binary_or_binary_name>` environment variable
* Or maybe you already have `TERRAGRUNT_TFPATH=<path_to_binary_or_binary_name>` environment variable set
* If none of the above is set/available, hook defaults to `terraform` binary found in the user's `$PATH`
* Else hook tries to find and use `tofu` binary in the user's `$PATH`
```

I dropped last list item as it states obvious from my point of view.

README.md
```diff
@@ -103,7 +115,7 @@ All available tags [here](https://github.com/antonbabenko/pre-commit-terraform/p
 **Build from scratch**:
 
 > [!IMPORTANT]
-> To build image you need to have [`docker buildx`](https://docs.docker.com/build/install-buildx/) enabled as default builder.  
+> To build image you need to have [`docker buildx`](https://docs.docker.com/build/install-buildx/) enabled as default builder.
```
yermulnik, line 118: Please put two trailing spaces back — they denote carriage return in markdown.
Same applies to the similar changes across this file in the PR.

hooks/_common.sh
```diff
@@ -445,6 +445,46 @@ function common::colorify {
   echo -e "${COLOR}${TEXT}${RESET}" >&2
 }
 
+#######################################################################
+# Set terraform binary path
```
yermulnik, line 449:
```suggestion
# Set Terraform binary path
```

hooks/_common.sh
```diff
@@ -445,6 +445,46 @@ function common::colorify {
...
+# Allows user to set the path to custom Terraform or OpenTofu binaries
```
yermulnik, line 450:
```suggestion
# Allows user to set the path to custom Terraform or OpenTofu binary
```

hooks/_common.sh
```diff
@@ -445,6 +445,46 @@ function common::colorify {
...
+# Outputs:
+#   If failed - exit 1 with error message about missing Terraform/Opentofu binary
```
yermulnik, line 452:
```suggestion
#   If failed - exit 1 with error message about missing Terraform/OpenTofu binary
```

hooks/_common.sh
```diff
@@ -445,6 +445,46 @@ function common::colorify {
...
+function common::get_tf_binary {
+  local hook_config_tf_binary
```
yermulnik, line 454: For the sake of consistency:
```suggestion
function common::get_tf_path {
  local hook_config_tf_path
```
Please also change the rest accordingly.

hooks/_common.sh
```diff
@@ -445,6 +445,46 @@ function common::colorify {
...
+  for config in "${HOOK_CONFIG[@]}"; do
+    if [[ $config == tf_binary=* ]]; then
+      hook_config_tf_binary="${config#*=}"
+      break
+    fi
+  done
+
+  # direct hook config, has the highest priority
+  if [[ -n "${hook_config_tf_binary}" ]]; then
+    echo "${hook_config_tf_binary}"
```
yermulnik, line 465: - Curly brackets are redundant where there's no value expansion or substitution is performed 
- Double quotes inside double square brackets are redundant generally
- `-n` is implied within double square brackets
- From my point of view it makes sense to `return` immediately when the condition is true
```suggestion
  if [[ $hook_config_tf_binary ]]; then
    echo "$hook_config_tf_binary"
    return
```

This applies to all conditionals within this function. Please adjust.

yermulnik: I'm not entirely sure, though would it make sense to check whether target binary is executable and fail otherwise to fail early and to provide descriptive error to a user instead of relaying this to where the TF binary is executed actually? 🤔

hooks/_common.sh
```diff
@@ -445,6 +445,46 @@ function common::colorify {
...
+  local hook_config_tf_binary
+
+  for config in "${HOOK_CONFIG[@]}"; do
+    if [[ $config == tf_binary=* ]]; then
+      hook_config_tf_binary="${config#*=}"
+      break
+    fi
+  done
+
+  # direct hook config, has the highest priority
```
yermulnik, line 463: Not essential, though maybe "precedence" is a bit more correct in this context? 🤔 
```suggestion
  # direct hook config, has the highest precedence
```

hooks/_common.sh
```diff
@@ -445,6 +445,46 @@ function common::colorify {
...
+  # environment variable
+  elif [[ -n "${PCT_TFPATH}" ]]; then
+    echo "${PCT_TFPATH}"
+
+  # Maybe there is a similar setting for Terragrunt already
+  elif [[ -n "${TERRAGRUNT_TFPATH}" ]]; then
+    echo "${TERRAGRUNT_TFPATH}"
+
+  # check if Terraform binary is available
+  elif command -v terraform >/dev/null 2>&1; then
```
yermulnik, line 476: For the sake of brevity
```suggestion
  elif command -v terraform &>/dev/null; then
```

hooks/_common.sh
```diff
@@ -445,6 +445,46 @@ function common::colorify {
...
+    command -v terraform
+
+  # finally, check if Tofu binary is available
+  elif command -v tofu >/dev/null 2>&1; then
+    command -v tofu
+  else
+    echo "Neither Terraform nor Tofu binary could be found. Please set the TERRAGRUNT_TFPATH environment variable, set the tf_binary hook configuration, set the PRECOMMIT_TF_BINARY environment variable, or install terraform or tofu globally." >&2
```
yermulnik, line 483: - It is probably best to utilize `common::colorify` function to produce consistent output
- Also I'd suggest to improve wording
- And it is probably worth of providing resolution option in the order of their precedence
```suggestion
    common::colorify "red" "Neither Terraform nor OpenTofu binary could be found. Please either set the \"--tf-path\" hook configuration argument, or set the \"PCT_TFPATH\" environment variable, or set the \"TERRAGRUNT_TFPATH\" environment variable, or install Terraform or OpenTofu globally."
```

hooks/_common.sh
```diff
@@ -468,6 +508,8 @@ function common::terraform_init {
   local exit_code=0
   local init_output
 
+  TF_BINARY=$(common::get_tf_binary)
```
yermulnik, line 511: For the consistency's sake
```suggestion
  TF_PATH=$(common::get_tf_binary)
```

hooks/_common.sh
yermulnik: Also would it make sense to export `TF_PATH` var (it's what `TF_BINARY` is in your version) to replace var assignments in each hook that depends on it? What I mean is we could add call to this function into e.g. `common::initialize` to make this sort of global thing instead of adding this into each place where it is required. What do you think? 🤔

yermulnik (commented):
Thank you for the contribution and apologies for the delayed review. Hope my comments below make sense.
Also I'd like @MaxymVlasov to review it too, as he's the main dev in here.

ps: is there a reason you didn't update `terraform_providers_lock.sh` hook to utilize this feature? 🤔

den-is:
@yermulnik if you've missed it please join group conversation in UKR slack (you know which one)

yermulnik:
> @yermulnik if you've missed it please join group conversation in UKR slack (you know which one)

@den-is If this is about ukrops, then some time ago I was kicked from that Slack team w/o any decent reason by its notorious owner 🤷🏻

yermulnik:
@den-is When you're done updating PR and addressing comments, please re-request review, so that we all know it's ready for another round of review and discussion. Thanks.

yermulnik:
Also please pet `shfmt` (https://github.com/antonbabenko/pre-commit-terraform/actions/runs/9016122008/job/24780937313?pr=670) as it can't push to your repo/branch.
Not entirely sure, but maybe allowing maintainers to push to your branch, would allow GH workflow to push too 🤔 
image

den-is:
@yermulnik yes, ukrops... sad...
Just @MaxymVlasov is responding there

fixed shfmt recommendations

yermulnik:
> @den-is When you're done updating PR and addressing comments, please re-request review, so that we all know it's ready for another round of review and discussion. Thanks.

@den-is Just a heads up.

JFYI: Max is on vacation and thus this PR will be reviewed by him too when he's back.

README.md
```diff
@@ -66,28 +67,36 @@ If you are using `pre-commit-terraform` already or want to support its developme
...
-* [`infracost`](https://github.com/infracost/infracost) required for `infracost_breakdown` hook.
-* [`jq`](https://github.com/stedolan/jq) required for `terraform_validate` with `--retry-once-with-cleanup` flag, and for `infracost_breakdown` hook.
-* [`tfupdate`](https://github.com/minamijoyo/tfupdate) required for `tfupdate` hook.
-* [`hcledit`](https://github.com/minamijoyo/hcledit) required for `terraform_wrapper_module_for_each` hook.
+* [`pre-commit`](https://pre-commit.com/#install)
+* [`terraform`](https://www.terraform.io/downloads.html) and/or [`opentofu`](https://opentofu.org/docs/intro/install/)
+* [`git`](https://git-scm.com/downloads)
+* [BASH `3.2.57` or newer](https://www.gnu.org/software/bash/#download)
+* Internet connectivity (on first run)
+* x86_64 or arm64 compatible operation system
```
MaxymVlasov, line 75:
```suggestion
* [`pre-commit`](https://pre-commit.com/#install),
  <sub><sup>[`terraform`](https://www.terraform.io/downloads.html),
  <sub><sup>[`git`](https://git-scm.com/downloads),
  <sub><sup>[BASH `3.2.57` or newer](https://www.gnu.org/software/bash/#download),
  <sub><sup>Internet connection (on first run),
  <sub><sup>x86_64 or arm64 compatible operation system,
  <sub><sup>Some hardware where this OS will run,
  <sub><sup>Electricity for hardware and internet connection,
  <sub><sup>Some basic physical laws,
  <sub><sup>Hope that it all will work.
  </sup></sub></sup></sub></sup></sub></sup></sub></sup></sub></sup></sub></sup></sub></sup></sub></sup></sub><br><br>
```

MaxymVlasov:
Ah, it discared full readme, not just this commit, meh, 1 min

hooks/_common.sh
```diff
@@ -283,6 +283,9 @@ function common::per_dir_hook {
   # despite there's only one positional ARG left
   local -a -r files=("$@")
 
+  # shellcheck disable=SC2155
+  export TF_PATH=$(common::get_tf_path)
```
MaxymVlasov, line 287: Okay, I want to avoid Global vars whenever possible, as it basically an implicit dependency for function.

I'll quickly rewrite it to function args

den-is:
Pushed final README fixes. Missed the broken README part the last time when I was reverting back to `*/  `

hooks/_common.sh
```diff
@@ -445,13 +447,66 @@ function common::colorify {
...
+    return
+
+  # finally, check if Tofu binary is available
+  elif command -v tofu &> /dev/null; then
+    command -v tofu
+    return
+
+  else
+    common::colorify "red" "Neither Terraform nor OpenTofu binary could be found. Please either set the \"--tf-path\" hook configuration argument, or set the \"PCT_TFPATH\" environment variable, or set the \"TERRAGRUNT_TFPATH\" environment variable, or install Terraform or OpenTofu globally."
+    exit 1
```
MaxymVlasov, line 498:
```suggestion
    common::colorify "yellow" "Neither Terraform nor OpenTofu binary could be found. If hook crashed, please either set the '--tf-path' hook configuration argument, or set the 'PCT_TFPATH' environment variable, or set the 'TERRAGRUNT_TFPATH' environment variable, or install Terraform or OpenTofu globally."
```

We have hooks that don't need tf binaries, so it can be not error in some cases

hooks/_common.sh
```diff
@@ -445,13 +447,66 @@ function common::colorify {
...
+  done
+
+  # direct hook config, has the highest precedence
+  if [[ $hook_config_tf_path ]]; then
+    echo "${hook_config_tf_path}"
+    return
+
+  # environment variable
+  elif [[ $PCT_TFPATH ]]; then
+    echo "${PCT_TFPATH}"
```
yermulnik, line 478:
```suggestion
    echo "$PCT_TFPATH"
```

hooks/_common.sh
```diff
@@ -445,13 +447,66 @@ function common::colorify {
...
+    return
+
+  # Maybe there is a similar setting for Terragrunt already
+  elif [[ $TERRAGRUNT_TFPATH ]]; then
+    echo "${TERRAGRUNT_TFPATH}"
```
yermulnik, line 483:
```suggestion
    echo "$TERRAGRUNT_TFPATH"
```

hooks/_common.sh
```diff
@@ -445,13 +447,66 @@ function common::colorify {
...
+    return
```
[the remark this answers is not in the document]
yermulnik: Not sure I'm following the gist of the question, though probably (I'm purely guesstimating) this way it would be easier for people using both TF and TG hooks to specify custom binary only once using TG's method but covering also TF? 🤔 And from this point of view it makes sense to me.

@den-is What's your idea?

hooks/_common.sh
yermulnik: @MaxymVlasov Hope you don't mind that I replaced `crashed` with `fails` in your commit suggestion 🤞🏻 
I also replaced `globally` with `so it is available in the 'PATH' environment variable` to make it clear that binary is looked up from the `PATH`.

tools/install/opentofu.sh
```diff
@@ -0,0 +1,18 @@
...
+# shellcheck source=_common.sh
+. "$SCRIPT_DIR/_common.sh"
+
+#
+# Unique part
+#
+GH_ORG="opentofu"
+GH_RELEASE_REGEX_SPECIFIC_VERSION="https://.+?v${VERSION}_${TARGETOS}_${TARGETARCH}.zip"
+GH_RELEASE_REGEX_LATEST="https://.+?_${TARGETOS}_${TARGETARCH}.zip"
+DISTRIBUTED_AS="zip"
```
yermulnik, line 13: Out of curiosity: why not grab `tar.gz`?

Thoughts aloud: should we also grab and verify against signature/checksum files for where they are available (like GH release assets)? 🤔

tools/install/opentofu.sh
```diff
@@ -0,0 +1,18 @@
...
+
+common::install_from_gh_release "$GH_ORG" "$DISTRIBUTED_AS" \
+  "$GH_RELEASE_REGEX_LATEST" "$GH_RELEASE_REGEX_SPECIFIC_VERSION"
+
+rm CHANGELOG.md LICENSE README.md
```
yermulnik, line 18: 1. It's probably best to check if file exists before unlinking it (or force unlink) just in case they re-pack archive contents and add or remove some arbitrary files.
2. Starting the https://github.com/hashicorp/terraform/releases/tag/v1.8.2 release, TF also ships copy of their license file, so we may need to adjust `tools/install/terraform.sh` accordingly
3. For Linux distros we may provide the target filename to be extracted like this is implemented for tgz files (`unzip` on Linuxes, at least on some, supports that). While on other OSes we still need to clean up and keep an eye on archive contents (and maybe move this part to https://github.com/antonbabenko/pre-commit-terraform/blob/master/tools/install/_common.sh#L80-L82 ? 🤔)

MaxymVlasov: 2. Not a problem for TF, we unzip just binary
   https://github.com/antonbabenko/pre-commit-terraform/blob/a7697a2cf8c0b7df939e0eda38af3dc8ee882841/tools/install/terraform.sh#L19
1., 3. Yeah, make sense switch to tar.gz, as it already extracts only binary
  https://github.com/antonbabenko/pre-commit-terraform/blob/a7697a2cf8c0b7df939e0eda38af3dc8ee882841/tools/install/_common.sh#L70-L78

tools/install/opentofu.sh
MaxymVlasov: Mm, it could be a nice feature, but it 100% out-of-cope

tools/install/opentofu.sh
yermulnik: Oops, got distracted and forgot that I looked in TF install script and it was ok 🤦🏻 
And re the `unzip` bit I also missed this is about Docker and hence diffs between *NIXes doesn't matter 🤦🏻

Also given `unzip` in our Docker config can extract specific files, would it make sense to update the `zip` part in `tools/install/_common.sh` accordingly? 🤔

yermulnik: Probs in another PR though.

README.md
```diff
@@ -120,17 +135,19 @@ To install a specific version of individual tools, define it using `--build-arg`
 ```bash
 docker build -t pre-commit-terraform \
     --build-arg PRE_COMMIT_VERSION=latest \
+    --build-arg OPENTOFU_VERSION=latest \
+    --build-arg TERRAFORM_VERSION=1.5.7 \
     --build-arg TERRAFORM_VERSION=latest \
```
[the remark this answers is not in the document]
yermulnik: @MaxymVlasov Why TF 1.5.7 while Tofu is latest? 🤔

yermulnik (approved):
LGTM apart from https://github.com/antonbabenko/pre-commit-terraform/pull/670#discussion_r1610093420 that I didn't get why TF switched from `latest` to exact version.

README.md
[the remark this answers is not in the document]
MaxymVlasov: It is just an example, I don't think that someone will use it "as is". But answering your question:
TF 1.5.7 is the latest under the MPL2.0 license, and I suppose that "custom builders" will be here more from:
*  Tofu side
*  "we freeze TF on 1.5.7 for better times"

then from the latest TF users who just want to drop the Tofu binary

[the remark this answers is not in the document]
yermulnik: I got your point 👍🏻

yermulnik:
> Erm... this was already replied to by me yesterday.

@den-is My bad. I missed that comment (found it now). Thanks.

README.md
yermulnik: I personally have no issues with it using an editor of my preference.
Not sure where the `much more common public """standard""" for another markdown` comes from, though when I'm looking at https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#lists I see none out of three options to gain preference 🤷🏻 
If you think it's beneficial to switch from `*` to `-`, please create a separate PR for that so that we don't mix different scopes in this PR.
Thanks.

yermulnik:
> This was fixed by VSCode automatically.
> Will check for workaround

Add the below to config of VSCode of yours:
```json
    "[markdown]": {
        "files.trimTrailingWhitespace": false
    }
```

Other common useful settings VSCode usually is missing from I've found from experience of others using it:
```json
    "files.trimTrailingWhitespace": true,
    "files.insertFinalNewline": true,
    "files.trimFinalNewlines": true,
```

hooks/_common.sh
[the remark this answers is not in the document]
MaxymVlasov: Folks that use terragrunt should use `terragrunt` hooks, not `terraform` hooks - `terraform` will just not work in most cases. It's hard to imagine that someone uses both terragrunt and vanilla terraform, but if so - why do we think that folks would be happy with interference between "prj1 with TG and OpenTofu" and "prj2 with vanilla TF"?

And another question: do we need to specify `$tf_path` for terragrunt hooks?

Like `terragrunt validate --terragrunt-tfpath "$tf_path" "${args[@]}"`

https://github.com/antonbabenko/pre-commit-terraform/blob/1ac4f2d16814e0d4fed1175562161ff9a20bacea/hooks/terragrunt_validate.sh#L46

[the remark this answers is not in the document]
yermulnik: > Erm... this was already replied to by me yesterday.

@den-is My bad. I missed that comment (found it now). Thanks.

README.md
```diff
@@ -67,7 +68,7 @@ If you are using `pre-commit-terraform` already or want to support its developme
 ### 1. Install dependencies
 
 * [`pre-commit`](https://pre-commit.com/#install),
-  <sub><sup>[`terraform`](https://www.terraform.io/downloads.html),
```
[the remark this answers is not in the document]
MaxymVlasov: Nope, it looks exactly how it should look :)

This info need mostly for non-engineers, which came time to time and start asking exactly what wrote in by this tiny text

README.md
yermulnik: > This is oppresive and excess body movement.

Nothing personal, but it's first time I meet someone using VSCode (?) and being so frustrated by it pushing its very own "much more common public standards" 🤪 

I even tried to google it and found no documents from the 1st dozen of search results that would give a preference to any of unordered lists markers (`*`, `-`, and `+`).

den-is: My own preference is only dictated by common practice and every readme I have seen so far. which are not using `*`. If VSCode or vim fixes something using default linter settings I often prefer to accept it.

hooks/_common.sh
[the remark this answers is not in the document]
yermulnik: > It's hard to imagine that someone uses both terragrunt and vanilla terraform

I don't in particular, but a fellow squad in my project — does. And I'd imagine there are people in the wild who do the same and hence they may like using `TERRAGRUNT_TFPATH` to cover both use cases if they use `pre-commit-terraform`.

> And another question: do we need to specify `$tf_path` for terragrunt hooks?

That's a good question. We probably should. But only when it is provided via TG hook config. In the rest of use cases TG should rely upon its own logic, including `TERRAGRUNT_TFPATH` and others described at https://terragrunt.gruntwork.io/docs/reference/cli-options/#terragrunt-tfpath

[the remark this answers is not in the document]
MaxymVlasov: Had a conversation with @den-is 
TL;DR: we assume that it will no break anything in 99% of cases. And if such a thing happened - I'll mention Den is such an issue :D

MaxymVlasov (approved):

README.md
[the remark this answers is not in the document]
yermulnik: It makes some confusion for engineers as well. Most probably because they look into code and never into what it looks like when rendered =)
I even think we should add a tiny comment there (invisible when markdown is rendered into human-readable text) to clarify that this is way it should be 🤪

den-is:
> > This was fixed by VSCode automatically.
> > Will check for workaround
> 
> Add the below to config of VSCode of yours:
> 
> ```json
>     "[markdown]": {
>         "files.trimTrailingWhitespace": false
>     }
> ```
> 
> Other common useful settings VSCode usually is missing from what I've found from experience of others using it:
> 
> ```json
>     "files.trimTrailingWhitespace": true,
>     "files.insertFinalNewline": true,
>     "files.trimFinalNewlines": true,
> ```

lol dude... i always have this setting in all my editors... for years... you basically copy pasted my snippets which I add to different projects were noobs can't solve their whitespace issues.

Only thing I was not able to find correct settings for VSCode to fix `*` replacement with `-`. (Well I did not spend hours to find it in VSCode... just used vim to revert to your standard)

README.md
yermulnik: I'd suggest to stop pushing "common" when it's "personal". I have seen loads of READMEs using both (the `+` is what I indeed seen pretty much rare) and hence I have no preference unless Markdown language spec pushes it. Given VSCode (or whatever IDE "automates" this for you) and its devs are not those who maintain this language, I'd prefer you to create a separate PR for this change and we can discuss (as you're not the first one bringing this automated changes that are out of scope of their PR).

> I often prefer to accept it.

I respect your **personal** position, though it's not the best place to push it as eventually we'll find ourselves debating on OSX/iOS vs Linux/Android 🤪

hooks/_common.sh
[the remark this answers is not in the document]
yermulnik: Works for me 👍🏻 

> I'll mention Den is such an issue :D

I feel like `is` should be `in`, but with `is` the sentence sounds fabulous 🤣

MaxymVlasov:
Dpace trimming and so on should be done by [`.editorconfig`](https://editorconfig.org/) in the first place, as is supported by everything, including vim.

TOC is not supported by built-in MD engine, for this I use https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one
image

With 

```json
    "markdownlint.config": {
        "MD029": false,
        "MD012": {
            "maximum": 5
        }
    },
    "markdown.extension.toc.unorderedList.marker": "*",
    "markdown.extension.toc.levels": "2..6",
```

I'll check vanilla VS Code settings and then will send PR which will add needed configs and/or rewrite MD file

README.md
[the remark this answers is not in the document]
den-is: I meant I made it visible. Kinda what's bad in these requirements listed in normal text. Looked ok to me. It's known humanoids with different brain disabilities exist.
```
  * [`terraform`](https://www.terraform.io/downloads.html) or [`opentofu`](https://opentofu.org/docs/intro/install/),
  * [`git`](https://git-scm.com/downloads),
  * [BASH `3.2.57` or newer](https://www.gnu.org/software/bash/#download),
```
and just removed these
```
  <sub><sup>Some basic physical laws,
  <sub><sup>Hope that it all will work.
```

But yea. the question's closed. Looks good. Readme is not broken.

den-is:
In case if person uses VSCode, most of .editorconfig settings are supported natively. Otherwise you have to install another extension.

But yea, addined `.editorconfig` to you project makes it clear what formatting you require and ask to match.

yermulnik:
> Only thing I was not able to find correct settings for VSCode to fix `*` replacement with `-`. (Well I did not spend hours to find it in VSCode...

It's probably not the VSCode per se, but an extension. Like some of these: https://github.com/yzhang-gh/vscode-markdown or https://github.com/prettier/prettier-vscode

BTW FWIW: https://github.com/yzhang-gh/vscode-markdown/issues/145

> just used vim to revert to your standard)

It's not ours. It's Markdown language specification. Unless you prove otherwise.

PS: I looked at `markdownlint` rules and they seem to prefer `*` over `-` for unordered lists in documentation — https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md 🤔

antonbabenko (approved):

antonbabenko:
This PR is included in version 1.90.0 :tada:

PR #693, opened 2024-07-31
fix(`terraform_docs`): Suppress "terraform command not found" error message in case binary does not exist

ponceps:

- [X] This PR fixes a bug.

If you are using OpenTofu and don't have Terraform binary installed, terraform_docs hook will display a `command not found` error. The error line checks if Terraform release is 0.12, so it's possible suppress this error message if Terraform binary doesn't exist.

image

### How can we test changes

Below are the steps to reproduce the error.

1) Install OpenTofu
2) Uninstall Terraform (or remove it from PATH)
3) Make any changes to variables.tf or outputs.tf
4) Run `pre-commit -a`

image

### DIFF
--- a/hooks/terraform_docs.sh
+++ b/hooks/terraform_docs.sh
@@ -47,7 +47,7 @@ function terraform_docs_ {
   IFS=";" read -r -a configs <<< "$hook_config"
 
   local hack_terraform_docs
-  hack_terraform_docs=$(terraform version | sed -n 1p | grep -c 0.12) || true
+  hack_terraform_docs=$(terraform version 2> /dev/null | sed -n 1p | grep -c 0.12) || true
 
   if [[ ! $(command -v terraform-docs) ]]; then
     echo "ERROR: terraform-docs is required by terraform_docs pre-commit hook but is not installed or in the system's PATH."
--- a/hooks/terraform_docs.sh
+++ b/hooks/terraform_docs.sh
@@ -47,7 +47,10 @@ function terraform_docs_ {
   IFS=";" read -r -a configs <<< "$hook_config"
 
   local hack_terraform_docs
-  hack_terraform_docs=$(terraform version 2> /dev/null | sed -n 1p | grep -c 0.12 || tofu version 2> /dev/null | echo 0) || true
+  local hack_terraform_docs=0
+  if [[ $(\command -V terraform 2> /dev/null) ]]; then
+    hack_terraform_docs=$(terraform version | sed -n 1p | grep -c 0.12) || true
+  fi
 
   if [[ ! $(command -v terraform-docs) ]]; then
     echo "ERROR: terraform-docs is required by terraform_docs pre-commit hook but is not installed or in the system's PATH."
--- a/hooks/terraform_docs.sh
+++ b/hooks/terraform_docs.sh
@@ -47,7 +47,7 @@ function terraform_docs_ {
   IFS=";" read -r -a configs <<< "$hook_config"
 
   local hack_terraform_docs
-  hack_terraform_docs=$(terraform version 2> /dev/null | sed -n 1p | grep -c 0.12) || $(tofu version 2> /dev/null | echo 0) || true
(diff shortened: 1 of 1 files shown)


yermulnik:
@MaxymVlasov Are you ok with the change?

MaxymVlasov (approved):
Looks legit

MaxymVlasov:
just lmc that it will not fail silently

hooks/terraform_docs.sh
```diff
@@ -47,7 +47,7 @@ function terraform_docs_ {
   IFS=";" read -r -a configs <<< "$hook_config"
 
   local hack_terraform_docs
-  hack_terraform_docs=$(terraform version | sed -n 1p | grep -c 0.12) || true
+  hack_terraform_docs=$(terraform version 2> /dev/null | sed -n 1p | grep -c 0.12) || true
```
MaxymVlasov, line 50:
```suggestion
  hack_terraform_docs=$(terraform version 2> /dev/null | sed -n 1p | grep -c 0.12) || $(t tofu version 2> /dev/null |  echo 0)  || true
```

yermulnik:
> just lmc that it will not fail silently

What's `lmc`? 🤔 
If it fails, then we suppress failure (`|| true`) and var is left empty.
The change simply suppresses error output.

ps: it might make sense to wrap the `hack_terraform_docs=$(terraform version 2> /dev/null | sed -n 1p | grep -c 0.12) || true` command in conditional like this:
```bash
if [[ $(\command -V terraform 2> /dev/null) ]]; then
  hack_terraform_docs=$(terraform version | sed -n 1p | grep -c 0.12) || true
else
  hack_terraform_docs=0
fi
```
WDTY?

hooks/terraform_docs.sh
```diff
@@ -47,7 +47,7 @@ function terraform_docs_ {
   IFS=";" read -r -a configs <<< "$hook_config"
 
   local hack_terraform_docs
-  hack_terraform_docs=$(terraform version | sed -n 1p | grep -c 0.12) || true
+  hack_terraform_docs=$(terraform version 2> /dev/null | sed -n 1p | grep -c 0.12) || $(tofu version 2> /dev/null | echo 0) || true
```
MaxymVlasov, line 50:
```suggestion
  hack_terraform_docs=$(terraform version 2> /dev/null | sed -n 1p | grep -c 0.12 || $(tofu version 2> /dev/null | echo 0)) || true
```

hooks/terraform_docs.sh
```diff
@@ -47,7 +47,7 @@ function terraform_docs_ {
   IFS=";" read -r -a configs <<< "$hook_config"
 
   local hack_terraform_docs
-  hack_terraform_docs=$(terraform version | sed -n 1p | grep -c 0.12) || true
+  hack_terraform_docs=$(terraform version 2> /dev/null | sed -n 1p | grep -c 0.12 || tofu version 2> /dev/null | echo 0) || true
```
MaxymVlasov, line 50:
```suggestion
  local hack_terraform_docs=0
  if [[ $(\command -V terraform 2> /dev/null) ]]; then
    hack_terraform_docs=$(terraform version | sed -n 1p | grep -c 0.12) || true
  fi
```

hooks/terraform_docs.sh
yermulnik: 1. I thought we only need to check whether TF is `0.12` (`1`) or else (`0`). OpenTofu is unrelated IMHO.
2. The `| echo 0` is why? 😕

hooks/terraform_docs.sh
MaxymVlasov: @yermulnik Yeah, that is the best possible variant, tnx

hooks/terraform_docs.sh
yermulnik: Ah, nevermind me. I missed that you already pushed the change.

yermulnik (approved):

antonbabenko:
This PR is included in version 1.92.1 :tada:

Issue #382, opened 2022-05-11
`terraform_docs`: Check abillity to use native `terraform-docs` config as replacement for custom `--hook-config=`


MaxymVlasov:
In #380 I discovered that `terraform-docs` now have a big config, that, theoretically, can be used as a replacement for all or part of custom `--hook-config=`'s 

Need to investigate that option, hope we can remove part of the custom logic.

yermulnik:
> In #380 I discovered that `terraform-docs` now have a big config

Me too 😂 

> that, theoretically, can be used as a replacement for all or part of custom `--hook-config=`'s

...which would mean we need to ship `pre-commit-terraform` along with a `terraform-docs` config file, which consumers of `pre-commit-terraform` would need to amend per their preferences and requirements and put into a standard location where `terraform-docs` expect it to be (the precedence of config vs cmdline options is not obvious from the `terraform-docs` README file).
So this may suffice for dockerized `pre-commit-terraform` only, I seem to think (I might be wrong).
ON the other hand, `terraform-docs` README says to use cmdline options when used with `pre-commit`: https://github.com/terraform-docs/terraform-docs#pre-commit-hook

drAlberT:
I'm using the following hook configuration and everything related to terraform-doc is managed by the config files. It works like a charm

 ```
- repo: https://github.com/antonbabenko/pre-commit-terraform
    rev: v1.72.1
    hooks:
      - id: terraform_docs
        args:
          - --args=--config=infra/.terraform-docs.yml
          - --hook-config=--create-file-if-not-exist=true
```

**Comment by UkklyDukkling** [comment]
Adding my experience here using the `.terraform-docs.yml` config file with this hook. Generally passing in the config file works, however there is a slight bug that can produce some undesirable effects.

## Example setup for various files:

Config file `.pre-commit-config.yaml`
```yml
repos:
  - repo: https://github.com/antonbabenko/pre-commit-terraform
    rev: v1.81.0
    hooks:
      - id: terraform_docs
        args:
          - --args=--config=.terraform-docs.yml
```

Config file `.terraform-docs.yml`
```yml
formatter: "markdown"
version: "0.16.0"

content: |-
  {{ .Requirements }}
  ...
  various other customizations for our documentation
  ...

output:
  file: README.md
  mode: inject
  template: |-
    <!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->
    {{ .Content }}
[...]

PR #698, opened 2024-08-16
fix(`terraform_docs`): Fix issue and prioritize `output.file` setting from `.terraform-docs.yml` config over `--hook-config=--path-to-file=`

### LINKED ISSUE #697: `terraform_docs` generated output incorrect when `output.file` is configured
### Describe the bug

When `output.file` is specified, `terraform-docs` directly updates this file and prints the message: `README.md updated successfully` (or similar) to stdout. Because the `terraform_docs` hook always redirects stdout to the file, we end up with the below:

```md
<!-- BEGIN_TF_DOCS -->
README.md updated successfully
<!-- END_TF_DOCS -->
```

### How can we reproduce it?

in `.terraform-docs.yml`:

```yaml
output:
  file: "./README.md"
  # ... other settings as required
```

in `.pre-commit-config.yaml`:
```yaml
repos:
    rev: v1.92.1
    hooks:
      - id: terraform_docs
        args:
          - --hook-config=--path-to-file=README.md
          - --hook-config=--create-file-if-not-exist=true
          - --hook-config=--use-standard-markers=true
```

### Environment information

* OS:  Windows 10

* `uname -a` and/or `systeminfo | Select-String "^OS"` output:

```bash
OS Name:                   Microsoft Windows 10 Enterprise
OS Version:                10.0.19045 N/A Build 19045
OS Manufacturer:           Microsoft Corporation
OS Configuration:          Member Workstation
OS Build Type:             Multiprocessor Free
```

```bash
GNU bash, version 5.2.26(1)-release (x86_64-pc-msys)
pre-commit 3.8.0
/usr/bin/bash: line 3: tofu: command not found
Terraform v1.5.5
Python 3.10.11
Python 3.10.11
checkov 3.2.223
infracost SKIPPED
terraform-docs version v0.18.0 228c7a7 windows/amd64
terragrunt SKIPPED
terrascan SKIPPED
TFLint version 0.52.0
+ ruleset.terraform (0.9.1)
+ ruleset.aws (0.32.0)
tfsec v1.28.10
trivy Version: 0.54.1
Check Bundle:
  Digest: sha256:ef2d9ad4fce0f933b20a662004d7e55bf200987c180e7f2cd531af631f408bb3
  DownloadedAt: 2024-08-15 02:35:34.5007199 +0000 UTC
tfupdate SKIPPED
hcledit SKIPPED
```

* `.pre-commit-config.yaml`:

```yaml
# See http://pre-commit.com for more information
# See http://pre-commit.com/hooks.html for more hooks

repos:
    rev: v1.92.1
    hooks:
      - id: terraform_docs
        args:
          - --hook-config=--path-to-file=README.md
          - --hook-config=--create-file-if-not-exist=true
          - --hook-config=--use-standard-markers=true
```

</details>

sworisbreathing:

- [x] This PR fixes a bug.

This PR explicitly overrides any output files configured in `.terraform-docs.yml` to ensure the rendered documentation is written to stdout (which the hook pipes to a file and then injects into the appropriate destination file).

Fixes #697

### How can we test changes

setup as per the steps to reproduce in #697 and run the hook.

### DIFF
--- a/hooks/terraform_docs.sh
+++ b/hooks/terraform_docs.sh
@@ -168,10 +168,12 @@ function terraform_docs {
     # Prioritize `.terraform-docs.yml` `output.file` over
     # `--hook-config=--path-to-file=` if it set
     local output_file
-    output_file=$(grep -A1000 -e '^output:$' "$config_file" | grep ' file:' |
-      awk -F':' '{print $2}' | tr -d '[:space:]"' | tr -d "'")
+    # Get latest non-commented `output.file` from `.terraform-docs.yml`
+    output_file=$(grep -A1000 -e '^output:$' "$config_file" | grep ' file:' | grep -v '#' || true)
+
     if [ "$output_file" ]; then
-      text_file=$output_file
+      # Extract filename from `output.file` line
+      text_file=$(echo "$output_file" | awk -F':' '{print $2}' | tr -d '[:space:]"' | tr -d "'")
     fi
 
     # Suppress terraform_docs color
--- a/hooks/terraform_docs.sh
+++ b/hooks/terraform_docs.sh
@@ -121,6 +121,7 @@ function terraform_docs {
   # Get hook settings
   #
   local text_file="README.md"
+  local use_path_to_file=false
   local add_to_existing=false
   local create_if_not_exist=false
   local use_standard_markers=false
@@ -136,6 +137,7 @@ function terraform_docs {
     case $key in
       --path-to-file)
         text_file=$value
+        use_path_to_file=true
         ;;
       --add-to-existing-file)
         add_to_existing=$value
@@ -169,11 +171,16 @@ function terraform_docs {
     # `--hook-config=--path-to-file=` if it set
     local output_file
     # Get latest non-commented `output.file` from `.terraform-docs.yml`
-    output_file=$(grep -A1000 -e '^output:$' "$config_file" | grep ' file:' | grep -v '#' | tail -n 1 || true)
+    output_file=$(grep -A1000 -e '^output:$' "$config_file" | grep -E '^[[:space:]]+file:' | tail -n 1) || true
 
-    if [ "$output_file" ]; then
+    if [[ $output_file ]]; then
       # Extract filename from `output.file` line
       text_file=$(echo "$output_file" | awk -F':' '{print $2}' | tr -d '[:space:]"' | tr -d "'")
+
+      if [[ $use_path_to_file ]]; then
+        common::colorify "yellow" "NOTE: You set both '--hook-config=--path-to-file=' and 'output.file' in '$config_file'"
+        common::colorify "yellow" "      'output.file' from '$config_file' will be used."
+      fi
     fi
 
     # Suppress terraform_docs color
--- a/hooks/terraform_docs.sh
+++ b/hooks/terraform_docs.sh
@@ -159,13 +159,22 @@ function terraform_docs {
   if [[ "$args" != *"--config"* ]]; then
     local tf_docs_formatter="md"
 
-  # Suppress terraform_docs color
   else
 
     local config_file=${args#*--config}
     config_file=${config_file#*=}
     config_file=${config_file% *}
 
+    # Prioritize `.terraform-docs.yml` `output.file` over
+    # `--hook-config=--path-to-file=` if it set
+    local output_file
+    output_file=$(grep -A1000 -e '^output:$' "$config_file" | grep ' file:' |
+      awk -F':' '{print $2}' | tr -d '[:space:]"' | tr -d "'")
+    if [ "$output_file" ]; then
+      text_file=$output_file
+    fi
+
+    # Suppress terraform_docs color
     local config_file_no_color
     config_file_no_color="$config_file$(date +%s).yml"
 
--- a/hooks/terraform_docs.sh
+++ b/hooks/terraform_docs.sh
@@ -228,7 +228,7 @@ function terraform_docs {
 
     if [[ "$terraform_docs_awk_file" == "0" ]]; then
       # shellcheck disable=SC2086
-      terraform-docs $tf_docs_formatter $args ./ > "$tmp_file"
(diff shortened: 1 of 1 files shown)


MaxymVlasov:
This actually ignore any setting set in `.terraform-docs.yml` `output.file` 

Maybe that's a good idea to prioritize `.terraform-docs.yml` `output.file`  over  `- --hook-config=--path-to-file=` as it make step to resolve https://github.com/antonbabenko/pre-commit-terraform/issues/382

hooks/terraform_docs.sh
```diff
@@ -159,13 +159,24 @@ function terraform_docs {
...
+    # Prioritize `.terraform-docs.yml` `output.file` over
+    # `--hook-config=--path-to-file=` if it set
+    local output_file
+    # Get latest non-commented `output.file` from `.terraform-docs.yml`
+    output_file=$(grep -A1000 -e '^output:$' "$config_file" | grep ' file:' | grep -v '#' | tail -n 1 || true)
+
+    if [ "$output_file" ]; then
+      # Extract filename from `output.file` line
+      text_file=$(echo "$output_file" | awk -F':' '{print $2}' | tr -d '[:space:]"' | tr -d "'")
+    fi
```
MaxymVlasov, line 177: Done it this way to avoid introduction of new dependencies (`yq`)

hooks/terraform_docs.sh
```diff
@@ -159,13 +159,24 @@ function terraform_docs {
...
 
     local config_file=${args#*--config}
     config_file=${config_file#*=}
     config_file=${config_file% *}
 
+    # Prioritize `.terraform-docs.yml` `output.file` over
+    # `--hook-config=--path-to-file=` if it set
+    local output_file
+    # Get latest non-commented `output.file` from `.terraform-docs.yml`
+    output_file=$(grep -A1000 -e '^output:$' "$config_file" | grep ' file:' | grep -v '#' | tail -n 1 || true)
```
yermulnik, line 172: * I think `|| true` should be moved out of parenthesis to denote it is to suppress `errexit`
* Not sure whether `grep` on macOS is fully compatible with "regular" `grep` (can someone test please?), though if yes, then we can optimize cmdline
  * Also added `-m 1` to print 1st result and exit (not sure whether we indeed want to use **last** result rather than **first** — I might be wrong, though YAML doesn't allow duplicate keys I guess)
```suggestion
    output_file=$(grep -A1000 -e '^output:$' "$config_file" | grep -E -m 1 '^[[:space:]]+file:') || true
```

hooks/terraform_docs.sh
```diff
@@ -159,13 +159,24 @@ function terraform_docs {
...
+
+    if [ "$output_file" ]; then
```
yermulnik, line 174: Bashism for consistency with other conditionals
```suggestion
    if [[ $output_file ]]; then
```

hooks/terraform_docs.sh
```diff
@@ -159,13 +159,24 @@ function terraform_docs {
...
 
-  # Suppress terraform_docs color
   else
 
     local config_file=${args#*--config}
     config_file=${config_file#*=}
     config_file=${config_file% *}
 
+    # Prioritize `.terraform-docs.yml` `output.file` over
+    # `--hook-config=--path-to-file=` if it set
```
yermulnik, line 169: I can't seem to understand the reasoning: ain't it better raise error when both options are used instead of silently ignoring hook config's one? 🤔

yermulnik: What I mean is that if a user provides this option to hook, then the user expects it to use this option. When there's a race condition, we should raise error or indicate incorrect setting by other mean(s) (e.g. by printing warning message that the one from YAML is used and the hook config's one is ignored).
I'm ok with either approach when this is clearly brought to user's attention and knowledge.

hooks/terraform_docs.sh
MaxymVlasov: YAML allow duplicate keys, and use last one

hooks/terraform_docs.sh
MaxymVlasov: Reasoning was described in https://github.com/antonbabenko/pre-commit-terraform/issues/382
I'll add warning msg

yermulnik: Took a read and... and didn't get it =( I'd appreciate "tl;dr" =)
It's most common that command line parameters override config file params. Which is sort of comparable with `--hook-config`.
Either way I'm relying on your opinion in this case 👍🏻

yermulnik (approved):
@MaxymVlasov LGTM

hooks/terraform_docs.sh
MaxymVlasov: I love removing code. Much more than write new one :D 
Because you don't need to maintain code which not exist.

At the end, I'd like to replace as much hand-made settings as possible by 3rd-party, in this case - by terraform-docs settings. So, at the end, I'd like to deprecate and remove every `--hook-config` which can be replaced by `terraform-docs` settings, as long as it will not violate main idea of how pre-commit hooks works

MaxymVlasov: So keeping this in mind, we can clearly see that native terraform-docs settings should have precedence over `--hook-config` settings, as long as they not break hook execution - in last case we try explicitly disable such settings (or at least mention in docs that they shouldn't be used)

MaxymVlasov: IE
image

MaxymVlasov (approved):

antonbabenko:
This PR is included in version 1.92.2 :tada:

hooks/terraform_docs.sh
yermulnik: Aha, now I think I got the point: hook config's parameter is a workaround from when tf-docs wasn't mature enough — this makes sense to me now 👍🏻 Thanks.

sworisbreathing:
Wow thanks for fixing up the PR and the quick review/merge @yermulnik/@antonbabenko. I've just upgraded to the new release and it's working great!

Issue #699, opened 2024-08-23
`terraform_docs` hook broken in `v1.92.2`


Shr3ps:
### Describe the bug

Using the latest version of the hook (v1.92.2), pre-commit is always in "PASSED" state even if we change variables, outputs or versions constrains.

### How can we reproduce it?

```bash
vim versions.tf
sed -i 's/version = "~> 3.80"/version = "~> 3.86"/g' versions.tf
pre-commit run -a

Terraform fmt...........................................................................Passed
Terraform docs..........................................................................Passed
```

It should failed and update README.md file.

### Environment information

* OS:  

OS: Win11 with Ubuntu 20.04 on WSL2
OS: Ubuntu 20.04

* Docker image tag/git commit:  v1.92.2

* `.pre-commit-config.yaml`:

<details><summary>file content</summary>

```yaml
  25   │   - repo: https://github.com/antonbabenko/pre-commit-terraform.git
  26   │     rev: v1.92.2
  27   │     hooks:
  28   │       - id: terraform_fmt
  29   │         stages: [pre-commit]
  30   │       - id: terraform_docs
  31   │         stages: [pre-commit]
  32   │         args:
  33   │           - --args=--config=.config/terraform-docs.yml
```

</details>

Shr3ps:
(everything is working as expected with v1.92.1)

yermulnik:
Might probably be related to https://github.com/antonbabenko/pre-commit-terraform/pull/698 🤔 

@Shr3ps Would you please also share contents of the `.config/terraform-docs.yml` file?

Shr3ps:
> Might probably be related to #698 🤔
> 
> @Shr3ps Would you please also share contents of the `.config/terraform-docs.yml` file?

Yep, here you go: 

```yaml
---
# This file is automatically maintained within this module repository -- DO NOT EDIT
formatter: "markdown"

settings:
  anchor: false
  lockfile: false

output:
  file: "README.md"

sections:
  hide: [requirements]

content: |-
  ## Global versioning rule for Claranet Azure modules

  | Module version | Terraform version | AzureRM version |
  | -------------- | ----------------- | --------------- |
  | >= 7.x.x       | 1.3.x             | >= 3.0          |
  | >= 6.x.x       | 1.x               | >= 3.0          |
  | >= 5.x.x       | 0.15.x            | >= 2.0          |
  | >= 4.x.x       | 0.13.x / 0.14.x   | >= 2.0          |
[...]

PR #701, opened 2024-08-27
feat(`terraform_docs`): Start seamless migration to `terraform-docs` markers

MaxymVlasov:

- [ ] This PR fixes a bug.

In #699 was mentioned that v1.92.2 break unexpected behavior which requires to be set `- --hook-config=--create-file-if-not-exist=true`.

Because we want:
*  to set `- --hook-config=--create-file-if-not-exist=true` in v2 as one possible solution
* do not ask folks with broken configs to introduce config line which should be removed in the future

I decided that this a time to start gradually switching folks to standard `terraform-docs` markers

Relates #699

### How can we test changes

```yaml
repos:
    rev: e5209b3b0b27a5c7aea11aac86a93dfff1f0ac21
    hooks:
      - id: terraform_docs
        args:
          - --hook-config=--use-standard-markers=true
```


hooks/terraform_docs.sh
```diff
@@ -147,14 +162,16 @@ function terraform_docs {
...
         use_standard_markers=$value
+        common::colorify "yellow" "WARNING: --use-standard-markers is deprecated and will be removed in the future."
+        common::colorify "yellow" "         All needed changes already done by the hook, feel free to remove this line."
         ;;
     esac
   done
 
-  if [ "$use_standard_markers" = true ]; then
+  if [ "$use_standard_markers" = false ]; then
     # update the insertion markers to those used by terraform-docs
```
MaxymVlasov, line 172:
```suggestion
    # update the insertion markers to those used by pre-commit-terraform before v1.93
```

MaxymVlasov:
And maybe worth to check terraform-socs config provided to hook in case it uses output.template with pre-commit-terraform markers

And if we change this config, maybe remove obsolote `--hook-config` in pre-commit-config.yaml make sense too?

README.md
```diff
@@ -595,7 +595,7 @@ Unlike most other hooks, this hook triggers once if there are any changed files
         - --hook-config=--path-to-file=README.md        # Valid UNIX path. I.e. ../TFDOC.md or docs/README.md etc.
         - --hook-config=--add-to-existing-file=true     # Boolean. true or false
         - --hook-config=--create-file-if-not-exist=true # Boolean. true or false
-        - --hook-config=--use-standard-markers=true     # Boolean. Defaults in v1.x to false. Set to true for compatibility with terraform-docs
+        - --hook-config=--use-standard-markers=true     # Boolean. Default to true (v1.93+), false (<v1.93). Set to true for compatibility with terraform-docs
```
yermulnik, line 598:
```suggestion
        - --hook-config=--use-standard-markers=true     # Boolean. Defaults to true (v1.93+), false (<v1.93). Set to true for compatibility with terraform-docs
```

hooks/terraform_docs.sh
```diff
@@ -29,6 +27,23 @@ function main {
...
+#   old_insertion_marker_begin - Old insertion marker at beginning
+#   old_insertion_marker_end - Old insertion marker at the end
+# Arguments:
+#   file (string) filename to check
+#######################################################################
+function replace_old_markers {
+  local -r file=$1
+
+  sed -i "s/${old_insertion_marker_begin}/${insertion_marker_begin}/g" "$file"
+  sed -i "s/${old_insertion_marker_end}/${insertion_marker_end}/g" "$file"
```
yermulnik, line 44: * Should the replacement happen when both markers present only? 🤔 I mean if only leading or trailing old marker is present, this may identify an error in user configuration and probably should raise warning rather then silently replace only marker that is present.  Kinda `[[ $(egrep -c "^($old_insertion_marker_begin|$old_insertion_marker_end)$") -eq 2 ]]`
* I think curly brackets are extraneous as they only are needed when var expansion (et alii) is expected)
* It may make sense to add anchors of beginning and end of line to avoid inadvertent replacement of arbitrary markers used inside comment, description, or similar.
* Also the `g` option to `sed`'s replacement operator — are there expected multiple markers in the same `$file`? 🤔 
```suggestion
  sed -i "s/^$old_insertion_marker_begin$/$insertion_marker_begin/" "$file"
  sed -i "s/^$old_insertion_marker_end$/$insertion_marker_end/" "$file"
```

hooks/terraform_docs.sh
```diff
@@ -147,14 +162,16 @@ function terraform_docs {
         ;;
       --use-standard-markers)
         use_standard_markers=$value
+        common::colorify "yellow" "WARNING: --use-standard-markers is deprecated and will be removed in the future."
+        common::colorify "yellow" "         All needed changes already done by the hook, feel free to remove this line."
```
yermulnik, line 166: > this line

Which line? 🤔 Hook configuration option/parameter? I'd say let's make suggestion more specific to avoid confusion.

hooks/terraform_docs.sh
```diff
@@ -147,14 +162,16 @@ function terraform_docs {
...
         ;;
     esac
   done
 
-  if [ "$use_standard_markers" = true ]; then
-    # update the insertion markers to those used by terraform-docs
-    insertion_marker_begin="$standard_insertion_marker_begin"
-    insertion_marker_end="$standard_insertion_marker_end"
+  if [ "$use_standard_markers" = false ]; then
```
yermulnik, line 171: Using Bashism (as we relay upon Bash) may be worthwhile (I also prefer `==` over `=` to not mess with assignment operator):
```suggestion
  if [[ $use_standard_markers == false ]]; then
```

hooks/terraform_docs.sh
MaxymVlasov: >I think curly brackets are extraneous as they only are needed when var expansion (et alii) is expected)
They're here for readability reasons

Everything else applied

hooks/terraform_docs.sh
MaxymVlasov:
```suggestion
        common::colorify "yellow" "         All needed changes already done by the hook, feel free to remove --use-standard-markers setting from your pre-commit-config"
```

yermulnik (approved):

hooks/terraform_docs.sh
yermulnik: > They're here for readability reasons

Aha, got it, makes sense 👍🏻 

> Everything else applied

Just in case:
> - Should the replacement happen when both markers present only? 🤔 I mean if only leading or trailing old marker is present, this may identify an error in user configuration and probably should raise warning rather then silently replace only marker that is present. Kinda `[[ $(egrep -c "^($old_insertion_marker_begin|$old_insertion_marker_end)$" "$file") -eq 2 ]]`

ps: PR approved. the above "just in case" is up to you though.

hooks/terraform_docs.sh
```diff
@@ -147,14 +162,16 @@ function terraform_docs {
...
+        common::colorify "yellow" "         All needed changes already done by the hook, feel free to remove --use-standard-markers setting from your pre-commit config"
         ;;
     esac
   done
 
-  if [ "$use_standard_markers" = true ]; then
-    # update the insertion markers to those used by terraform-docs
-    insertion_marker_begin="$standard_insertion_marker_begin"
-    insertion_marker_end="$standard_insertion_marker_end"
+  if [[ $use_standard_markers == false ]]; then
```
MaxymVlasov, line 171: This not works
```suggestion
  if [[ $use_standard_markers == "false" ]]; then
```

antonbabenko:
This PR is included in version 1.93.0 :tada:

PR #716, opened 2024-09-09
fix(`terraform_docs`): Restore `--hook-config=--add-to-existing-file` default behavior. Regression from 1.94.0.

### LINKED ISSUE #715: `terraform_docs` starting from 1.94.0 ignores `--hook-config=--add-to-existing-file=false` (our default behavior)
### Describe the bug

```
terraform-docs --output-mode="inject" --output-file=README.md md .
```
which is exactly what we run by default 
https://github.com/antonbabenko/pre-commit-terraform/blob/91b5ba0c184ca6ef14cdca0b6d3266084dbba166/hooks/terraform_docs.sh#L279

by

```yaml
repos:
    rev: v1.94.2 # issue introduced in 1.94.0
    hooks:
      - id: terraform_docs
```

always updates/add `terraform-docs` section, as long as file exist.

This break your `--add-to-existing-file` setting, which by default set to `false`

### How can we reproduce it?

```bash
testdir=$(mktemp -d)
cd $testdir
touch main.tf README.md
echo "
repos:
    rev: v1.94.2 # issue introduced in 1.94.0
    hooks:
      - id: terraform_docs
" > .pre-commit-config.yaml
git init
git add -A
pre-commit run
```

### Workaround
* if you on Linux (and you have no Mac users) - stick to [v1.93.1](https://github.com/antonbabenko/pre-commit-terraform/releases/tag/v1.93.1)
* otherwise, stick to [v1.92.3](https://github.com/antonbabenko/pre-commit-terraform/releases/tag/v1.92.3)

MaxymVlasov:

- [x] This PR fixes a bug.

Fixes #715

### How can we test changes

broken:
```
testdir=$(mktemp -d)
cd $testdir
touch main.tf README.md
echo "
repos:
    rev: v1.94.2 # issue introduced in 1.94.0
    hooks:
      - id: terraform_docs
" > .pre-commit-config.yaml
git init
git add -A
pre-commit run
```
fixed:

```bash
testdir=$(mktemp -d)
cd $testdir
touch main.tf README.md
echo "
repos:
    rev: c0b1f38bae3b34f3f223bc73a8b384e2e7a766d7
    hooks:
      - id: terraform_docs
" > .pre-commit-config.yaml
git init
git add -A
pre-commit run
```

```bash
testdir=$(mktemp -d)
cd $testdir
touch main.tf README.md
echo "
repos:
    rev: c0b1f38bae3b34f3f223bc73a8b384e2e7a766d7
    hooks:
      - id: terraform_docs
        args:
" > .pre-commit-config.yaml
git init
git add -A
pre-commit run
```

### DIFF
--- a/hooks/terraform_docs.sh
+++ b/hooks/terraform_docs.sh
@@ -264,11 +264,11 @@ function terraform_docs {
       #? TF 0.12+ and terraform-docs 0.12.0+
 
       #
-      # If `--add-to-existing-file=false` (default behavior), check is in file exist "hook markers",
-      # and if not skip execution to avoid addition of terraform-docs section -
-      # terraform-docs in 'inject' mode adds markers by default if they not present
+      # If `--add-to-existing-file=false` (default behavior), check if "hook markers" exist in file,
+      # and, if not, skip execution to avoid addition of terraform-docs section, as
+      # terraform-docs in 'inject' mode adds markers by default if they are not present
       #
-      if ! $add_to_existing; then
+      if [[ $add_to_existing == false ]]; then
         HAVE_MARKER=$(grep -o "$insertion_marker_begin" "$output_file" || exit 0)
         [[ ! $HAVE_MARKER ]] && continue
       fi
@@ -278,18 +278,18 @@ function terraform_docs {
     else
       #? TF 0.12+ and terraform-docs < 0.8
       #? Yes, we don't cover case of TF 0.12+ and terraform-docs 0.8-0.11
-      #? but I portably just drop this section in next release of the hook,
-      #? as there no sense to support hacks for tool versions which was released more than 3 years ago
+      #? but I probably just drop this section in next release of the hook,
+      #? as there's no sense to support hacks for tool versions which were released more than 3 years ago
 
       #
-      # If `--add-to-existing-file=true` set, check is in file exist "hook markers",
-      # and if not - append "hook markers" to the end of file.
+      # If `--add-to-existing-file=true` set, check if "hook markers" exist in file,
+      # and, if not, append "hook markers" to the end of the file.
       #
-      if $add_to_existing; then
+      if [[ $add_to_existing == true ]]; then
         HAVE_MARKER=$(grep -o "$insertion_marker_begin" "$output_file" || exit 0)
 
-        if [ ! "$HAVE_MARKER" ]; then
-          # Use of insertion markers, where addToExisting=true, with no markers in the existing file
+        if [[ ! $HAVE_MARKER ]]; then
+          # Use of insertion markers, when "add_to_existing=true" with no markers in the existing file
           echo "$insertion_marker_begin" >> "$output_file"
           echo "$insertion_marker_end" >> "$output_file"
         fi
--- a/hooks/terraform_docs.sh
+++ b/hooks/terraform_docs.sh
@@ -264,17 +264,13 @@ function terraform_docs {
       #? TF 0.12+ and terraform-docs 0.12.0+
 
       #
-      # If `--add-to-existing-file=true` set, check is in file exist "hook markers",
-      # and if not - append "hook markers" to the end of file.
+      # If `--add-to-existing-file=false` (default behavior), check is in file exist "hook markers",
+      # and if not skip execution to avoid addition of terraform-docs section -
+      # terraform-docs in 'inject' mode adds markers by default if they not present
       #
-      if $add_to_existing; then
+      if ! $add_to_existing; then
         HAVE_MARKER=$(grep -o "$insertion_marker_begin" "$output_file" || exit 0)
-
-        if [ ! "$HAVE_MARKER" ]; then
-          # terraform-docs in 'inject; mode adds markers by default if they not present.
-          # So, we need need just skip execution to skip addition of terraform-docs section
-          continue
-        fi
+        [[ ! $HAVE_MARKER ]] && continue
       fi
       # shellcheck disable=SC2086
       terraform-docs --output-mode="$output_mode" --output-file="$output_file" $tf_docs_formatter $args ./ > /dev/null
@@ -285,6 +281,10 @@ function terraform_docs {
       #? but I portably just drop this section in next release of the hook,
       #? as there no sense to support hacks for tool versions which was released more than 3 years ago
 
+      #
+      # If `--add-to-existing-file=true` set, check is in file exist "hook markers",
+      # and if not - append "hook markers" to the end of file.
+      #
       if $add_to_existing; then
         HAVE_MARKER=$(grep -o "$insertion_marker_begin" "$output_file" || exit 0)
 
--- a/hooks/terraform_docs.sh
+++ b/hooks/terraform_docs.sh
@@ -269,8 +269,8 @@ function terraform_docs {
       # terraform-docs in 'inject' mode adds markers by default if they are not present
       #
       if [[ $add_to_existing == false ]]; then
-        HAVE_MARKER=$(grep -o "$insertion_marker_begin" "$output_file") || true
-        [[ ! $HAVE_MARKER ]] && continue
+        have_marker=$(grep -o "$insertion_marker_begin" "$output_file") || unset have_marker
+        [[ ! $have_marker ]] && continue
       fi
       # shellcheck disable=SC2086
       terraform-docs --output-mode="$output_mode" --output-file="$output_file" $tf_docs_formatter $args ./ > /dev/null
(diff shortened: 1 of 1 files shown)


hooks/terraform_docs.sh
```diff
@@ -260,24 +260,40 @@ function terraform_docs {
...
-
-      if [ ! "$HAVE_MARKER" ]; then
-        # Use of insertion markers, where addToExisting=true, with no markers in the existing file
-        echo "$insertion_marker_begin" >> "$output_file"
-        echo "$insertion_marker_end" >> "$output_file"
-      fi
-    fi
-
     if [[ "$terraform_docs_awk_file" == "0" ]]; then
+      #? TF 0.12+ and terraform-docs 0.12.0+
```
yermulnik, line 264: Question mark is intentional? Looks a bit unclear of what it means.

hooks/terraform_docs.sh
```diff
@@ -260,24 +260,40 @@ function terraform_docs {
...
+
+      #
+      # If `--add-to-existing-file=false` (default behavior), check is in file exist "hook markers",
+      # and if not skip execution to avoid addition of terraform-docs section -
+      # terraform-docs in 'inject' mode adds markers by default if they not present
```
yermulnik, line 269:
```suggestion
      # If `--add-to-existing-file=false` (default behavior), check if "hook markers" exist in file,
      # and, if not, skip execution to avoid addition of terraform-docs section, as
      # terraform-docs in 'inject' mode adds markers by default if they are not present
```

hooks/terraform_docs.sh
```diff
@@ -260,24 +260,40 @@ function terraform_docs {
...
+      #
+      if ! $add_to_existing; then
```
yermulnik, line 271:
```suggestion
      if [[ $add_to_existing == false ]]; then
```

hooks/terraform_docs.sh
```diff
@@ -260,24 +260,40 @@ function terraform_docs {
...
+        HAVE_MARKER=$(grep -o "$insertion_marker_begin" "$output_file" || exit 0)
```
yermulnik, line 272:
```suggestion
        HAVE_MARKER=$(grep -o "$insertion_marker_begin" "$output_file") || true
```

yermulnik: Out of curiosity: why are you using uppercase var name here (which is good), while in vast majority of use cases — lower case? 🤔 I'm not quite clear why this inconsistency is introduced here 🤷🏻 

Just for reference:
> The name space of environment variable names containing lowercase letters is reserved for applications. Applications can define any environment variables with names from this name space without modifying the behavior of the standard utilities.

hooks/terraform_docs.sh
```diff
@@ -260,24 +260,40 @@ function terraform_docs {
...
+        [[ ! $HAVE_MARKER ]] && continue
```
yermulnik, line 273: Since this var is re-used in the loop, it is best to reset it (`unset HAVE_MARKER`) at the beginning of each loop iteration, so that value from previous iteration by an accident isn't re-used by next iteration(s).

hooks/terraform_docs.sh
```diff
@@ -260,24 +260,40 @@ function terraform_docs {
...
+      fi
       # shellcheck disable=SC2086
       terraform-docs --output-mode="$output_mode" --output-file="$output_file" $tf_docs_formatter $args ./ > /dev/null
+
     else
+      #? TF 0.12+ and terraform-docs < 0.8
+      #? Yes, we don't cover case of TF 0.12+ and terraform-docs 0.8-0.11
+      #? but I portably just drop this section in next release of the hook,
```
yermulnik, line 281: Is `portably` meant to be `probably`? 🤔

hooks/terraform_docs.sh
```diff
@@ -260,24 +260,40 @@ function terraform_docs {
...
+      #? as there no sense to support hacks for tool versions which was released more than 3 years ago
```
yermulnik, line 282:
```suggestion
      #? as there's no sense to support hacks for tool versions which were released more than 3 years ago
```

hooks/terraform_docs.sh
```diff
@@ -260,24 +260,40 @@ function terraform_docs {
...
+
+      #
+      # If `--add-to-existing-file=true` set, check is in file exist "hook markers",
+      # and if not - append "hook markers" to the end of file.
```
yermulnik, line 286:
```suggestion
      # If `--add-to-existing-file=true` set, check if "hook markers" exist in file,
      # and, if not, append "hook markers" to the end of the file.
```

hooks/terraform_docs.sh
```diff
@@ -260,24 +260,40 @@ function terraform_docs {
...
+      #
+      if $add_to_existing; then
```
yermulnik, line 288:
```suggestion
      if [[ $add_to_existing == true ]]; then
```

hooks/terraform_docs.sh
```diff
@@ -260,24 +260,40 @@ function terraform_docs {
...
+        HAVE_MARKER=$(grep -o "$insertion_marker_begin" "$output_file" || exit 0)
```
yermulnik, line 289:
```suggestion
        HAVE_MARKER=$(grep -o "$insertion_marker_begin" "$output_file") || true
```

hooks/terraform_docs.sh
```diff
@@ -260,24 +260,40 @@ function terraform_docs {
...
+
+        if [ ! "$HAVE_MARKER" ]; then
```
yermulnik, line 291:
```suggestion
        if [[ ! $HAVE_MARKER ]]; then
```

hooks/terraform_docs.sh
```diff
@@ -260,24 +260,40 @@ function terraform_docs {
...
+          # Use of insertion markers, where addToExisting=true, with no markers in the existing file
```
yermulnik, line 292:
```suggestion
          # Use of insertion markers, when "add_to_existing=true" with no markers in the existing file
```

hooks/terraform_docs.sh
MaxymVlasov: Syntax highlighting 
image
https://marketplace.visualstudio.com/items?itemName=aaron-bond.better-comments

hooks/terraform_docs.sh
MaxymVlasov:
```suggestion
      #? but I probably just drop this section in next release of the hook,
```

hooks/terraform_docs.sh
MaxymVlasov: It is not introduced, it was here

We didn't adopt Google style guide, but I +- follow it
https://google.github.io/styleguide/shellguide.html#s7.2-variable-names

hooks/terraform_docs.sh
MaxymVlasov: how it can be reused if it always set right before `if`?

MaxymVlasov: Then 

```
HAVE_MARKER=$(grep -o "$insertion_marker_begin" "$output_file" || exit 0)
```

did it for us

hooks/terraform_docs.sh
yermulnik: VSCode-specific sugar?
I'm aware of `FIXME`, `TODO`, and `XXX` common keywords that are highlighted for many types of the code by variety of IDEs incl CLI editors. While it's first time is see `#?`, which is, well, is an odd sequence of chars for non-vscode users I guess 😢

hooks/terraform_docs.sh
yermulnik: Is it constant, env var or readonly? 😲 

My point is more of this looks odd among lowercase vars that are 95% of vars in this file I guess.

hooks/terraform_docs.sh
MaxymVlasov: If we add `local` to it, it will unset it on every cirlce?

yermulnik: > how it can be reused if it always set right before `if`?

believe me, people make mistakes. though it is "best to" rather than "must", so up to you to decide.

> did it for us

the `|| exit 0)` (which should be `) || true`) is to bypass `errexit`.

yermulnik: > If we add `local` to it, it will unset it on every cirlce?

I don't think so. `local` is to reduce var visibility scope.

yermulnik (approved):

hooks/terraform_docs.sh
MaxymVlasov: Changed to lowercase

Issue #718, opened 2024-09-12
Question about Invalid value for variable because of a validation rule.


dmattoninnovtech:
Hello everyone,

I test the pre-commit hook for Terraform with terraform_validate. And I have this error **Error: Invalid value for variable** because of **This was checked by the validation rule** for *var.publisher_name*.

The files are in a directory that will be used as a Terraform module.

Error message
```bash
╷
│ Error: Invalid value for variable
│
│   on variables.tf line 81:
│   81: variable "publisher_name" {
│     ├────────────────
│     │ var.publisher_name is a string
│
│ Publisher name can't be empty. The length is between 3 and 65 characters.
│
│ This was checked by the validation rule at variables.tf:85,3-13.
```

In the main.tf
```hcl
resource "azurerm_api_management" "az_apim" {
  name                = local.resource_name
  location            = var.location
  resource_group_name = var.resource_group_name
  publisher_name      = var.publisher_name
  ...
}
```

And in the variables.tf
```hcl
variable "publisher_name" {
  type        = string
  description = "The name of publisher/company."

  validation {
    condition     = can(regex("^(?!\\s*$).{3,65}$", var.publisher_name))
    error_message = "Publisher name can't be empty. The length is between 3 and 65 characters."
  }
}
```

The resource will be used later as a module and the variable *publisher_name* will be required.

So is it possible to use mock data or may be there is another processs or method to avoid this kind of error ?

Thanks for your help.
David

MaxymVlasov:
How do you provide value to this variable now?

Because in current setup `teraform plan` will fail with same error

dmattoninnovtech:
Actually I don't provide any value, as it is a Terraform module.
I just want to validate that my development is OK for this TF module, but may be I shouldn't use terraform_validate.

Later I will use this module with Terragrunt and provide values.

MaxymVlasov:
That's interesting, as  we have similar validation options in module and never faced such errors.

Whatever, try https://github.com/antonbabenko/pre-commit-terraform?tab=readme-ov-file#all-hooks-set-env-vars-inside-hook-at-runtime

https://developer.hashicorp.com/terraform/language/values/variables#environment-variables

dmattoninnovtech:
Well, I just try with these options and it doesn't work. I have the same error message

```
# TEST 1
- id: terraform_validate
  args:
    - --env-vars=TF_VAR_publisher_name=publisher_name

# TEST 2
- id: terraform_validate
  args:
[...]

PR #719, opened 2024-09-13
feat: Expand environment variables in `--args=` which contains lowercase symbols, like `${TF_VAR_lowercase}` 

MaxymVlasov:

- [ ] This PR fixes a bug.

Improves https://github.com/antonbabenko/pre-commit-terraform?tab=readme-ov-file#all-hooks-usage-of-environment-variables-in---args
Found in https://github.com/antonbabenko/pre-commit-terraform/issues/718#issuecomment-2347102202

Support of [this option](https://developer.hashicorp.com/terraform/language/values/variables#environment-variables) probably will not be needed in any observed use cases, but it is cheap to implement, so here we are

### DIFF
--- a/hooks/_common.sh
+++ b/hooks/_common.sh
@@ -128,7 +128,7 @@ function common::parse_and_export_env_vars {
     while true; do
       # Check if at least 1 env var exists in `$arg`
       # shellcheck disable=SC2016 # '${' should not be expanded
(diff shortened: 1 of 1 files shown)


antonbabenko:
This PR is included in version 1.96.0 :tada:

PR #714, opened 2024-09-05
fix: Support custom TF paths which contains spaces

robinbowes:
Put an `x` into the box if that apply:

- [ ] This PR introduces breaking change.
- [x] This PR fixes a bug.
- [ ] This PR adds new functionality.
- [ ] This PR enhances existing functionality.

### Description of your changes

Add quotes around variable containing path to terraform so ti works when the path contains a space.

Fixes #713

PR #723, opened 2024-10-10
chore: Update CHANGELOG.md

MaxymVlasov:
I don't know is it will not be removed by GHA in the future, but let add it here too

Based on https://github.com/antonbabenko/pre-commit-terraform/pull/714#issuecomment-2405443291


antonbabenko (commented):
It will be gone when changelog file is updated on release

MaxymVlasov:
Meh. Okay then, will hope that update in GitHub releases will be sufficient

antonbabenko:
It generates from commit messages. No commit - no mention in the changelog file.

PR #728, opened 2024-10-30
fix(WSL): Make parallelism work appropriately

### LINKED ISSUE #727: Script `_common.sh` fails to detect correct CPU count in WSL (Windows Subsystem for Linux) distro
### Describe the bug

The script `_common.sh` fails to detect correct CPU count in WSL distributions.

### How can we reproduce it?

Export the following two functions from the mentioned script:

- `common::get_cpu_num`
- `common::colorify`

Run `common::get_cpu_num`

You will see this output:

```
Unable to derive number of available CPU cores.
 Running inside K8s pod without limits or inside DinD without limits propagation.
 To avoid possible harm, parallelism is disabled.
 To re-enable it, set corresponding limits, or set the following for the current hook:
   args:
     - --hook-config=--parallelism-ci-cpu-cores=N
 where N is the number of CPU cores to allocate to pre-commit.
1
```

This is wrong. This snippet from get_cpu_num shows the correct CPU count:

```
> nproc 2> /dev/null || sysctl -n hw.ncpu 2> /dev/null || echo 1
20
```

But the script never reaches this point, because

```
/sys/fs/cgroup/cpu/cpu.cfs_quota_us
``` 

exists (and it contains -1).

So the problem is that the hook is not running inside K8s pod or inside DinD but inside WSL. It is possible to set CPU limits for WSL. But by default WSL is allowed to use all available CPUs with no limit.

### Environment information

```
WSL-Version: 2.3.24.0
Kernelversion: 5.15.153.1-2
WSLg-Version: 1.0.65
MSRDC-Version: 1.2.5620
Direct3D-Version: 1.611.1-81528511
DXCore-Version: 10.0.26100.1-240331-1435.ge-release
Windows-Version: 10.0.22621.4317
```

Hook version is v4.6.0

---

Thanks for the hooks.

yermulnik:

- [x] This PR fixes a bug.

Fixes #727

### DIFF
--- a/hooks/_common.sh
+++ b/hooks/_common.sh
@@ -199,8 +199,9 @@ function common::get_cpu_num {
 
   local millicpu
 
-  if [[ -f /sys/fs/cgroup/cpu/cpu.cfs_quota_us ]]; then
-    # Inside K8s pod or DinD in K8s
+  if [[ -f /sys/fs/cgroup/cpu/cpu.cfs_quota_us &&
+    ! -f /proc/sys/fs/binfmt_misc/WSLInterop ]]; then
+    # Inside K8s pod or DinD in K8s (but not inside WSL)
     millicpu=$(< /sys/fs/cgroup/cpu/cpu.cfs_quota_us)
 
     if [[ $millicpu -eq -1 ]]; then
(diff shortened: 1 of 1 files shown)


MaxymVlasov (approved):
Let's wait for confirmation from reporter that it works for him, and then it ready to merge

antonbabenko:
This PR is included in version 1.96.2 :tada:

PR #734, opened 2024-12-28
ci: Disable checks unrunable @ pre-commit.ci

webknjaz:

- [ ] This PR fixes a bug.
- [x] This is a contributor/maintainer facing change that in no way affects the end-users

Enabling faster feedback loop for new contributors by enabling https://pre-commit.ci in this repo. 

The only visible effect of this is when said app is installed and the service is running the checks, it would skip the checks that rely on the external executables that might be missing in runtime. It's necessary since that testing environment is controlled by someone else.

### How can we test changes

@antonbabenko  go to https://github.com/apps/pre-commit-ci/installations/select_target, install the app into this repo and then merge the PR.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,3 @@
----
-
 ci:
   autoupdate_schedule: quarterly
   skip:
@@ -68,5 +66,3 @@ repos:
   - id: prettier
     # https://prettier.io/docs/en/options.html#parser
     files: '.json5$'
-
-...
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,12 @@
+---
+
+ci:
+  autoupdate_schedule: quarterly
+  skip:
+    - shfmt
+    - shellcheck
+    - hadolint
+
 repos:
 - repo: https://github.com/pre-commit/pre-commit-hooks
   rev: v5.0.0
@@ -59,3 +68,5 @@ repos:
   - id: prettier
     # https://prettier.io/docs/en/options.html#parser
     files: '.json5$'
+
+...


MaxymVlasov:
We have pre-commit via GHA, and it runs all hooks specified in .pre-commit-config, not just some of them

webknjaz:
Alternatively, there's https://github.com/apps/pre-commit-ci-lite/installations/new + https://pre-commit.ci/lite.html that implements backpushes.

yermulnik:
M-m-m, I definitely would suggest to add more context for other contributors, including a brief explanation added as a comment to the code that is changing for future reference and for posterity.

webknjaz:
Done.

.pre-commit-config.yaml
```diff
@@ -1,3 +1,12 @@
+---
+
```
MaxymVlasov, line 2: It's a single YAML document, so no need to add this boilerplate 
```suggestion
```

.pre-commit-config.yaml
```diff
@@ -59,3 +68,5 @@ repos:
   - id: prettier
     # https://prettier.io/docs/en/options.html#parser
     files: '.json5$'
+
+...
```
MaxymVlasov, line 72: Same for [end file notation](https://stackoverflow.com/a/66542407). TBH, that's first time when I see that syntax, and I saw much more YAMLs than I'd like to have.

Can't imagine situation when someone will try to merge this file with anything else and don't implement basic separation between files

```suggestion
```

.pre-commit-config.yaml
yermulnik: That also designates a start of YAML document =)
Though I agree: no need for it if it is not required.

.pre-commit-config.yaml
yermulnik: +1

webknjaz:
pre-commit.ci run

webknjaz:
It's not marked as required in branch protection, though.

antonbabenko:
@webknjaz Updated.
image

webknjaz:
This check is now failing in other PRs and will keep failing until this PR is merged.

antonbabenko:

PR #735, opened 2024-12-28
chore(packaging): Fix computing SCM-based version @ the dists

webknjaz:
The previously existing attempt to use `setuptools-git-version` never worked as it wasn't integrated correctly — it was added as a runtime dependency while it's meant to be a build dependency. This resulted in 0.0.0 being used all the time.

This patch adds a hatchling plugin that implements the original idea in a way that actually works.

- [x] This PR fixes a bug.

$sbj.

### How can we test changes

`python -Im build`

### DIFF
--- a/hatch.toml
+++ b/hatch.toml
@@ -10,3 +10,10 @@ exclude = [
 packages = [
   'hooks/',
 ]
+
+[metadata.hooks.vcs.urls]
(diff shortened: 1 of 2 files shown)


MaxymVlasov (approved):

antonbabenko:

PR #736, opened 2024-12-28
chore(packaging): X📦 Convert project to src-layout

webknjaz:
📦 Convert project to src-layout

This is one of the most resilient Python project structures that
offers isolation and prevents accidental local imports during testing
[[1]].

The change also includes moving the deprecation message into a warning
raised from the affected script in runtime as opposed to import time.
In general, it's harmful to have such logic in `__init__.py` since
it's executed at import time and would get triggered when interacting
with other importables from the same namespace just as well.

[1]: https://blog.ganssle.io/articles/2019/08/test-as-installed.html

- [ ] This PR fixes a bug.

$sbj.

### How can we test changes

`python -Im build`


src/pre_commit_terraform/terraform_docs_replace.py
```diff
@@ -27,6 +28,14 @@ def main(argv=None):
     parser.add_argument('filenames', nargs='*', help='Filenames to check.')
     args = parser.parse_args(argv)
 
+    warnings.warn(
+        '`terraform_docs_replace` hook is DEPRECATED.'
+        'For migration instructions see '
+        'https://github.com/antonbabenko/pre-commit-terraform/issues/248'
+        '#issuecomment-1290829226',
```
MaxymVlasov, line 35: There is a linter in my branch, which ask to avoid implicit string concatenations, so, let's not make it less happy (no matter that this hook is not supported anymore)
```suggestion
        'For migration instructions see '
        + 'https://github.com/antonbabenko/pre-commit-terraform/issues/248'
        + '#issuecomment-1290829226',
```

MaxymVlasov (approved):
I surprised that this time `__init__.py` can be finally removed and not causes any issues :tada:

src/pre_commit_terraform/terraform_docs_replace.py
webknjaz: You missed one line in the code suggestion. Also, we talked about disabling that rule, so I suggest skipping this change. Besides, this isn't exactly in the scope of the PR as the implicit concatenation was used in `__init__.py` and I didn't change that.

MaxymVlasov: Yep, I run test and explitit concatenation is =< 1 * 10^-8 more time consuming than implisit
I don't think that it's a big deal in our case, but i'm good to merge it now as is

yermulnik: > is =< 1 * 10^-8 seconds more time consuming
> I don't think that it's a big deal in our case, but i'm good to merge it now as is

Jeez, dude, you're throwing nanoseconds around like a drunken sailor 🤣

antonbabenko:

PR #738, opened 2024-12-30
chore(refactoring): Introduce a Python CLI app layout

webknjaz:
This includes a structure with purpose-based modules and a standard mechanism for adding more subcommands.

When adding a new subcommand, one has to define the `invoke_cli_app()` and `populate_argument_parser()` hooks in their new module together with a `CLI_SUBCOMMAND_NAME` string constant. I can then be wired into the framework via `_cli_subcommands.py`. This is the only integration point necessary.

`populate_argument_parser()` accepts a subparser instance of `argparse.ArgumentParser()` that a new subcommand would need to attach new arguments into. It does not need to return anything. And the `invoke_cli_app()` hook is called with an instance of `argparse.Namespace()` with all the arguments parsed and pre-processed. This function is supposed to have the main check logic and return an instance of `._structs.ReturnCode()` or `int`.
`CLI_SUBCOMMAND_NAME` needs to be annotated as a `Final[str]`.

`pre-commit` integration is done through the runpy interface: `python -m pre_commit_terraform <subcommand>`.

The import package folder also contains a more detailed maintenance manual: https://github.com/antonbabenko/pre-commit-terraform/tree/8e77580805322e5d264a2667b4f4cba53e8bf94c/src/pre_commit_terraform#readme.

- [ ] This PR fixes a bug.
- [x] This PR is an internal restructuring change.

I think the most important details are in the commit messages. Everything that is unlikely to change much over time is in "_private" modules.

### How can we test changes

Something like `python -Im pip install -e . && python -Im pre_commit_terraform replace-docs`. Or invoke it via pre-commit.

### DIFF
--- a/src/pre_commit_terraform/_cli_parsing.py
+++ b/src/pre_commit_terraform/_cli_parsing.py
@@ -22,11 +22,11 @@ def attach_subcommand_parsers_to(root_cli_parser: ArgumentParser, /) -> None:
         required=True,
     )
     for subcommand_module in SUBCOMMAND_MODULES:
-        replace_docs_parser = subcommand_parsers.add_parser(subcommand_module.CLI_SUBCOMMAND_NAME)
-        replace_docs_parser.set_defaults(
+        subcommand_parser = subcommand_parsers.add_parser(subcommand_module.CLI_SUBCOMMAND_NAME)
+        subcommand_parser.set_defaults(
             invoke_cli_app=subcommand_module.invoke_cli_app,
         )
-        subcommand_module.populate_argument_parser(replace_docs_parser)
+        subcommand_module.populate_argument_parser(subcommand_parser)
 
 
 def initialize_argument_parser() -> ArgumentParser:
--- a/src/pre_commit_terraform/_cli_parsing.py
+++ b/src/pre_commit_terraform/_cli_parsing.py
@@ -1,0 +1,43 @@
+"""Argument parser initialization logic.
+
+This defines helpers for setting up both the root parser and the parsers
+of all the sub-commands.
+"""
+
+from argparse import ArgumentParser
+
+from .terraform_docs_replace import (
+    populate_argument_parser as populate_replace_docs_argument_parser,
+)
+
+
+PARSER_MAP = {
+    'replace-docs': populate_replace_docs_argument_parser,
+}
+
+
+def attach_subcommand_parsers_to(root_cli_parser: ArgumentParser, /) -> None:
+    """Connect all sub-command parsers to the given one.
+
+    This functions iterates over a mapping of subcommands to their
+    respective population functions, executing them to augment the
+    main parser.
+    """
+    subcommand_parsers = root_cli_parser.add_subparsers(
+        dest='check_name',
+        help='A check to be performed.',
+        required=True,
+    )
+    for subcommand_name, initialize_subcommand_parser in PARSER_MAP.items():
+        replace_docs_parser = subcommand_parsers.add_parser(subcommand_name)
+        initialize_subcommand_parser(replace_docs_parser)
+
+
+def initialize_argument_parser() -> ArgumentParser:
+    """Return the root argument parser with sub-commands."""
+    root_cli_parser = ArgumentParser(prog=f'python -m {__package__ !s}')
+    attach_subcommand_parsers_to(root_cli_parser)
+    return root_cli_parser
+
+
+__all__ = ('initialize_argument_parser',)
--- a/src/pre_commit_terraform/_cli_parsing.py
+++ b/src/pre_commit_terraform/_cli_parsing.py
@@ -6,14 +6,7 @@ of all the sub-commands.
 
 from argparse import ArgumentParser
 
-from .terraform_docs_replace import (
-    populate_argument_parser as populate_replace_docs_argument_parser,
-)
-
-
-PARSER_MAP = {
-    'replace-docs': populate_replace_docs_argument_parser,
-}
+from ._cli_subcommands import SUBCOMMAND_MODULES
 
 
 def attach_subcommand_parsers_to(root_cli_parser: ArgumentParser, /) -> None:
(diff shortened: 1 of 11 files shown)


yermulnik:
> This includes a structure with purpose-based modules and a standard mechanism for adding more subcommands.

Cool. Thanks. Too much of Python'ish stuff for me 🤣 

> When adding a new subcommand, one has to wire the `invoke_cli_app()` and `populate_argument_parser()` from their new module into the mappings defined in `_cli_subcommands.py` and `_cli_parsing.py` respectively. This is the only integration point necessary.
> 
> `populate_argument_parser()` accepts a subparser instance of `argparse.ArgumentParser()` that a new subcommand would need to attach new arguments into. It does not need to return anything. And the `invoke_cli_app()` hook is called with an instance of `argparse.Namespace()` with all the arguments parsed and pre-processed. This function is supposed to have the main check logic and return an instance of `._structs.ReturnCode()` or `int`.
>
> ### Description of your changes
> I think the most important details are in the commit message. Everything that is unlikely to change much over time is in "_private" modules.
> 
> ### How can we test changes
> Something like `python -Im pip install -e . && python -Im pre_commit_terraform replace-docs`. Or invoke it via pre-commit.

Could you please drop a README.md file with the above info into the directory with these files for future reference? Thank you.

src/pre_commit_terraform/_cli_parsing.py
```diff
@@ -0,0 +1,43 @@
...
+from argparse import ArgumentParser
+
+from .terraform_docs_replace import (
+    populate_argument_parser as populate_replace_docs_argument_parser,
+)
+
+
+PARSER_MAP = {
+    'replace-docs': populate_replace_docs_argument_parser,
+}
```
MaxymVlasov, line 16: So I will need to add here all python hooks?

src/pre_commit_terraform/_structs.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class ReturnCode(IntEnum):
+    """POSIX-style return code values.
+
+    To be used in check callable implementations.
+    """
+
+    OK = 0
+    ERROR = 1
```
MaxymVlasov, line 13: There can be other return codes. We do not control what 3-rd party tools return. At least few of them uses codes from 0 to 4.

In actual hooks, not deprecated many years ago `terraform_docs_replace.py`, return codes will be looks like

https://github.com/antonbabenko/pre-commit-terraform/blob/33e15479baaf742929c12b1464dec873a5004bfe/hooks/terraform_checkov.py#L98
(result of `subprocess.run`)

MaxymVlasov (commented):
Why there so much boilerplate -_-

src/pre_commit_terraform/_cli_parsing.py
MaxymVlasov: And how it should work with `common.py`?

I don't understand how I will use it with 

https://github.com/antonbabenko/pre-commit-terraform/blob/33e15479baaf742929c12b1464dec873a5004bfe/hooks/terraform_fmt.py#L29

MaxymVlasov:
Why you used `terraform_docs_replace.py` at all? This crap does not represent anything how current hooks works. 

Sorry, but I have no idea how to reuse half of code implemented here with actual hooks

webknjaz:
> Why you used `terraform_docs_replace.py` at all? This crap does not represent anything how current hooks works.

Because it is a working example of implementing custom check subcommands until you add new ones. You can swap it out for new things later.

> Sorry, but I have no idea how to reuse half of code implemented here with actual hooks

Hence, the example integration of the only existing thing. I'll drop the instructions into a readme as suggested.

src/pre_commit_terraform/_structs.py
webknjaz: That's true. Which is why typing is set up to allow arbitrary integers. But for standard codes, it's good to have something reusable, structured and type-checked.

Some tools also represent a number of linting violations in the status code. There's many things that are possible. The scope of this PR is a bare-bones skeleton that can be augmented later on.

src/pre_commit_terraform/_cli_parsing.py
webknjaz: Yes, this mapping is for parsers, and another file has a mapping for callables with the logic.

webknjaz: > I don't understand how I will use it with

You won't need to call `common.parse_cmdline(argv)` manually in several places anymore. It'll be called for you, and you'll get the result of a `parse_args()` invocation into a function you'll define to accept an `argparse.Namespace` instance.

webknjaz:
I've added a document. While doing that, I realized that I can simplify the new subcommand integration a bit, so I've marked this as a draft until I do that.

yermulnik:
> I've added a document.

Pretty nice and clear ❤️

webknjaz:
@yermulnik @MaxymVlasov this is now mergeable, but #734 should go in first to unblock the CI.

webknjaz:
pre-commit.ci run

yermulnik (commented):
LGTM. Though since @MaxymVlasov is mainly the only one who develops code at this moment I'm deferring final decision to him.

src/pre_commit_terraform/_cli_parsing.py
```diff
@@ -0,0 +1,39 @@
...
+        dest='check_name',
+        help='A check to be performed.',
+        required=True,
+    )
+    for subcommand_module in SUBCOMMAND_MODULES:
+        replace_docs_parser = subcommand_parsers.add_parser(subcommand_module.CLI_SUBCOMMAND_NAME)
+        replace_docs_parser.set_defaults(
+            invoke_cli_app=subcommand_module.invoke_cli_app,
+        )
+        subcommand_module.populate_argument_parser(replace_docs_parser)
```
MaxymVlasov, line 29:
```suggestion
        subcommand_parser = subcommand_parsers.add_parser(subcommand_module.CLI_SUBCOMMAND_NAME, add_help=False,)
        subcommand_parser.set_defaults(
            invoke_cli_app=subcommand_module.invoke_cli_app,
        )
        subcommand_module.populate_argument_parser(subcommand_parser)
```

MaxymVlasov (approved):
Let's hope that I'll understand all later :)

IT somehow works with new hooks, but need some polishing, which will be done in other PRs

antonbabenko:

PR #740, opened 2024-12-31
fix(wip): Access `.pre-commit-hooks.yaml` as artifact

webknjaz:
Previously, the path lookup logic was unstable due to a flawed heuristic, which meant guessing the artifact path incorrectly in some cases.

This patch improves that by employing a reliable approach of putting the artifact into a predictable location within the installed import package. It then uses a standard `importlib.resources` module to retrive it from the always-existing place.

The only possibly unobvious quirk is that during development, whenever this file changes in the Git repository, the `pip install` command must be executed again since this is what refreshes the file contents in the installed location.

- [ ] This PR fixes a bug.
- [x] This PR is made against a WIP branch and does not immediately affect anybody.

@MaxymVlasov this fixes the file lookup instability we faced during the call.

### How can we test changes

`pip install` and run the thing using the direct execution via `python -m` and both `pre-commit` tool methods.

### DIFF
--- a/hatch.toml
+++ b/hatch.toml
@@ -1,5 +1,6 @@
 [build.targets.sdist]
 include = [
+  '.pre-commit-hooks.yaml',
   'src/',
 ]
 
@@ -7,6 +8,9 @@ include = [
 packages = [
   'src/pre_commit_terraform/',
 ]
+
+[build.targets.wheel.force-include]
+'.pre-commit-hooks.yaml' = 'pre_commit_terraform/_artifacts/.pre-commit-hooks.yaml'
 
 [metadata.hooks.vcs.urls]
 'Source Archive' = 'https://github.com/antonbabenko/pre-commit-terraform/archive/{commit_hash}.tar.gz'
--- a/src/pre_commit_terraform/_cli.py
+++ b/src/pre_commit_terraform/_cli.py
@@ -26,7 +26,7 @@ def invoke_cli_app(cli_args: list[str]) -> ReturnCodeType:
         raise
     except PreCommitTerraformRuntimeError as unhandled_exc:
         print(
-            f'App execution took an unexpected turn: {unhandled_exc !s}. ' 'Exiting...',
+            f'App execution took an unexpected turn: {unhandled_exc !s}. Exiting...',
             file=stderr,
         )
         return ReturnCode.ERROR
--- a/src/pre_commit_terraform/_cli.py
+++ b/src/pre_commit_terraform/_cli.py
@@ -3,11 +3,9 @@
 from sys import stderr
 
 from ._cli_parsing import initialize_argument_parser
-from ._errors import (
-    PreCommitTerraformBaseError,
-    PreCommitTerraformExit,
-    PreCommitTerraformRuntimeError,
-)
+from ._errors import PreCommitTerraformBaseError
+from ._errors import PreCommitTerraformExit
+from ._errors import PreCommitTerraformRuntimeError
 from ._structs import ReturnCode
 from ._types import ReturnCodeType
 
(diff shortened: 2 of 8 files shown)


src/pre_commit_terraform/_cli.py
```diff
@@ -28,8 +26,7 @@ def invoke_cli_app(cli_args: list[str]) -> ReturnCodeType:
         raise
     except PreCommitTerraformRuntimeError as unhandled_exc:
         print(
-            f'App execution took an unexpected turn: {unhandled_exc !s}. '
-            'Exiting...',
+            f'App execution took an unexpected turn: {unhandled_exc !s}. ' 'Exiting...',
```
yermulnik, line 29: This folding looks incorrect
```suggestion
            f'App execution took an unexpected turn: {unhandled_exc !s}. Exiting...',
```

hatch.toml
```diff
@@ -8,6 +9,9 @@ packages = [
   'src/pre_commit_terraform/',
 ]
 
+[build.targets.wheel.force-include]
+'.pre-commit-hooks.yaml' = 'pre_commit_terraform/_artifacts/.pre-commit-hooks.yaml'
```
yermulnik, line 13: I'm a way too unfamiliar with all this Python magic, though I'm looking into how we can skip the non-obvious step:
> The only possibly unobvious quirk is that during development, whenever this file changes in the Git repository, the `pip install` command must be executed again since this is what refreshes the file contents in the installed location.

Can this be remediated by e.g. symlinking the `/.pre-commit-hooks.yaml` into `/pre_commit_terraform/_artifacts/.pre-commit-hooks.yaml`?
Else we need to either add a GHA to get this file copied automatically or drop a comment into `/.pre-commit-hooks.yaml` file about a need to run a command once the file is changed. WDYT?

src/pre_commit_terraform/_cli.py
webknjaz: This is how the formatters are set up in Max's branch, it's not something that's a part of the PR originally.

hatch.toml
webknjaz: I don't think this is necessary. The GHA bit would be a part of the test infra PR that I didn't get to yet. The local quirk is already mentioned in the readme. And the pre-commit invocations do this under the hood, we don't have any control over it, this is always done.

webknjaz: Also note that this is a PR against a non-master branch.

webknjaz: Symlinking won't work because those are turned into regular files during packaging and aren't preserved outside the repo.

yermulnik: Great. Thanks for details. That all makes sense. Especially the "the pre-commit invocations do this under the hood, we don't have any control over it, this is always done" bit which just discards my inquiry as the automation is kinda there already 👍🏻

MaxymVlasov (approved):

hatch.toml
webknjaz: Yeah.. To expand on this, basically, `pre-commit` uses `pip install` as a method to source the dependencies of your hook. As a side effect, it also installs it as a regular Python package distribution. To install a regular folder or a Git repo this way, a dist should first be built. That dist is a wheel (`.whl`: https://packaging.python.org/en/latest/glossary/#term-Wheel) — a zip file with your project in it's “built” state; it's essentially a `zip` archive that gets unpacked into `site-packages/` (plus some metadata / deps handling on install).

When you what to test your app locally w/o pre-commit, you can reproduce a similar thing via `pip install .` and also `pip install pytest` when you get to using `pytest` for testing. However, during development you usually use editable mode `pip install -e .` which allows changing files in your Git repo without reinstalling (mostly) — it also builds wheels under the hood, but those wheels are special and almost don't contain any payload, short of the metadata and the artifacts. This is one of those artifacts that gets into these wheels and then is copied into `site-packages/`.

PR #741, opened 2025-01-03
Rewrite hooks on python back to main

MaxymVlasov:

- [ ] This PR fixes a bug.


.github/CONTRIBUTING.md
```diff
@@ -154,3 +155,12 @@ You can use [this PR](https://github.com/antonbabenko/pre-commit-terraform/pull/
...
 2. Create and populate a new hook section in [Hooks usage notes and examples](../README.md#hooks-usage-notes-and-examples).
+
+
+## Testing
+
+```
+pip install pytest pytest-mock
+
+pytest -vv
+```
```
MaxymVlasov, line 166:
````suggestion
## Testing python hooks

Tu run tests, you need:

```bash
pip install pytest pytest-mock

pytest -vv
```

To run and debug hooks as CLI:

1. Create [`venv`](https://docs.python.org/3/library/venv.html) and activate it, IE by `virtualenv venv`
2. Inside virtual env install pre-commit-terraform as package:
	```bash
	pip install --editable .
	```

3. Run next to show basic help:
	```bash
	python -Im pre_commit_terraform --help
	```
````

File: .github/CONTRIBUTING.md [hunk 2]
```diff
@@ -18,6 +18,7 @@ Enjoy the clean, valid, and documented code!
   * [Prepare basic documentation](#prepare-basic-documentation)
   * [Add code](#add-code)
   * [Finish with the documentation](#finish-with-the-documentation)
+* [Testing](#testing)
```
MaxymVlasov, line 21:
```suggestion
* [Testing](#testing-python-hooks)
```

.pre-commit-hooks.yaml
```diff
@@ -15,6 +15,15 @@
...
 
+- id: terraform_fmt_py
+  name: Terraform fmt
+  description: Rewrites all Terraform configuration files to a canonical format.
+  require_serial: true
+  entry: python -Im pre_commit_terraform terraform_fmt_py
+  language: python
+  files: \.tf(vars)?$
+  exclude: \.terraform/.*$
+
```
MaxymVlasov, line 26:
```suggestion
```

.pre-commit-hooks.yaml
```diff
@@ -138,6 +147,16 @@
...
+- id: terraform_checkov_py
+  name: Checkov
+  description: Runs checkov on Terraform templates.
+  entry: python -Im pre_commit_terraform terraform_checkov_py
+  language: python
+  always_run: false
+  files: \.tf$
+  exclude: \.terraform/.*$
+  require_serial: true
+
```
MaxymVlasov, line 159:
```suggestion
```

.github/CONTRIBUTING.md
```diff
@@ -157,10 +157,25 @@ You can use [this PR](https://github.com/antonbabenko/pre-commit-terraform/pull/
 2. Create and populate a new hook section in [Hooks usage notes and examples](../README.md#hooks-usage-notes-and-examples).
 
 
-## Testing
+## Testing python hooks
 
-```
+Tu run tests, you need:
```
robinbowes, line 162: Trivial typo: `s/Tu/To/`

.github/CONTRIBUTING.md
```diff
@@ -18,6 +18,7 @@ Enjoy the clean, valid, and documented code!
   * [Prepare basic documentation](#prepare-basic-documentation)
   * [Add code](#add-code)
   * [Finish with the documentation](#finish-with-the-documentation)
+* [Testing](#testing-python-hooks)
```
MaxymVlasov, line 21:
```suggestion
* [Testing python hooks](#testing-python-hooks)
```

.github/CONTRIBUTING.md
MaxymVlasov:
```suggestion
To run tests, you need:
```

src/pre_commit_terraform/_common.py
```diff
@@ -0,0 +1,184 @@
...
+    for env_var_str in env_var_strs:
+        name, value = env_var_str.split('=', 1)  # noqa: WPS110 # 'value' is valid var name here
+        if value.startswith('"') and value.endswith('"'):
+            value = value[1:-1]  # noqa: WPS110 # 'value' is valid var name here
+        env_var_dict[name] = value
+
+    return env_var_dict
+
+
+def _get_unique_dirs(files: list[str]) -> set[str]:
```
yermulnik, line 38: The underscore as a prefix looks a bit extraneous since other functions seem to don't have it. Is it intentional e.g. to replicate shell-based origin?

src/pre_commit_terraform/_common.py
```diff
@@ -0,0 +1,184 @@
...
+    tofu_path = shutil.which('tofu')
+    if tofu_path:
+        return tofu_path
+
+    # If no binary is found, raise an exception
+    raise BinaryNotFoundError(
+        'Neither Terraform nor OpenTofu binary could be found. Please either set the "--tf-path"'
+        + ' hook configuration argument, or set the "PCT_TFPATH" environment variable, or set the'
+        + ' "TERRAGRUNT_TFPATH" environment variable, or install Terraform or OpenTofu globally.',
+    )
```
yermulnik, line 184: Maybe split to multiline to improve readability for users? 🤔

src/pre_commit_terraform/_logger.py
```diff
@@ -0,0 +1,87 @@
+"""Here located logs-related functions."""
```
yermulnik, line 1:
```suggestion
"""Logs-related functions."""
```

src/pre_commit_terraform/_run_on_whole_repo.py
```diff
@@ -0,0 +1,103 @@
...
+        func_name (str): The name of the function to check.
+
+    Returns:
+        bool: True if the function is defined, False otherwise.
+    """
+    is_defined = func_name in scope
+    is_callable = callable(scope[func_name]) if is_defined else False
+
+    logger.debug(
+        'Checking if "%s":\n1. Defined in hook: %s\n2. Is it callable: %s',
```
yermulnik, line 28:
```suggestion
        'Checking if "%s":\n1. Defined in hook: %s\n2. Is callable: %s',
```

src/pre_commit_terraform/_run_on_whole_repo.py
```diff
@@ -0,0 +1,103 @@
...
+
+    # Get the directory containing the packaged `.pre-commit-hooks.yaml` copy
+    artifacts_root_path = access_artifacts_of('pre_commit_terraform') / '_artifacts'
+    pre_commit_hooks_yaml_path = artifacts_root_path / '.pre-commit-hooks.yaml'
+    pre_commit_hooks_yaml_path.read_text(encoding='utf-8')
+
+    logger.debug('Hook config path: %s', pre_commit_hooks_yaml_path)
+
+    # Read the .pre-commit-hooks.yaml file
+    pre_commit_hooks_yaml_txt = pre_commit_hooks_yaml_path.read_text(encoding='utf-8')
```
yermulnik, line 61: What's the point of the `read_text()` on line 56? 🤔

src/pre_commit_terraform/_run_on_whole_repo.py
```diff
@@ -0,0 +1,103 @@
...
+            break
+    else:
+        raise ValueError(f'Hook ID "{hook_id}" not found in .pre-commit-hooks.yaml')
+
+    logger.debug(
+        'Included files pattern: %s\nExcluded files pattern: %s',
+        included_pattern,
+        excluded_pattern,
+    )
+    # S607 disabled as we need to maintain ability to call git command no matter where it located.
```
yermulnik, line 78: * It's best to provide also human readable description of excluded check — this allows readers to more clearly understand logic
* plus this syggestion
```suggestion
    # S607 disabled as we need to maintain ability to call git command no matter where it is located.
```

src/pre_commit_terraform/_run_on_whole_repo.py
```diff
@@ -0,0 +1,103 @@
...
+    git_ls_files_cmd = ['git', 'ls-files']  # noqa: S607
```
yermulnik, line 79: I've noticed you already used `shutil.which()` to determine path to binary. It might make sense to use it here too to use `git` binary directly instead of an arbitrary `git` command be means of shell alias, function, or similar. WDYT?

src/pre_commit_terraform/_run_on_whole_repo.py
```diff
@@ -0,0 +1,103 @@
...
+        ]
+    else:
+        all_file_paths_that_can_be_checked = [
+            file_path for file_path in git_ls_file_paths if included_pattern.search(file_path)
+        ]
+
+    # Get the sorted list of files passed to the hook
+    file_paths_to_check = sorted(file_paths)
+    logger.debug(
+        'Files to check:\n%s\n\nAll files that can be checked:\n%s\n\nIdentical lists: %s',
```
yermulnik, line 97:
```suggestion
        'Files to check:\n%s\n\nAll files that can be checked:\n%s\n\nAre these lists identical: %s',
```

src/pre_commit_terraform/_run_on_whole_repo.py
```diff
@@ -0,0 +1,103 @@
...
+        file_paths_to_check,
+        all_file_paths_that_can_be_checked,
+        file_paths_to_check == all_file_paths_that_can_be_checked,
+    )
+    # Compare the sorted lists of file
```
yermulnik, line 102:
```suggestion
    # Compare the sorted lists of files
```

src/pre_commit_terraform/terraform_docs_replace.py
```diff
@@ -1,77 +1,95 @@
...
+
+    Args:
+        parsed_cli_args: Parsed arguments from CLI.
+
+    Returns:
+        int: The exit code of the hook.
+    """
+
     warnings.warn(
         '`terraform_docs_replace` hook is DEPRECATED.'
```
yermulnik, line 65:
```suggestion
        '`terraform_docs_replace` hook is DEPRECATED. '
```

tests/pytest/test__cli_parsing.py
```diff
@@ -0,0 +1,223 @@
...
+import pytest
+
+from pre_commit_terraform._cli_parsing import attach_subcommand_parsers_to
+from pre_commit_terraform._cli_parsing import initialize_argument_parser
+from pre_commit_terraform._cli_parsing import populate_common_argument_parser
+
+
+# ?
+# ? populate_common_argument_parser
+# ?
```
yermulnik, line 12: The meaning of `# ?` isn't really clear =( Better to leave human-parsable comments please =)

tests/pytest/test__common.py
```diff
@@ -0,0 +1,275 @@
...
+
+# ?
+# ? per_dir_hook  # TODO: Requires `terraform_fmt` to be implemented
+# ?
+# @pytest.fixture
+# def mock_per_dir_hook_unique_part(mocker):
+#     return mocker.patch('pre_commit_terraform.terraform_fmt.per_dir_hook_unique_part')
+
+
+# def test_per_dir_hook_empty_files(mock_per_dir_hook_unique_part):
```
yermulnik, line 55: It's probably best — for the sake of posterity — to leave comments to disabled functions of why they are commented out.

.github/CONTRIBUTING.md
```diff
@@ -154,3 +155,27 @@ You can use [this PR](https://github.com/antonbabenko/pre-commit-terraform/pull/
...
+pytest -vv
+```
+
+To run and debug hooks as CLI:
+
+1. Create [`venv`](https://docs.python.org/3/library/venv.html) and activate it, IE by `virtualenv venv`
+2. Inside virtual env install pre-commit-terraform as package:
+	```bash
+	pip install --editable .
+	```
```
webknjaz, line 176: #742 already covers this in a better way. It's best to avoid steering people towards calling the underlying low-level tooling directly.

.pre-commit-config.yaml
```diff
@@ -61,8 +61,99 @@ repos:
 
 # JSON5 Linter
 - repo: https://github.com/pre-commit/mirrors-prettier
-  rev: v3.1.0
+  rev: v4.0.0-alpha.8
```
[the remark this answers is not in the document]
webknjaz: I think this is the output of `pre-commit autoupdate` which doesn't care about the tag semantics and just pulls in whatever's latest in the Git repo. I'm not sure if you can restrict it. But also, I probably wouldn't do this in a combined PR anyway.

.pre-commit-config.yaml
```diff
@@ -61,8 +61,99 @@ repos:
...
+
+- repo: https://github.com/astral-sh/ruff-pre-commit
+  rev: v0.8.4
+  hooks:
+    - id: ruff
+      args: [--fix]
+      types_or: [python, pyi]
+    - id: ruff-format
+      types_or: [python, pyi]
+      args: [--config, format.quote-style = 'single', --config, line-length = 100]
```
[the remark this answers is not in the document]
webknjaz: I also have this sort of comments in a draft review that I haven't yet posted in #743.

On the CLI settings, they should almost never be used not because of the defaults but because any other calls to those tools (like IDE extensions automatically picking them up w/o the user configuring anything) would not consume these settings and will result in different sets of violations reported and different behaviors of tools that aren't linters.

And for consistency, I'd rather enforce `yamllint`. I can send in my default config separately.

Additionally, the organization of integrations in this config is suboptimal that I'll expand on in the other review.

.pre-commit-config.yaml
```diff
@@ -61,8 +61,99 @@ repos:
...
+  rev: 5.13.2
+  hooks:
+    - id: isort
+      name: isort
+      args: [--force-single-line, --profile=black]
+      exclude: |
+        (?x)
+          # It set wrong indent in imports in place where it shouldn't
+          # https://github.com/PyCQA/isort/issues/2315#issuecomment-2566703698
+          (^tests/pytest/test__cli_subcommands.py$
```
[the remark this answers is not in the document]
webknjaz: The correct fix is to avoid runtime imports, though.

.pre-commit-hooks.yaml
```diff
@@ -37,7 +37,7 @@
   name: Terraform docs (overwrite README.md)
   description: Overwrite content of README.md with terraform-docs.
   require_serial: true
-  entry: python -Im pre_commit_terraform replace-docs
+  entry: python -Im pre_commit_terraform terraform_docs_replace
```
webknjaz, line 40: Plz, stick to kebab-case CLI arguments. It's rather uncommon to use snake_case in CLIs.

.vscode/settings.json
```diff
@@ -7,4 +7,7 @@
     "code-block-style": false
   },
   "markdown.validate.enabled": true,
+  "python.analysis.extraPaths": [
+    "./src"
```
webknjaz, line 11: should also check the tests dirs

hatch.toml
```diff
@@ -8,6 +9,9 @@ packages = [
   'src/pre_commit_terraform/',
 ]
 
+[build.targets.wheel.force-include]
+'.pre-commit-hooks.yaml' = 'pre_commit_terraform/_artifacts/.pre-commit-hooks.yaml'
```
webknjaz, line 13: This should probably just be picked into master directly.

pyproject.toml
```diff
@@ -9,16 +9,15 @@ build-backend = 'hatchling.build'
 name = 'pre-commit-terraform'
 classifiers = [
   'License :: OSI Approved :: MIT License',
-  'Programming Language :: Python :: 2',
-  'Programming Language :: Python :: 2.7',
   'Programming Language :: Python :: 3',
-  'Programming Language :: Python :: 3.6',
-  'Programming Language :: Python :: 3.7',
+  'Programming Language :: Python :: 3 :: Only',
```
webknjaz, line 13: It's common to list which environments it's being tested under. And additionally, the most important bit in the metadata would be the `requires-python` value. The classifiers are for humans to read, but the Python requirement is for the dependency resolver to actually take into account.

.pre-commit-hooks.yaml
yermulnik: +1

.pre-commit-config.yaml
```diff
@@ -61,8 +61,99 @@ repos:
...
+      - --disable=R0801 # Similar lines in 2 files. Currently I don't think that it possible to DRY hooks unique files boilerplate
+
+      exclude: test_.+\.py$
+
+- repo: https://github.com/pre-commit/mirrors-mypy
+  rev: v1.14.1
+  hooks:
+    - id: mypy
+      additional_dependencies:
+          - types-PyYAML
```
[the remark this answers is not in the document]
webknjaz: I'd even go as far as, to match one of the existing sequence styles. But really, it's best to make a separate PR with the yamllint integration before this one.
```suggestion
      - types-PyYAML
```

.pre-commit-config.yaml
```diff
@@ -61,8 +61,99 @@ repos:
...
+
+- repo: https://github.com/wemake-services/wemake-python-styleguide
+  rev: 1.0.0
+  hooks:
+    - id: wemake-python-styleguide
+      args:
+      - --allowed-module-metadata=__all__ # Default to ''
+      - --max-local-variables=6 # Default to 5
+      - --max-returns=6 # Default to 5
+      - --max-imports=15 # Default to 12
```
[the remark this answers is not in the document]
webknjaz: Don't accept this just yet. It shouldn't even be in this config. See my other comments.

.pre-commit-config.yaml
```diff
@@ -61,8 +61,99 @@ repos:
...
+          WPS226 <!-- Found string literal over-use - append > 3 -->
+
+
+      exclude: |
+        (?x)
+          # Ignore tests
+          (^tests/pytest/test_.+\.py$
+          # Ignore deprecated hook
+          |^src/pre_commit_terraform/terraform_docs_replace.py$
+        )
```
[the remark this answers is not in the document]
webknjaz: A better style would be
```suggestion
        ^(
          tests/pytest/test_[_\w]+\.py
          # Deprecated hook:
          |src/pre_commit_terraform/terraform_docs_replace\.py
        )$
```

But I would say that these shouldn't be in this config but in the flake8 config. Also, it's a bad idea not to lint the tests as they are code too.

Anyway, my PR #742 adds the proper structure where exclusions may exist in a more granular form.

pytest.ini
```diff
@@ -0,0 +1,5 @@
+[pytest]
+pythonpath = src
```
webknjaz, line 2: This is not needed. And also, I've got a proper config in #742 anyway.

src/pre_commit_terraform/__main__.py
```diff
@@ -1,9 +1,9 @@
 """A runpy-style CLI entry-point module."""
 
-from sys import argv, exit as exit_with_return_code
-
-from ._cli import invoke_cli_app
+from sys import argv
```
webknjaz, line 3: This will conflict with a fix in #742 FYI (this specific import is being dropped).

src/pre_commit_terraform/_cli.py
```diff
@@ -21,15 +19,14 @@ def invoke_cli_app(cli_args: list[str]) -> ReturnCodeType:
     root_cli_parser = initialize_argument_parser()
     parsed_cli_args = root_cli_parser.parse_args(cli_args)
 
-    try:
+    try:  # noqa: WPS225 - Found too many `except` cases: 4 > 3
```
webknjaz, line 22: Yes! This is the correct way of suppressing rules for legitimate cases granularly :)

tests/pytest/test__common.py
webknjaz: Better yet — use `@pytest.mark.skip(reason='Could not figure it out, because X. Details in #666')`

webknjaz: FWIW, the purpose of comments in code is not to add commented out source code, but to contain explanations of motivations and really unobvious things.

If linting wasn't disabled, WPS would've caught this with `flake8-erradicate` that it has built-in. This is why any exclusions should go through the flake8 mechanisms, accompanied by code comments with motivation/justification.

src/pre_commit_terraform/terraform_docs_replace.py
webknjaz: While correct, and I noticed this earlier, I'd say this is unrelated to this specific PR.

src/pre_commit_terraform/_cli_parsing.py
```diff
@@ -6,7 +6,50 @@
...
-from ._cli_subcommands import SUBCOMMAND_MODULES
+from pre_commit_terraform._cli_subcommands import SUBCOMMAND_MODULES
+
+
+def populate_common_argument_parser(parser: ArgumentParser) -> None:
+    """
+    Populate the argument parser with the common arguments.
+
+    Args:
+        parser (argparse.ArgumentParser): The argument parser to populate.
```
webknjaz, line 17: This API doc style is non-native to Sphinx. You might have to specify it in `.darglint`. But beware that darglint is deprecated and its parsing of non-native docstrings is known to be extremely slow. I once debugged it to a single Python file being linted during 11 **minutes** when I forgot to specify that I wanted the native one.

Here's the native syntax FYI: https://www.sphinx-doc.org/en/master/usage/domains/python.html#info-field-lists.

I haven't checked if WPS replaced darglint w/ something else, it might not be as bad.

Also, with some Sphinx extensions, there shouldn't be a need to duplicate the argument types in docstrings. Plus, since you don't have Sphinx, this is definitely not something you should be spending additional time on. Just describe the args and the return value, omitting the types as it's not worth it.

src/pre_commit_terraform/_cli_parsing.py
```diff
@@ -18,19 +61,27 @@ def attach_subcommand_parsers_to(root_cli_parser: ArgumentParser, /) -> None:
...
     subcommand_parsers = root_cli_parser.add_subparsers(
         dest='check_name',
-        help='A check to be performed.',
         required=True,
     )
     for subcommand_module in SUBCOMMAND_MODULES:
-        subcommand_parser = subcommand_parsers.add_parser(subcommand_module.CLI_SUBCOMMAND_NAME)
+        subcommand_parser = subcommand_parsers.add_parser(
+            subcommand_module.HOOK_ID,
+            add_help=False,
```
webknjaz, line 69: I still maintain that it'd be useful to allow `--help` in subcommands.

src/pre_commit_terraform/_cli_parsing.py
```diff
@@ -18,19 +61,27 @@ def attach_subcommand_parsers_to(root_cli_parser: ArgumentParser, /) -> None:
...
+        populate_common_argument_parser(subcommand_parser)
+        subcommand_module.populate_hook_specific_argument_parser(subcommand_parser)
 
 
 def initialize_argument_parser() -> ArgumentParser:
-    """Return the root argument parser with sub-commands."""
+    """
+    Parse the command line arguments and return the parsed arguments.
+
+    Return the root argument parser with sub-commands.
```
webknjaz, line 82: Only the docstring title needs to be in imperative mood. The long description paragraph doesn't need to do that, it can be more explanatory.

src/pre_commit_terraform/_cli_subcommands.py
```diff
@@ -1,12 +1,9 @@
 """A CLI sub-commands organization module."""
 
-from . import terraform_docs_replace
-from ._types import CLISubcommandModuleProtocol
+from pre_commit_terraform import terraform_docs_replace
```
webknjaz, line 3: FTR #742 has proper configuration that can work with relative imports just fine.

src/pre_commit_terraform/_common.py
```diff
@@ -0,0 +1,184 @@
...
+
+    Returns:
+        dict[str, str]: A dictionary mapping variable names to their corresponding values.
+    """
+    env_var_dict = {}
+
+    for env_var_str in env_var_strs:
+        name, value = env_var_str.split('=', 1)  # noqa: WPS110 # 'value' is valid var name here
+        if value.startswith('"') and value.endswith('"'):
+            value = value[1:-1]  # noqa: WPS110 # 'value' is valid var name here
```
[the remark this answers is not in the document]
webknjaz: This is not equivalent, though. It can turn `'""""""'` into `''` while `'""""""'[1:-1] == '""""'`. FWIW, this implementation has a lot of corner cases potentially, and I'd look into bringing in a third-party library that would have a more well-thought way of processing such strings.

src/pre_commit_terraform/_common.py
webknjaz: @yermulnik the convention is that a single leading `_` makes things “private”. They are still accessible, but it's common to use this marker. When a module has `__all__` declared, `from mod import *` imports all the things listed there (not that asterisk imports are a good idea anyway). But when it doesn't, it only imports things that are non-private. This is where it has an impact. Some analysis tools would also treat the underscored things as private. Some linters wouldn't demand docstrings in them, for example.

src/pre_commit_terraform/_common.py
```diff
@@ -0,0 +1,184 @@
...
+    Returns:
+        Set of unique directories.
+    """
+    unique_dirs = set()
+
+    for file_path in files:
+        dir_path = os.path.dirname(file_path)
+        unique_dirs.add(dir_path)
+
+    return unique_dirs
```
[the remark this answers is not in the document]
webknjaz: @yermulnik you're almost correct, but your suggestion uses a *list* comprehension that produces an in-memory throw-away list object. It's better to skip this step and go straight for a generator expression that doesn't do such thing.
```suggestion
    return set(os.path.dirname(path) for path in files)
```

(another improvement would be to stop using strings for paths and switch to `pathlib.Path` everywhere)

.pre-commit-config.yaml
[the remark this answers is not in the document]
yermulnik: > A better style would be

I'm just worrying about extraneous spaces

tests/pytest/test__common.py
yermulnik: > FWIW, the purpose of comments in code is not to add commented out source code, but to contain explanations of motivations and really unobvious things.

+1

src/pre_commit_terraform/terraform_docs_replace.py
yermulnik: Not sure I'm following what you mean =) It's just a commit suggestion for this PR to not glue text together but to inject a whitespace between sentences =)

src/pre_commit_terraform/_common.py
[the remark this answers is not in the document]
yermulnik: > This is not equivalent, though

Yep. It's what I meant with comment in parens though.

Looking at this yet another time, I do agree we should strip only equal number of double quotes from both ends. Resolving this thread. @MaxymVlasov Please discard this commit suggestion.

src/pre_commit_terraform/_common.py
yermulnik: @webknjaz Thanks a lot for explanation 🤝 Appreciate the detailed outline so that I personally learn a lot from your code and comments 👍🏻 
@MaxymVlasov Please disregard my initial comment if this function is meant to be private. Thanks.

src/pre_commit_terraform/terraform_docs_replace.py
webknjaz: Since this line was never modified in the PR, it's out of the scope.

yermulnik: The line(s) below this one were modified. Though for the consistency's sake adding whitespace makes sense at this line rather than at the one below that was changed.

yermulnik: W/o this whitespace the message is going to have no space between this and the other sentence.
I have no preference on where this change is going to be applied: here or at any other PR 🤷🏻

yermulnik: It's actually a thing from somewhere else, but since this PR is touching upon this bit, why don't we fix it at once 🤔

webknjaz: Because it'll create a conflict with #742, for example. Plus, since this module is going to be deleted at some point anyway, I wouldn't care about it at all.

yermulnik: If it's going to introduce conflict, then we should not push similar or alike changes by means of different PRs I guess 😺 
Either way pushing to `master` apparently incorrect code bits isn't great I guess. Even if it's going to be removed at some non-definitive point in time.
Up to you and Max though.

webknjaz: Yeah, I expect that #742 gets it before any other PR so that introduction of the linters would include relevant fixes atomically.

.github/CONTRIBUTING.md
MaxymVlasov: Ok, I'll remove it from this PR

.pre-commit-config.yaml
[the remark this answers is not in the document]
MaxymVlasov: >this is the output of `pre-commit autoupdate`

Yep.

>I'm not sure if you can restrict it. 

Can be done by using `renovate` for this, not  `pre-commit autoupdate` itself.

Anyway, I'll revert it for now
```suggestion
  rev: v3.1.0
```

.pre-commit-config.yaml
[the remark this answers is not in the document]
MaxymVlasov: It wasn't work w/o them.
There should be a solution, but I didn't find it quickly and that on current iteration I'm okay to have "at least somehow working tests" - theier not optimal definition shoudn't slow down migration process, otherwise we will migrate to Python for whole year

.pre-commit-config.yaml
```diff
@@ -61,8 +61,99 @@ repos:
...
+  hooks:
+    - id: isort
+      name: isort
+      args: [--force-single-line, --profile=black]
+      exclude: |
+        (?x)
+          # It set wrong indent in imports in place where it shouldn't
+          # https://github.com/PyCQA/isort/issues/2315#issuecomment-2566703698
+          (^tests/pytest/test__cli_subcommands.py$
+        )
```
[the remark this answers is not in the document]
MaxymVlasov: Yes.
They full construction (if there will be more exclusions) will be looks like
```yaml
(?x)
  (regex1
  |regex2
  |regexN
)
```

.pre-commit-hooks.yaml
MaxymVlasov: I want to maintain one name across everything to not make create more entropy.

If hook has named `terraform_docs_replace` it MUST be called this way everywhere, no matter what.

src/pre_commit_terraform/_cli_parsing.py
MaxymVlasov: Interesting, what google use them for docs
https://google.github.io/styleguide/pyguide.html#383-functions-and-methods

src/pre_commit_terraform/_cli_parsing.py
```diff
@@ -6,7 +6,50 @@
...
+        metavar='KEY=VALUE',
+        help='Arguments that configure hook behavior',
+        default=[],
+    )
+    parser.add_argument(
+        '-i',
+        '--tf-init-args',
+        '--init-args',
+        action='append',
+        help='Arguments for `tf init` command',
```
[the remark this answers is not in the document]
MaxymVlasov: There could be not only `terraform`, but `tofu`, `terraform-alpha-1.11.2025.99.99` etc

src/pre_commit_terraform/_cli_parsing.py
MaxymVlasov: Agree, but that's out of scope for now - firstly we need to reimplement same hooks in Python w/o any breaking changes about what we are aware. 

Anyway, I added that as TODO in `rewrite_hooks_on_python` branch

src/pre_commit_terraform/_run_on_whole_repo.py
MaxymVlasov: Yeah, looks like I forgot to remove code after debugging. Good catch

src/pre_commit_terraform/_run_on_whole_repo.py
```diff
@@ -0,0 +1,103 @@
...
+
+    Raises:
+        ValueError: If the hook ID is not found in the .pre-commit-hooks.yaml file.
+    """
+    logger.debug('Hook ID: %s', hook_id)
+
+    # Get the directory containing the packaged `.pre-commit-hooks.yaml` copy
+    artifacts_root_path = access_artifacts_of('pre_commit_terraform') / '_artifacts'
+    pre_commit_hooks_yaml_path = artifacts_root_path / '.pre-commit-hooks.yaml'
+    pre_commit_hooks_yaml_path.read_text(encoding='utf-8')
```
MaxymVlasov, line 56:
```suggestion
```

tests/pytest/test__cli_parsing.py
MaxymVlasov: Just change highlighting for everyone who use this extension
https://marketplace.visualstudio.com/items?itemName=aaron-bond.better-comments

.pre-commit-config.yaml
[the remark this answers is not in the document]
MaxymVlasov: About
`^( ... )$` - you not always have full path specified, sometime you want to ignore directories etc.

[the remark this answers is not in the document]
MaxymVlasov:
```suggestion
          (^tests/pytest/test_.+\.py$
          # Ignore deprecated hook
          |^src/pre_commit_terraform/terraform_docs_replace\.py$
        )
```

.pre-commit-config.yaml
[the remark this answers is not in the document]
yermulnik: Oh, so the `(?x)` I guess is to ignore whitespaces in the pattern, so we should be good with something like this (or similar):
```
(?x)
^(
  regex1|
  regex2|
  regexN
)$
```

src/pre_commit_terraform/_cli_parsing.py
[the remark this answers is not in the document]
yermulnik: Then let's abstract the name, like f.e.:
```suggestion
        help="Arguments to tool's `init` sub-command",
```

tests/pytest/test__cli_parsing.py
yermulnik: Ah, yeah, I seem to recall that you already redirected me to something like this before. Though... I don't use VSCode and we all often review and view code via e.g. GH Web UI which isn't aware of this capability, specific to VSCode only(?):
изображение
The https://github.com/aaron-bond/better-comments/blob/master/README.md has no cross-refs to any other IDEs, editors that support this "feature", or common conventions with regards to this capability.
I know of pretty common `FIXME`, `TODO`, and `XXX` (with less common `NOTE`, `DEBUG`, `OPTIMIZE`, `HACK`, and `BUG`) keywords, which are highlighted more or less in all more or less common IDEs and editors.
Though the best solution here would be to leave a human-oriented brief comment so that others can figure out what author meant.

> [!TIP]
> 💡«_You’re writing code for others to read/maintain. It’s secondary that computers can interpret it_»

tests/pytest/test__common.py
yermulnik: > L48 - # TODO: Requires `terraform_fmt` to be implemented
> 
> This function can't be tested till actual hooks will be introduced

Please add the `# TODO: Requires `terraform_fmt` to be implemented` to each commented out code block here.
It's not nitpicking. I actually didn't get that that comment for `def mock_per_dir_hook_unique_part(mocker)` function does relate to all the rest of commented out functions.
Thanks.

.pre-commit-config.yaml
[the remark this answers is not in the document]
yermulnik: > About `^( ... )$` - you not always have full path specified, sometime you want to ignore directories etc.

Yep, that makes sense. I optimized the existing regex. Can do it later, when we have more items in regex.
Either way you always can add something like `.+` like this:
```shell
^(some/full/path|another/ful/one|.+/some/path/with/anything/in/the/beginning|some/path/with/anything/at/the/end/.+)$
```

src/pre_commit_terraform/_cli_parsing.py
webknjaz: It's a custom style that's not enabled by default. It's implemented by the napoleon extension: https://sphinxcontrib-napoleon.rtfd.io / https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html.

.pre-commit-hooks.yaml
webknjaz: That's the thing — the only place for the hook subcommand name is the constant in the module, not the module itself. `terraform_docs_replace.py` is just a Python module that backs it, not a hook. The hook name would be `terraform-docs-replace`. This is a common convention, and the underlying implementation details shouldn't leak into UI/UX.

Here's some discussion in other popular CLI frameworks: https://github.com/fastapi/typer/issues/341

webknjaz: Here's some public arguments discussing the ergonomics of having to use Shift-key vs. not, leaning towards kebab-space or no-delimiter:
* https://devcenter.heroku.com/articles/cli-style-guide#naming-the-command
* https://smallstep.com/blog/the-poetics-of-cli-command-names/
* https://github.com/leemunroe/cli-style-guide?tab=readme-ov-file#naming-commands

Do you have many programs on your system that have `snake_case` CLIs? Could you name a few?

src/pre_commit_terraform/_run_on_whole_repo.py
webknjaz: No, it's not a shell alias or a function because the subprocess is called without `shell=True`. It's a direct invocation. It uses whatever's first in the passed iterable as the executable to launch: https://github.com/python/cpython/blob/953b49e5/Lib/subprocess.py#L1822C13-L1831C37. It basically calls `posix_spawn()`: https://man7.org/linux/man-pages/man3/posix_spawn.3.html. Though, it doesn't seem to use `$PATH`-based lookup according to the man page (unlike `posix_spawnp()`).

src/pre_commit_terraform/_run_on_whole_repo.py
```diff
@@ -0,0 +1,102 @@
...
+    if excluded_pattern:
+        all_file_paths_that_can_be_checked = [
+            file_path
+            for file_path in git_ls_file_paths
+            if included_pattern.search(file_path) and not excluded_pattern.search(file_path)
+        ]
+    else:
+        all_file_paths_that_can_be_checked = [
+            file_path for file_path in git_ls_file_paths if included_pattern.search(file_path)
+        ]
```
webknjaz, line 91: Instead of duplicating this, just conditionally make `exclude_patterns` never matching. Or alternatively, conditionally use a no-op function instead of calling it. This would greatly collapse the structs.
Also, you could do something like this
```suggestion
    all_file_paths_that_can_be_checked = (
        file_path for file_path in git_ls_file_paths
        if included_pattern.search(file_path)
    )
    if excluded_pattern:
        all_file_paths_that_can_be_checked = (
            file_path for file_path in all_file_paths_that_can_be_checked
            if not excluded_pattern.search(file_path)
        )
    all_file_paths_that_can_be_checked = list(all_file_paths_that_can_be_checked)
```
Though it could be
```suggestion
    all_file_paths_that_can_be_checked = [
        file_path for file_path in git_ls_file_paths
        if included_pattern.search(file_path)
        and (not excluded_pattern or not excluded_pattern.search(file_path))
    ]
```

src/pre_commit_terraform/_types.py
```diff
@@ -1,28 +1,27 @@
...
-ReturnCodeType = ReturnCode | int
+ReturnCodeType: TypeAlias = ReturnCode | int
 
 
+@runtime_checkable
 class CLISubcommandModuleProtocol(Protocol):
     """A protocol for the subcommand-implementing module shape."""
 
-    CLI_SUBCOMMAND_NAME: Final[str]
+    HOOK_ID: str
```
webknjaz, line 18: Technically, it's not a hook but a subcommand. You're mapping it to a pre-commit hook so in the pre-commit config, the entire command is a hook. But on this layer, it's just a subcommand and nothing more.

src/pre_commit_terraform/terraform_docs_replace.py
```diff
@@ -1,77 +1,95 @@
...
+from argparse import ArgumentParser
+from argparse import Namespace
 from typing import Final
 
-from ._structs import ReturnCode
-from ._types import ReturnCodeType
+from pre_commit_terraform._structs import ReturnCode
+from pre_commit_terraform._types import ReturnCodeType
+
+HOOK_ID: Final[str] = __name__.rpartition('.')[-1]
```
webknjaz, line 13:
```suggestion
HOOK_ID: Final[str] = __name__.rpartition('.')[-1].replace('_', '.')
```

src/pre_commit_terraform/_common.py
```diff
@@ -0,0 +1,184 @@
...
+    Args:
+        env_var_strs (list[str]): A list of environment variable strings in the format "name=value".
+
+    Returns:
+        dict[str, str]: A dictionary mapping variable names to their corresponding values.
+    """
+    env_var_dict = {}
+
+    for env_var_str in env_var_strs:
+        name, value = env_var_str.split('=', 1)  # noqa: WPS110 # 'value' is valid var name here
```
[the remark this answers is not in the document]
webknjaz: This could be a generator expression too, FWIW. No need for a throw-away list on each iteration.

src/pre_commit_terraform/_common.py
```diff
@@ -0,0 +1,178 @@
...
```
webknjaz, line 30: > `'value' is valid var name here`

Not really, it could also be `env_var_val`. But whatever `env_var_str.split('=', 1)` returns is a value. Also, those arguments `'='` and `1` are values. Each of the variables `name` and `value` get assigned different values, unpacked from the value returned by that function that you call and pass values in. Is `value` valid? I don't think there's a value in calling a variable `value` instead of just fixing the value of its name and dropping the ignore :man_shrugging:
Honestly, it's no better than using one-latter vars:
```python
        j, k = env_var_str.split('=', 1)
```
No difference, really. And that's what WPS110 is about.

src/pre_commit_terraform/_common.py
```diff
@@ -0,0 +1,178 @@
...
+    Returns:
+        The arguments with expanded environment variables.
+    """
+    expanded_args = []
+
+    for arg in args:
+        for env_var_name, env_var_value in env_vars.items():
+            if f'${{{env_var_name}}}' in arg:
+                logger.info('Expanding ${%s} in "%s"', env_var_name, arg)
+                arg = arg.replace(f'${{{env_var_name}}}', env_var_value)
```
webknjaz, line 70: Consider not reinventing the wheel:
* https://docs.python.org/3/library/string.html#template-strings
* https://pypi.org/p/expandvars

src/pre_commit_terraform/_common.py
```diff
@@ -0,0 +1,178 @@
...
+    """
+    # consume modified files passed from pre-commit so that
+    # hook runs against only those relevant directories
+    unique_dirs = _get_unique_dirs(files)
+
+    tf_path = get_tf_binary_path(hook_config)
+
+    logger.debug(
+        'Iterate per_dir_hook_unique_part with values:'
+        + '\ntf_path: %s\nunique_dirs: %r\nargs: %r\nenv_vars: %r',
```
webknjaz, line 106: explicit concat is slow, remember?

src/pre_commit_terraform/_common.py
```diff
@@ -0,0 +1,178 @@
...
+        unique_dirs,
+        args,
+        env_vars,
+    )
+    final_exit_code = 0
+    for dir_path in unique_dirs:
+        exit_code = per_dir_hook_unique_part(tf_path, dir_path, args, env_vars)
+
+        if exit_code != 0:
+            final_exit_code = exit_code
```
webknjaz, line 117: How about reporting the number of failures through the return code?
```suggestion
        final_exit_code += exit_code
```

src/pre_commit_terraform/_common.py
```diff
@@ -0,0 +1,178 @@
...
+
+    return final_exit_code
+
+
+class BinaryNotFoundError(Exception):
```
webknjaz, line 122: This would be useful in a dedicated module. And it should probably inherit a `LookupError`, maybe together with something else:
```suggestion
class BinaryNotFoundError(FileNotFoundError, LookupError):
```

src/pre_commit_terraform/_common.py
webknjaz: and drop explicit concat

src/pre_commit_terraform/_common.py
webknjaz, line 1: In general, it's a bad structural decision to have a soup-mixture style module with just about any arbitrary low-level thing from the project. It's better to have more specific names that would represent what the module contains. Most of the modules in the project expose common functions, only the outer layer of the app doesn't re-export them, really.

Give this module a narrow purpose and stick with it. Then move out anything that doesn't fit into the category into other modules with names that have some semantical meaning beyond being a generic English word that means whatever the reader would like to guess it means.

Honestly, I'd hoped WPS100 would catch this, but that check is rather dumb right now: https://wemake-python-styleguide.rtfd.io/en/latest/pages/usage/violations/naming.html#wemake_python_styleguide.violations.naming.WrongModuleNameViolation.

src/pre_commit_terraform/_common.py
```diff
@@ -0,0 +1,178 @@
...
+                logger.info('Expanding ${%s} in "%s"', env_var_name, arg)
+                arg = arg.replace(f'${{{env_var_name}}}', env_var_value)
+                logger.debug('After ${%s} expansion: "%s"', env_var_name, arg)
+
+        expanded_args.append(arg)
+
+    return expanded_args
+
+
+def per_dir_hook(
```
webknjaz, line 78: This name doesn't look like a function. There's no action in the name, no verb. It looks like a name of a constant and should be improved to look like something that does things and not just exists.

src/pre_commit_terraform/_logger.py
```diff
@@ -0,0 +1,87 @@
+"""Logs-related functions."""
+
+import logging
+import os
+from copy import copy
+
+
+class ColoredFormatter(logging.Formatter):
```
webknjaz, line 8: Integrate Rich instead.

src/pre_commit_terraform/_run_on_whole_repo.py
webknjaz, line 1: This `whole` thing reads weird. Perhaps you meant `entire`?

src/pre_commit_terraform/_run_on_whole_repo.py
```diff
@@ -0,0 +1,102 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def is_function_defined(func_name: str, scope: dict) -> bool:
+    """
+    Check if a function is defined in the global scope.
+
+    Args:
+        scope (dict): The scope (usually globals()) to check in.
```
webknjaz, line 18: If you're ever going to integrate Sphinx, beware that it's expected that you use RST syntax in docstrings:
```suggestion
        scope (dict): The scope (usually ``globals()``) to check in.
```
(this is a simplistic example that would probably be `` :py:func:`globals` `` IRL)

src/pre_commit_terraform/_run_on_whole_repo.py
```diff
@@ -0,0 +1,102 @@
...
+        'Checking if "%s":\n1. Defined in hook: %s\n2. Is callable: %s',
+        func_name,
+        is_defined,
+        is_callable,
+    )
+
+    return is_defined and is_callable
+
+
+def is_hook_run_on_whole_repo(hook_id: str, file_paths: list[str]) -> bool:
```
webknjaz, line 37: (entire)

src/pre_commit_terraform/_types.py
```diff
@@ -1,28 +1,27 @@
...
 
-from ._structs import ReturnCode
+from pre_commit_terraform._structs import ReturnCode
 
-
-ReturnCodeType = ReturnCode | int
+ReturnCodeType: TypeAlias = ReturnCode | int
 
 
+@runtime_checkable
```
webknjaz, line 14: Any reason to do this in this PR?

src/pre_commit_terraform/terraform_docs_replace.py
webknjaz: Oh, I didn't originally notice that Max added slower explicit concat in the following lines. Now I understand why you thought it was related. Those should really be reverted.

tests/pytest/test___main__.py
```diff
@@ -0,0 +1,44 @@
...
+        'pre_commit_terraform._cli.invoke_cli_app',
+        return_value=mock_return_code,
+    )
+    mock_exit = mocker.patch('sys.exit')
+
+    # Reload the module to trigger the main logic
+    if 'pre_commit_terraform.__main__' in sys.modules:
+        importlib.reload(sys.modules['pre_commit_terraform.__main__'])
+    else:
+        import pre_commit_terraform.__main__  # noqa: F401
```
webknjaz, line 20: Nooooooo! Don't do this! Use `runpy` like you're supposed to. This module is to be invoked as an entry-point and not in the middle of runtime. So in general you should've instead just called it through `sys.executable` and checked the return code + the output. No need to patch stuff in this case.

tests/pytest/test___main__.py
```diff
@@ -0,0 +1,44 @@
+import importlib
+import sys
+
+
+def test_main_success(mocker):
+    mock_argv = ['__main__.py', 'arg1', 'arg2']
+    mock_return_code = 0
+
+    mocker.patch('sys.argv', mock_argv)
+    mock_invoke_cli_app = mocker.patch(
```
webknjaz, line 10: Also, you could do a `--help` instead of patching most of the logic out. It would give you your successful return code.

tests/pytest/test__cli_parsing.py
webknjaz: Exactly, this is unmaintainable. And it's right that any code base is read much more often than written. So you should optimize for readability and maintainability in general. Not for an evil-corp-specific editor :man_shrugging:

If we didn't need this, we'd just write in binary.

tests/pytest/test__cli_parsing.py
```diff
@@ -0,0 +1,223 @@
...
+            'file1',
+            'file2',
+        ],
+    )
+
+    assert args.args == ['arg1', 'arg2']
+    assert args.hook_config == ['hook1', 'hook2']
+    assert args.tf_init_args == ['init1', 'init2']
+    assert args.env_vars_strs == ['env1', 'env2']
+    assert args.files == ['file1', 'file2']
```
webknjaz, line 69: These are actually a single test, and it should be a single test function. To run it multiple times, use `@pytest.mark.parametrize`. Don't forget to specify test IDs for better UX/DX in test report and invocation.

tests/pytest/test__cli_parsing.py
```diff
@@ -0,0 +1,223 @@
...
+
+from pre_commit_terraform._cli_parsing import attach_subcommand_parsers_to
+from pre_commit_terraform._cli_parsing import initialize_argument_parser
+from pre_commit_terraform._cli_parsing import populate_common_argument_parser
+
+
+# ?
+# ? populate_common_argument_parser
+# ?
+def test_populate_common_argument_parser(mocker):
```
webknjaz, line 13: Also, why didn't you write a docstring, so I would know what you're testing/expecting here?

tests/pytest/test__cli_parsing.py
webknjaz, line 1: In general, I see a huge overuse of mocking here. It's best to avoid it entirely where possible. Or only have a few exceptions. Designing runtime code well (like through dependency injection) allows for testing without mocking random things out.

src/pre_commit_terraform/_run_on_whole_repo.py
yermulnik: Oh, I didn't look at how this works under the hood. Thanks for the pointers. Seem like it defo doesn't take shell aliases and all that kind of stuff into account. Thanks!

tests/pytest/test__cli_subcommands.py
```diff
@@ -0,0 +1,23 @@
...
+    mock_terraform_docs_replace = mocker.patch('pre_commit_terraform.terraform_docs_replace')
+
+    mock_subcommand_modules = (mock_terraform_docs_replace,)
+
+    mocker.patch(
+        'pre_commit_terraform._cli_subcommands.SUBCOMMAND_MODULES',
+        mock_subcommand_modules,
+    )
+
+    from pre_commit_terraform._cli_subcommands import (
```
webknjaz, line 15: Do not add imports within functions. It's a *really* bad idea that makes runtime less predictable / more difficult to interpret. Also, it's already imported at the top, it's not going to load the file for the second time, it's already cached. Don't disable the linters that report that this is dumb.

tests/pytest/test__common.py
```diff
@@ -0,0 +1,275 @@
+# pylint: skip-file
+import os
+from os.path import join
```
webknjaz, line 3: Double import. Don't. Just don't. Enable the linters and fix the problems instead of showing this to other humans.

tests/pytest/test__common.py
```diff
@@ -0,0 +1,275 @@
...
+from pre_commit_terraform._common import _get_unique_dirs
+from pre_commit_terraform._common import expand_env_vars
+from pre_commit_terraform._common import get_tf_binary_path
+from pre_commit_terraform._common import parse_env_vars
+
+
+# ?
+# ? get_unique_dirs
+# ?
+def test_get_unique_dirs_empty():
```
webknjaz, line 17: Docstring?

tests/pytest/test__common.py
webknjaz: > This function can't be tested till actual hooks will be introduced

Do write a test with an `xfail` marker. No need to comment it out.

tests/pytest/test__common.py
```diff
@@ -0,0 +1,275 @@
...
+def test_parse_env_vars_with_equal_sign_in_value():
+    env_var_strs = ['VAR1=value=1', 'VAR2=value=2']
+    result = parse_env_vars(env_var_strs)
+    assert result == {'VAR1': 'value=1', 'VAR2': 'value=2'}
+
+
+def test_parse_env_vars_with_empty_value():
+    env_var_strs = ['VAR1=', 'VAR2=']
+    result = parse_env_vars(env_var_strs)
+    assert result == {'VAR1': '', 'VAR2': ''}
```
webknjaz, line 181: parametrize

tests/pytest/test__common.py
```diff
@@ -0,0 +1,275 @@
...
+    env_vars = {'VAR1': 'value1'}
+    result = expand_env_vars(args, env_vars)
+    assert result == ['arg1', 'value1', '${VAR2}']
+
+
+def test_expand_env_vars_with_special_chars():
+    args = ['arg1', '${VAR_1}', 'arg3']
+    env_vars = {'VAR_1': 'value1'}
+    result = expand_env_vars(args, env_vars)
+    assert result == ['arg1', 'value1', 'arg3']
```
webknjaz, line 226: parametrize

tests/pytest/test__common.py
```diff
@@ -0,0 +1,275 @@
...
+    mocker.patch('shutil.which', return_value='/usr/local/bin/terraform')
+    result = get_tf_binary_path(hook_config)
+    assert result == '/usr/local/bin/terraform'
+
+
+def test_get_tf_binary_path_from_system_path_tofu(mocker):
+    hook_config = []
+    mocker.patch('shutil.which', side_effect=[None, '/usr/local/bin/tofu'])
+    result = get_tf_binary_path(hook_config)
+    assert result == '/usr/local/bin/tofu'
```
webknjaz, line 263: parametrize

tests/pytest/test__common.py
```diff
@@ -0,0 +1,275 @@
...
+# ?
+def test_get_tf_binary_path_from_hook_config():
+    hook_config = ['--tf-path=/custom/path/to/terraform']
+    result = get_tf_binary_path(hook_config)
+    assert result == '/custom/path/to/terraform'
+
+
+def test_get_tf_binary_path_from_pct_tfpath_env_var(mocker):
+    hook_config = []
+    mocker.patch.dict(os.environ, {'PCT_TFPATH': '/env/path/to/terraform'})
```
webknjaz, line 240: use `monkeypatch.setenv()`

tests/pytest/test__common.py
```diff
@@ -0,0 +1,275 @@
...
+
+def test_get_tf_binary_path_not_found(mocker):
+    hook_config = []
+    mocker.patch('shutil.which', return_value=None)
+    with pytest.raises(
+        BinaryNotFoundError,
+        match='Neither Terraform nor OpenTofu binary could be found. Please either set the "--tf-path"'
+        + ' hook configuration argument, or set the "PCT_TFPATH" environment variable, or set the'
+        + ' "TERRAGRUNT_TFPATH" environment variable, or install Terraform or OpenTofu globally.',
+    ):
```
webknjaz, line 274:
```suggestion
    error_msg = (
        r'^Neither Terraform nor OpenTofu binary could be found. Please either set the "--tf-path" '
        'hook configuration argument, or set the "PCT_TFPATH" environment variable, or set the '
        '"TERRAGRUNT_TFPATH" environment variable, or install Terraform or OpenTofu globally\.$'
    )
    with pytest.raises(BinaryNotFoundError, match=error_msg):
```

tests/pytest/test__errors.py
webknjaz, line 1: This module is rather pointless. Delete it.

tests/pytest/test__run_on_whole_repo.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+# ?
+# ? is_function_defined
+# ?
+def test_is_function_defined_existing_function():
+    def sample_function():
+        pass
+
+    scope = globals()
```
webknjaz, line 17: Wow, this is nasty! Messing up the entire process state for all the test functions, eh? Don't do this. At least copy the thing instead of side-effecting everything...
```suggestion
    scope = globals().copy()
```

tests/pytest/test__run_on_whole_repo.py
```diff
@@ -0,0 +1,108 @@
...
+    scope['sample_function'] = sample_function
```
webknjaz, line 18: But really.. Can't you just not go the unnecessarily complicated route? KISS.
```suggestion
    scope = {'sample_function': sample_function}
```

tests/pytest/test__run_on_whole_repo.py
```diff
@@ -0,0 +1,108 @@
...
+def test_is_function_defined_callable_object():
+    class CallableObject:
+        def __call__(self):
+            pass
+
+    callable_object = CallableObject()
+    scope = globals()
+    scope['callable_object'] = callable_object
+
+    assert is_function_defined('callable_object', scope) is True
```
webknjaz, line 46: Parametrize

tests/pytest/test__structs.py
webknjaz, line 1: This isn't very useful either. Delete.

tests/pytest/test__types.py
```diff
@@ -0,0 +1,36 @@
...
+
+    assert isinstance(MockSubcommandModule(), CLISubcommandModuleProtocol)
+
+    class InvalidSubcommandModule:
+        HOOK_ID = 'invalid_hook'
+
+        def populate_argument_parser(self, subcommand_parser: ArgumentParser) -> None:
+            pass
+
+    assert not isinstance(InvalidSubcommandModule(), CLISubcommandModuleProtocol)
```
webknjaz, line 36: This looks like two separate tests, they shouldn't be one.

tests/pytest/test__types.py
webknjaz, line 1: Honestly, this is merely checking that CPython stdlib works. I don't see it being useful.

.pre-commit-config.yaml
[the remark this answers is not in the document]
webknjaz: > About `^( ... )$` - you not always have full path specified, sometime you want to ignore directories etc.

The idea is that it shouldn't have any exclusions at all. Which is why I deem full paths as acceptable temporarily so that they could be removed from the list one-by-one. This shouldn't be a long-term suppression.

src/pre_commit_terraform/_run_on_whole_repo.py
webknjaz: Yep. `shell=True` is considered dangerous, though. Also, there's a security-focused linter called `bandit` that would warn on any subprocess use. For that, I'd consolidate all the calls to this library in a separate module, making sure that it's always called securely, and suppress the warnings there, disallowing any direct imports elsewhere.

I saw that `terraform_docs_replace.py` has this problem, but it's not exactly in the scope of the current PR, so I'm not bringing it up for now.

.pre-commit-config.yaml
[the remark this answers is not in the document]
webknjaz: @yermulnik `(?x)` also allows *inline* comments within the regex string, which can be very handy: https://docs.python.org/3/library/re.html#re.X.
Although, one has to remember to use `|` instead of `>` in YAML so that it won't merge it all into a single string, breaking the regex.

src/pre_commit_terraform/_common.py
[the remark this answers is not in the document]
yermulnik: Thanks. I updated my commit suggestion. Did I get you right?

src/pre_commit_terraform/_common.py
yermulnik: 👍🏻 

ps: I'm amazed with the way you put things straightly great 🌟

src/pre_commit_terraform/_common.py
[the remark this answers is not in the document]
webknjaz: Nope. You made it a set. Sets are unordered — their entries can be pretty much random. Plus, you're making a throw-away set in-memory, again. In your example, you're passing a generator expression into the set constructor, and then the control flow iterates over it to create said set. Then, the set is being unpacked into two individual variables.

Here's a barebones generator expression that gets unpacked into vars right away, without intermediary throw-away data structures:
```pycon
>>> a, b = (_ for _ in range(2))
>>> a
0
>>> b
1
```

P.S. If you wanted to have a *set* comprehension, it'd look like `{_ for _ in range(2)})` and is composable w/o an explicit `set()` call, since this syntax is built-in.

src/pre_commit_terraform/_run_on_whole_repo.py
yermulnik: While I support this kind of improvements overall, I guess @MaxymVlasov is attempting to map old to new one-to-one at this stage. It may help keeping track of "what was what" going path forward when we do code polishing later 🤞🏻

webknjaz: Ah, fair. I didn't know of the old names being a factor.

.pre-commit-config.yaml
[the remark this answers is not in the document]
yermulnik: Yeah, have never known about this cool capability. Thanks 👍🏻

src/pre_commit_terraform/_common.py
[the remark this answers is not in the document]
yermulnik: Ah, jeez, I knew I screwed it 🤣 Updated to be a generator finally. Appreciate assistance.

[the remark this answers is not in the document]
webknjaz: @yermulnik oh, I wasn't suggesting to use `_` as a var name. This was just a dummy REPL test for me as a demo. Meaningful var names are always better, even inline.
But if you're going the way of attempting to disregard the intermediate var, you could just as well return a `map()` iterator the FP way:

```suggestion
        name, value = map(str.strip, env_var_str.split('=', 1))
```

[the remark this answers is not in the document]
webknjaz: Yeah, I was never suggesting shadowing `str`. That'd be nasty. `str.strip(x)` is just a classmethod way of calling `x.strip()`.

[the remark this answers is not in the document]
yermulnik: Ah, ok, I might've misread you then =)

webknjaz:
@MaxymVlasov look into how the tests are structured in the test infra PR and follow that. Avoid mocks where possible. When we get to have a call, I'll show you how to inspect coverage locally to judge if you even need a test.

.gitignore
```diff
@@ -1 +1,4 @@
 tests/results/*
+__pycache__/
+*.py[cod]
+node_modules/
```
webknjaz, line 4: I don't see why this entry is needed. Could you post a justification?

.pre-commit-config.yaml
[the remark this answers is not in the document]
webknjaz: xref #747

.pre-commit-config.yaml
```diff
@@ -66,3 +66,94 @@ repos:
...
+
+- repo: https://github.com/pycqa/isort
+  rev: 5.13.2
+  hooks:
+    - id: isort
+      name: isort
+      args: [--force-single-line, --profile=black]
+      exclude: |
+        (?x)
+          # Uses incorrect indentation in imports in places where it shouldn't
```
webknjaz, line 93: FWIW, this ignore is irrelevant because your code defies best practices. It should be fixed instead of skipping the checks.

PR #742, opened 2025-01-03
ci: Set up Python-focused testing infra

webknjaz:
This patch integrates running pytest and pre-commit via tox.
It includes changes needed for tracking coverage automatically
for local inspection and reporting it to Codecov.

This is essentially a copy of my typical boilerplate with a
few more advanced hacks for communicating with GHA.

It can be invoked as `tox`, `tox r`, `tox -qq`,
`tox r -e pre-commit -qq -- mypy --all-files`,
`tox -qq -- -vv -s tests/pytest/'tests/pytest/_cli_test.py::test_known_interrupts[ctrl-c]'`
etc.

`tox` itself is installable or runnable ad-hoc via
`pipx` / `uvx` / `uv tool install` or one's favorite distro
package manager. The Python version of where it's installed is
not important. `tox` is not usually installed into regular
manually-managed virtualenvs. Mentioned tools maintain internal
venvs for isolating the apps and the system install would just
live side-by-side with other system stuff.

`tox` creates virtualenvs for its wrapped commands automatically
and keeps track of cache invalidation for them. The commands
wired via `tox` can either be invoked with the default arguments
or with the replacement passed to tox after the `--` separator.

Furthermore, the change introduces the MyPy integration into the
pre-commit configuration and is set up in strict mode.

This patch includes a config file for MyPy and adjusts the existing
Python modules to pass the checks on every other version down
to Python 3.9.

Additionally, a set of basic unit tests is added to reach the bar
of 100% code coverage metric, which is set as expected in the
configs.

Another patch included is making stderr interceptable. Previously,
`sys.stderr` was exposed as a module-scoped var, while the `capsys`
fixture patches the `stderr` attribute on the `sys` module. So such
patching did not influence the `stderr` variable in the `_cli` module
as it remained linked to the non-patched object.
With this change, the ability to inspect printing to stderr is
recovered and is then demonstrated in the enclosed tests.

Configuration note: each underlying tool must have its configuration
in a native file, it supports and *not* passed via CLI arguments in
the pre-commit or tox configs. This is important because those tools
may be invoked by things other than our direct automation, like
editor/IDE plugins or other platforms/tools. In order for other
invocations to work the same, the shared configuration must be the
same. For example, if we were to disable a rule in a pylint call in
pre-commit config, it wouldn't show in CI but would break in people's
editors that just call pylint directly and show errors inline.

- [ ] This PR fixes a bug.
- [x] This PR is internal maintenance, it's contributor-facing and not user-facing.

See above.

### How can we test changes

Install `tox` however you like (could also optionally attempt installing `tox-uv` with it, in the same env) and run `tox`. This will execute `pytest` under your current Python interpreter. You can also try things like `tox p -qq -e py313,py39,pre-commit`, `tox r -qq -e py313,py39,pre-commit`, `tox r -qq -e pre-commit -- mypy-py313 --all-files`, `tox r -qq -e pre-commit -- mypy --all-files`.

Keeping this as a draft to add some GHA automation before merging.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,49 @@
+---
+
+codecov:
+  notify:
+    after_n_builds: 21  # Number of test matrix+lint jobs uploading coverage
+    wait_for_ci: false
+
+  require_ci_to_pass: false
+
+  token: >-  # notsecret  # repo-scoped, upload-only, stability in fork PRs
+    REPLACE WITH A CLEAR-TEXT TOKEN
+
+comment:
+  require_changes: true
+
+coverage:
+  range: 100..100
+  status:
+    patch:
+      default:
+        target: 100%
+      pytest:
+        target: 50%
+        flags:
+        - pytest
+      typing:
+        flags:
+        - MyPy
+    project:
+      default:
+        target: 100%
+      lib:
+        flags:
+        - pytest
+        paths:
+        - src/
+        target: 100%
+      tests:
+        flags:
+        - pytest
+        paths:
+        - tests/
+        target: 100%
+      typing:
+        flags:
+        - MyPy
+        target: 100%
+
+...
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -8,7 +8,7 @@ codecov:
   require_ci_to_pass: false
 
   token: >-  # notsecret  # repo-scoped, upload-only, stability in fork PRs
-    REPLACE WITH A CLEAR-TEXT TOKEN
+    7316089b-55fe-4646-b640-78d84b79d109
 
 comment:
   require_changes: true
--- a/.github/workflows/scheduled-runs.yml
+++ b/.github/workflows/scheduled-runs.yml
@@ -1,0 +1,41 @@
+---
+
+name: ⏳
+
+on:
+  pull_request:
+    paths:  # only changes to this workflow itself trigger PR testing
+    - .github/workflows/scheduled-runs.yml
+  schedule:
+  - cron: 3 5 * * *  # run daily at 5:03 UTC
+  workflow_dispatch:  # manual trigger
+
+run-name: >-
+  🌃
+  Nightly run of
+  ${{
+      github.event.pull_request.number && 'PR' || ''
+  }}${{
+      !github.event.pull_request.number && 'Commit' || ''
+  }}
+  ${{ github.event.pull_request.number || github.sha }}
+  triggered by: ${{ github.event_name }} of ${{
+    github.ref
+  }} ${{
+    github.ref_type
+  }}
+  (workflow run ID: ${{
+    github.run_id
+  }}; number: ${{
+    github.run_number
+  }}; attempt: ${{
+    github.run_attempt
+  }})
+
+jobs:
+  main-ci-cd-pipeline:
+    name: 🧪 Main CI/CD pipeline
+    uses: ./.github/workflows/ci-cd.yml
+    secrets: inherit
+
+...
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,34 @@
+[html]
+show_contexts = true
+skip_covered = false
+
+[paths]
+_site-packages-to-src-mapping =
+  src
+  */src
+  *\src
+  */lib/pypy*/site-packages
+  */lib/python*/site-packages
+  *\Lib\site-packages
+
+[report]
+skip_covered = true
+skip_empty = true
+show_missing = true
+exclude_also =
+  ^\s*@pytest\.mark\.xfail
+  ^\s*\.\.\.\s*(#.*)?$
+
+[run]
+branch = true
+cover_pylib = false
+# https://coverage.rtfd.io/en/latest/contexts.html#dynamic-contexts
+# dynamic_context = test_function  # conflicts with `pytest-cov` if set here
+parallel = true
+plugins =
+  covdefaults
+relative_files = true
+source =
+  .
+source_pkgs =
+  pre_commit_terraform
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -1,0 +1,505 @@
+---
+
+name: 🧪
+
+on:
+  merge_group:
+  push:
+    branches-ignore:
+    - dependabot/**  # Dependabot always creates PRs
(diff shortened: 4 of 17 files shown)


.codecov.yml
```diff
@@ -0,0 +1,49 @@
...
+
+codecov:
+  notify:
+    after_n_builds: 5  # Number of test matrix+lint jobs uploading coverage
+    wait_for_ci: false
+
+  require_ci_to_pass: false
+
+  token: >-  # notsecret  # repo-scoped, upload-only, stability in fork PRs
+    REPLACE WITH A CLEAR-TEXT TOKEN
```
webknjaz, line 11: Plz install https://github.com/apps/codecov and fill out this value.

webknjaz: @antonbabenko ^

webknjaz:
While there's one thing to check over in the CI, this can be generally unblocked by setting up Codecov.

.codecov.yml
webknjaz: @antonbabenko post a suggested change with the token on this line, in this comment thread, so I'll be able to incorporate it.

Also, go to https://app.codecov.io/gh/antonbabenko/pre-commit-terraform/flags and enable the flag analytics.

.github/workflows/scheduled-runs.yml
```diff
@@ -0,0 +1,41 @@
+---
+
+name: ⏳
```
MaxymVlasov, line 3: Okay, I'll find and add linter which prevent usage of emojis in code

MaxymVlasov: image

these two equal
image

.codecov.yml
antonbabenko:
```
codecov: 
 token: 5db99ca9-e833-47ba-98d0-b0b58fa44d02
```

Done, analytics is turned on.

antonbabenko: image
I am not sure which auth option to choose?

webknjaz: I haven't tried the tokenless option. It's a new v5 thing. Rumor is that with it, the token shouldn't be necessary.

webknjaz:
@antonbabenko add `check` to branch protection too.

antonbabenko:
@webknjaz Done.

image

.codecov.yml
webknjaz: I've added this token manually.

webknjaz:
FTR, this is ready to be merged. I'll send a yamllint commit on top right after.

MaxymVlasov:
Let's make codecov happy first
image

MaxymVlasov:
@yermulnik JFYI, my strategy for this and other @webknjaz PRs for now - make sure that all introduced tests passed.

Emoji and other things will be wiped out by me later, by separate PR(s). Just need to track what we want to change, to not forget about them after all needed from @webknjaz perspective settings will be merged

That's done this way because @webknjaz do not want to see changes in his branches except him, and don't want to customize much his set of common settings which he uses across multiply projects. If there will be no such incredible value that @webknjaz can provide by his work to us, I will treat this as collaboration antipattern (which I believe it is)  but because all these changes not related to user-faced hooks configuration, and they overall improve workflow for Python hooks, I temporarily agree with this strategy, although I don't like this approach.

webknjaz:
> Let's make codecov happy first. Or fix codecov config, as it not much sense to me to write tests for tests, if I understand correctly what it wants
> image

@MaxymVlasov that's what I meant to show you post-merge. Note that none of these are marked as required. Also, until there's a config merged into `master`, Codecov doesn't really display coverage for different flags in its UI. So I wanted this merged to see what's missing. It might be better to tweak the actual config in a follow-up, when there's something to look at.

It's probably fine to drop `codecov/project`, but not others (this is lower because it combines MyPy type checking with pytest-produced based runtime coverage). MyPy coverage requires work, plus new to the current limitations in the ecosystem, there are corner cases that sometimes prevent 100%, and you can't do anything about them. MyPy coverage measures accuracy of typing. Specifically, whether there's anything in a type that evaluates to `typing.Any`. The complexity here is that this may be coming from third-party typing declarations, not from your code directly.
Hence, `codecov/project/typing` with its imperfect coverage value. Its threshold can be lowered, I suppose. `codecov/project` is a combination of coverage from MyPy and pytest, while `codecov/project/typing` is MyPy-only and `codecov/project/lib` is pytest-only under `src/`.

There's also a separate `codecov/project/tests` for pytest-produced coverage of the `tests/` directory. It's an important metric that must be at 100% at all times — if it's not, this would mean dead code in tests and a false sense of security by the way of thinking that all tests run while some don't.

The patch-marked metrics only take into account what's in the diff. This is effectively more stable because they depend on a fixed number of lines, if you can't set project to 100%. The project-marked ones measure coverage of the entire project which would be flaky since the number of lines to measure changes in patches, which doesn't hit the threshold in some cases. This is why 100% is important, when possible. It should be possible for pytest, but not for MyPy.

I wouldn't touch `codecov/patch` for now, though. It'll be different post-merge.

There's also a 100% requirement on the coveragepy level that must be honored. It's not acceptable to lower it at all. To keep up with it, use different coveragepy-provided mechanisms for things that are known to never be covered or covered only under certain branches. Those are line exclusion patterns in the config and `# pragma: no cover`-like branches (see `covdefaults` for what's configured and available to use).

MaxymVlasov (approved):
As `codecov/patch` check will be fixed by different PR, as codecov settings required to be merged first, approved

webknjaz:
> As `codecov/patch` check will be fixed by different PR

That's the thing. It might not be necessary. Let's observe what gets reported in the other PRs before making the judgement call.

webknjaz:
Here's a follow-up for the missing side effect introduced here but invisible in PR CI runs: #748.

yermulnik:
> @yermulnik JFYI, my strategy for this and other @webknjaz PRs for now - make sure that all introduced tests passed.

Works for me 👍🏻 

> That's done this way because […], I temporarily agree with this strategy, although I don't like this approach.

I'd suggest to not go into this topic using such harsh statements. I do appreciate @webknjaz's contribution, a fresh breath towards Pythonizing and all details provided in conversations. We're just not that mature as Python project (and we're not a Python project).

antonbabenko:

PR #743, opened 2025-01-04
chore: Update essential configurations

MaxymVlasov:

Subset of #741

* Let VS Code know about Python code location

TBD in separate PR add other Python hooks. [b05097e](https://github.com/antonbabenko/pre-commit-terraform/pull/743/commits/b05097ecf92c10374c23f3cb0a0b4955b1cbefd4)


MaxymVlasov:
@webknjaz is there a way to make `pre-commit.ci` run only on changes in PR, not on whole repo?

webknjaz:
> As it generates changes for all these files, when none of them were actually touched in this PR

This is actually good — new linters should include changes to the linted things. This makes such changes atomic.

.pre-commit-config.yaml
```diff
@@ -61,8 +61,89 @@ repos:
 
 # JSON5 Linter
 - repo: https://github.com/pre-commit/mirrors-prettier
-  rev: v3.1.0
+  rev: v4.0.0-alpha.8
```
yermulnik, line 64: Is using alpha/beta versions intentional? It's not the best pattern IMHO.

.pre-commit-config.yaml
```diff
@@ -61,8 +61,89 @@ repos:
...
+
+- repo: https://github.com/astral-sh/ruff-pre-commit
+  rev: v0.8.4
+  hooks:
+    - id: ruff
+      args: [--fix]
+      types_or: [python, pyi]
+    - id: ruff-format
+      types_or: [python, pyi]
+      args: [--config, format.quote-style = 'single', --config, line-length = 100]
```
yermulnik, line 83: There should be a good reason(s) to deviate from linter's default settings.

.pre-commit-config.yaml
```diff
@@ -61,8 +61,89 @@ repos:
...
+  hooks:
+    - id: add-trailing-comma
+
+- repo: https://github.com/pre-commit/mirrors-autopep8
+  rev: v2.0.4
+  hooks:
+    - id: autopep8
+      args:
+      - -i
+      - --max-line-length=100
```
yermulnik, line 103: Out of interest: why do you sometimes use `args: […]` and sometimes split `args` arrays to multiline? Such inconsistency is a bit confusing as it adds impression that this is done for a reason.

.pre-commit-config.yaml
```diff
@@ -61,8 +61,89 @@ repos:
...
+      - --disable=logging-fstring-interpolation # Conflict with "use a single formatting" WPS323
+      - --disable=ungrouped-imports # ignore `if TYPE_CHECKING` case. Other do reorder-python-imports
+      - --disable=R0801 # Similar lines in 2 files. Currently I don't think that it possible to DRY hooks unique files boilerplate
+
+- repo: https://github.com/pre-commit/mirrors-mypy
+  rev: v1.14.1
+  hooks:
+    - id: mypy
+      additional_dependencies:
+          - types-PyYAML
```
yermulnik, line 122: Why four rather than two spaces indentation at this line? 😕

pyproject.toml
```diff
@@ -9,15 +9,12 @@ build-backend = 'hatchling.build'
...
-  'Programming Language :: Python :: 2.7',
   'Programming Language :: Python :: 3',
-  'Programming Language :: Python :: 3.6',
-  'Programming Language :: Python :: 3.7',
+  'Programming Language :: Python :: 3 :: Only',
   'Programming Language :: Python :: Implementation :: CPython',
   'Programming Language :: Python :: Implementation :: PyPy',
 ]
-description = 'Pre-commit hooks for terraform_docs_replace'
+description='Pre-commit hooks for terraform'
```
yermulnik, line 17: Terraform is a proper name. Just like Pre-commit. Either capitalize both or lowercase both please.
```suggestion
description='Pre-commit hooks for Terraform'
```

src/pre_commit_terraform/_cli.py
```diff
@@ -28,8 +26,7 @@ def invoke_cli_app(cli_args: list[str]) -> ReturnCodeType:
         raise
     except PreCommitTerraformRuntimeError as unhandled_exc:
         print(
-            f'App execution took an unexpected turn: {unhandled_exc !s}. '
-            'Exiting...',
+            f'App execution took an unexpected turn: {unhandled_exc !s}. ' 'Exiting...',
```
yermulnik, line 29:
```suggestion
            f'App execution took an unexpected turn: {unhandled_exc !s}. Exiting...',
```

pyproject.toml
webknjaz:
```suggestion
description = 'Pre-commit hooks for Terraform'
```

src/pre_commit_terraform/_cli.py
webknjaz: If you integrate formatters that do this, `flake8-implicit-str-concat` *must* be integrated to combat this without imposing it on unsuspecting human beings...

.pre-commit-config.yaml
webknjaz: It's not because of the defaults (for which sometimes there are good reasons to change) but because other tools wouldn't pick this up, it should be avoided (with a few well-defined exceptions like the one I've got for MyPy).

For this reason, every low-level tool must be configured via its own autoloaded config file name. Otherwise, everybody would have to duplicate this config in their editors, IDEs, other tools and sometimes CI, and would have toe *know upfront* that this is necessary. Being surprised by unexpected behaviors when they don't. I have some notes on this in #742.

I wouldn't say that setting such a long line setting has a merit, though. I'm a firm believer that in the name of readability/inclusivity/maintainability, code should be readable in columns. Just like any text, typography can govern column width for code too. This is directly connected to readability (which is why magazines and newspapers don't print lines from one side of the page to the others, and some books / catalogs use columns). Typographically optimal columns are between 50 and 75 chars (opinions vary, but mostly close to this range): https://baymard.com/blog/line-length-readability. We have a long-standing convention of 79 that exceeds it a bit, but isn't as critical. It allows for columns in editors for working on multiple files while still using enlarged fonts. It also allows for side-by-side diffs. The lines fit on the screen in these settings as well as allow for column-based top-down reading (which is another thing humans do with text naturally, by habit).

.pre-commit-config.yaml
```diff
@@ -61,8 +61,89 @@ repos:
...
+    - id: isort
+      name: isort
+      args: [--force-single-line, --profile=black]
+
+- repo: https://github.com/asottile/add-trailing-comma
+  rev: v3.1.0
+  hooks:
+    - id: add-trailing-comma
+
+- repo: https://github.com/pre-commit/mirrors-autopep8
```
webknjaz, line 97: It's usually unnecessary to use multiple generic code formatters, as they tend to step on each other's toes, resulting in infinite reformatting of the same things back and forth. Choose one and stick with it. It's typically fine with a few smaller ones that change small non-intersecting aspects of files, though. But they may have to be ordered correctly.

.pre-commit-config.yaml
webknjaz: yamllint all the things!

I'd add https://github.com/ansible/awx-plugins/blob/53ebc59/.yamllint with `quoted-strings: {required: only-when-needed}` on top (https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.quoted_strings).

.pre-commit-config.yaml
```diff
@@ -61,8 +61,89 @@ repos:
...
+- repo: https://github.com/pre-commit/mirrors-mypy
+  rev: v1.14.1
+  hooks:
+    - id: mypy
+      additional_dependencies:
+          - types-PyYAML
+      args: [
+        --ignore-missing-imports,
+        --disallow-untyped-calls,
+        --warn-redundant-casts,
```
webknjaz, line 126: Nothing should be disabled globally. Suppressions must be granular, which can be implemented in the config and in-module.

.pre-commit-config.yaml
```diff
@@ -61,8 +61,89 @@ repos:
...
+  hooks:
+    - id: wemake-python-styleguide
+      args:
+      - --allowed-module-metadata=__all__ # Default to ''
+      - --max-local-variables=6 # Default to 5
+      - --max-returns=6 # Default to 5
+      - --max-imports=15 # Default to 12
+      # https://wemake-python-stylegui.de/en/latest/pages/usage/violations/index.html
+      - --extend-ignore=
+          E501   <!-- line too long (> 79 characters). Use 100 -->
```
webknjaz, line 140: Configurable rules shouldn't be ignored really.

pyproject.toml
```diff
@@ -9,15 +9,12 @@ build-backend = 'hatchling.build'
 name = 'pre-commit-terraform'
 classifiers = [
   'License :: OSI Approved :: MIT License',
-  'Programming Language :: Python :: 2',
-  'Programming Language :: Python :: 2.7',
   'Programming Language :: Python :: 3',
-  'Programming Language :: Python :: 3.6',
-  'Programming Language :: Python :: 3.7',
```
webknjaz, line 16: It's best to list envs tested in CI here.

src/pre_commit_terraform/terraform_docs_replace.py
webknjaz, line 1: It's usually best to couple formatting changes with adding a formatter. Not multiple automatic and manual formatting changes with formatters and linters all smashed together. It's difficult to figure out what's coming from where and why that is in this setting.

webknjaz (changes requested):
> to merge before #742

No, #742 must be merged first and this one is nowhere near ready with multiple problems mentioned below + the fact of not having config in correct places.

.pre-commit-config.yaml
MaxymVlasov:
```suggestion
  rev: v3.1.0
```

.gitignore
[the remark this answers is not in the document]
MaxymVlasov: You mean "above boilerplate"

Nobody cares about boilerplate, so it should be moved away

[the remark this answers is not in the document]
webknjaz: The parser does.

.pre-commit-config.yaml
MaxymVlasov: sometimes it depends on how hook configured - IE 

```yaml
[--config, format.quote-style = 'single', --config, line-length = 100]
```

will be actually more readble as

```yaml
[
  --config, format.quote-style = 'single',
  --config, line-length = 100
]
```

not 

```yaml
- --config
- format.quote-style = 'single'
- --config
- line-length = 100]
```

But I don't remember why I can't use multiline array with `[]`, maybe I just not set it in first place when set it a long time ago.
Second reason - our God of code reusage across projects - copy-paste :D 
Nobody never reevaluated it, so it is as it is till now

webknjaz: This is rather subjective. I'd argue that `[]` is the least readable way.

Actually, for `--long-args`, it's best to use `=` not a whitespace. This is useful not only in this config but in Python code too. Since you're integrating autoformatters, they will produce such weird constructs with args disconnected from their values unless you use equals that bundles them into the same string.

.pre-commit-config.yaml
webknjaz: @yermulnik #747

.pre-commit-config.yaml
```diff
@@ -66,3 +66,84 @@ repos:
...
+
+- repo: https://github.com/asottile/add-trailing-comma
+  rev: v3.1.0
+  hooks:
+    - id: add-trailing-comma
+
+- repo: https://github.com/pre-commit/mirrors-autopep8
+  rev: v2.0.4
+  hooks:
+    - id: autopep8
```
webknjaz, line 100: With Ruff in formatting mode, this is probably unnecessary. I'm yet to compare their output, but the recommendation is going to be the same as with pylint for now.

.pre-commit-config.yaml
```diff
@@ -66,3 +66,84 @@ repos:
...
+      args: [--config, format.quote-style = 'single', --config, line-length = 100]
+
+- repo: https://github.com/pycqa/isort
+  rev: 5.13.2
+  hooks:
+    - id: isort
+      name: isort
+      args: [--force-single-line, --profile=black]
+
+- repo: https://github.com/asottile/add-trailing-comma
```
webknjaz, line 92: This one might be fine to combine with Ruff. But you'll have to put it first and also check if they don't cancel each other out in the way they're configured.

.pre-commit-config.yaml
```diff
@@ -66,3 +66,84 @@ repos:
...
+- repo: https://github.com/pre-commit/mirrors-autopep8
+  rev: v2.0.4
+  hooks:
+    - id: autopep8
+      args:
+      - -i
+      - --max-line-length=100
+
+# Usage: http://pylint.pycqa.org/en/latest/user_guide/message-control.html
+- repo: https://github.com/PyCQA/pylint
```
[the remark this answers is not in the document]
yermulnik: Yeah, I guess it's a worthwhile suggestion.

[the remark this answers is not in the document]
webknjaz: FWIW, I'm planning to research replacing many things with Ruff in my projects but just didn't get to do a full comparison.

.pre-commit-config.yaml
MaxymVlasov: They don't

.pre-commit-config.yaml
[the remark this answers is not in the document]
MaxymVlasov: Hmm, ruff has rules that should be manually enabled? Didn't know it

From what I saw - pylint in its current configuration provides additional checks which are not covered by `ruff` in it's current configuration.
At the same time, it totally makes sense to check is `ruff` and `we-make-styleguide` both fully cover `pylint` out of the box or at least able to be configured in such way. 
If you do such research - that's would be lovely, but till I want to have all them in place, as better to catch all possible issues now, than deal with them in 6-12months when they detection will be implemented in `ruff` etc.
`pylint` check is relatively fast, so I want to preserve it for now

.pre-commit-config.yaml
```diff
@@ -61,8 +61,89 @@ repos:
...
+  rev: v3.3.3
+  hooks:
+    - id: pylint
+      args:
+      - --disable=import-error # E0401. Locally you could not have all imports.
+      - --disable=fixme # W0511. 'TODO' notations.
+      - --disable=logging-fstring-interpolation # Conflict with "use a single formatting" WPS323
+      - --disable=ungrouped-imports # ignore `if TYPE_CHECKING` case. Other do reorder-python-imports
+      - --disable=R0801 # Similar lines in 2 files. Currently I don't think that it possible to DRY hooks unique files boilerplate
```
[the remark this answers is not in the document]
MaxymVlasov: I can only pray that `pylint` maintainers will add human-readable names in their error msgs, and not require users to goggle is they exist and how they named

.pre-commit-config.yaml
MaxymVlasov: 1. It covered by other linters, so deduplicate similar errors
2. Idk why it not provide link to docs which describes possibilities how to deal with violation as many hooks do :thinking: It's not easy to google them. This one https://www.flake8rules.com/rules/E501.html doesn't include any suggestions about configuration -_-

.pre-commit-config.yaml
webknjaz: > They don't

I wouldn't be *so* sure. I've seen corner cases with various formatters where they'd do this very occasionally under certain conditions. It's good that we haven't observed such a behavior here. I'm just saying that this should be kept in mind for the future in case it does happen.

If Ruff does not implement a similar capability, it stands to reason that it's okay to keep this formatter. If it does, it might not make sense.

But if you decide to keep this one, make sure to move it along with any other small narrow-scoped formatters before the Ruff run.

.pre-commit-config.yaml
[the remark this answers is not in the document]
webknjaz: > `pylint` check is relatively fast

:laughing: this is the slowest checker I know. This is because it goes beyond static analysis and does some dynamic checking as well.

As for ruff, I wanted to research it more. And yes, there's rules that are disabled by default in all the linters I've ever met (flake8, pylint, ruff). Some updates to pylint sometimes move the rules from default to extensions too — this is why I prefer an explicit config to make sure that it doesn't suddenly stop checking something just because of a version bump.

While I do like pylint in general, I tried it locally on this codebase and there's bugs in it that block really adding it in full capacity. Hence, the idea to delay adding it. I may end up debugging the issue for my other projects. But in general, many people stick with flake8 because it's not as slow and easier to extend. We do run it in `ansible-test`, though.

My initial strategy for adopting Ruff would be enabling as much as possible there, and only if some rule is not ported, then run flake8/pylint (with plugins) only checking those rules, to reduce the overhead. It's good for DX when the same problem isn't reported by 5 different checkers.

.pre-commit-config.yaml
webknjaz: WPS can show shortlinks to the rules, you just didn't enable it :man_shrugging:

This is the setting: https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-max-line-length.

But anyway, it doesn't make sense to have such things as CLI args. I can send in a good config structure where it's more apparent what should go where.

.pre-commit-config.yaml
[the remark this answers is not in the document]
webknjaz: I have this configured by default :man_shrugging: You should've just used my config.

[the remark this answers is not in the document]
webknjaz: Unfortunately, Ruff does not allow using those names in ignores, while the rules may have human-readable names. There's a feature request about it.

yermulnik:
> Remove stale label

✔️ 
Cc @MaxymVlasov

.vscode/settings.json
```diff
@@ -7,4 +7,7 @@
     "code-block-style": false
   },
   "markdown.validate.enabled": true,
+  "python.analysis.extraPaths": [
+    "./src"
```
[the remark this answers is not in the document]
MaxymVlasov:
```suggestion
    "./src",
    "./tests/pytest"
```

pyproject.toml
```diff
@@ -27,18 +27,17 @@ upstreaming = [
...
   'Programming Language :: Python :: 3.10',
   'Programming Language :: Python :: 3.11',
   'Programming Language :: Python :: 3.12',
   'Programming Language :: Python :: 3.13',
-
   'Programming Language :: Python :: Implementation :: CPython',
   'Programming Language :: Python :: Implementation :: PyPy',
 ]
-description = 'Pre-commit hooks for terraform_docs_replace'
+description = 'Pre-commit hooks for Terraform'
```
MaxymVlasov, line 40:
```suggestion
description = 'Pre-commit hooks for Terraform, OpenTofu, Terragrunt and related tools'
```

.gitignore
```diff
@@ -177,3 +177,8 @@ pyrightconfig.json
 # End of https://www.toptal.com/developers/gitignore/api/python
 
 tests/results/*
+
+# Python
+__pycache__/
+*.py[cod]
+node_modules/
```
MaxymVlasov, line 184: Actually, I not remember from what hook it comes. I can discard changes to this file (or remove just node_modules)  as well

yermulnik: > (or remove just node_modules)

I'm ok with this option as we seem to don't have Node.js stuff here yet, right? 🤔

pyproject.toml
```diff
@@ -27,18 +27,17 @@ upstreaming = [
...
   'Programming Language :: Python :: 3.10',
   'Programming Language :: Python :: 3.11',
   'Programming Language :: Python :: 3.12',
   'Programming Language :: Python :: 3.13',
-
   'Programming Language :: Python :: Implementation :: CPython',
   'Programming Language :: Python :: Implementation :: PyPy',
 ]
-description = 'Pre-commit hooks for terraform_docs_replace'
+description = 'Pre-commit hooks for Terraform, OpenTofu, Terragrunt and related tools'
```
yermulnik, line 40:
```suggestion
description = 'Pre-commit hooks for Terraform, OpenTofu, Terragrunt, and related tools'
```

I guess it's more common to add serial comma in this case rather than to not add to distinguish between "hooks for related tools also" and "hook for TF, Tofu, TF + some other tools related to this".

yermulnik (dismissed):

pyproject.toml
yermulnik: Hmm, the more I re-read the sentence the more ambiguous is whether to use or to not use serial comma 😕 I hate English grammar 🤣

.gitignore
webknjaz: Might be good to get something from gitignore.io in a separate PR

webknjaz: > I'm ok with this option as we seem to don't have Node.js stuff here yet, right? 🤔

Could be pyright that VS Code uses. Not sure. Though, one might argue that non-project specific config should go into a user-global gitignore and not the local one.

webknjaz (approved):
Feel free to split this into more parts but I don't have any strong objections regarding stuff in the current diff.

pyproject.toml
webknjaz: There's many styles of English grammar. So you're good. If you want to enforce something specific, look into plugging Vale.

.gitignore
yermulnik: > Though, one might argue that non-project specific config should go into a user-global gitignore and not the local one.

💯

pyproject.toml
yermulnik: > If you want to enforce something specific, look into plugging Vale.

Nice to know. Thanks 👍🏻

.gitignore
MaxymVlasov: I dropped .gitignore changes for now

yermulnik (approved):

antonbabenko:
This PR is included in version 1.99.0 :tada:

PR #747, opened 2025-01-08
chore: Integrate `yamllint` into the project

webknjaz:
The initial config does not change many defaults and mostly relies on the upstream config. The only changes are to the sequence style, quoting and allowed truthy values that accomodate for GHA.

- [ ] This PR fixes a bug.
- [x] This PR is contributor-facing, internal stuff.

N/A

### How can we test changes

..stare at the diff? `tox r -qq -e pre-commit -- yamllint --all-files` otherwise.

### DIFF
--- a/.github/.container-structure-test-config.yaml
+++ b/.github/.container-structure-test-config.yaml
@@ -1,101 +1,145 @@
-schemaVersion: '2.0.0'
+---
+
+schemaVersion: 2.0.0
+
 commandTests:
-  - name: "git"
-    command: "git"
-    args: ["--version"]
-    expectedOutput: ["^git version 2\\.[0-9]+\\.[0-9]+\\n$"]
+- name: git
+  command: git
+  args:
+  - --version
+  expectedOutput:
+  - ^git version 2\.[0-9]+\.[0-9]+\n$
 
-  - name: "pre-commit"
-    command: "pre-commit"
-    args: ["-V"]
-    expectedOutput: ["^pre-commit ([0-9]+\\.){2}[0-9]+\\n$"]
+- name: pre-commit
+  command: pre-commit
+  args:
+  - -V
+  expectedOutput:
+  - ^pre-commit ([0-9]+\.){2}[0-9]+\n$
 
-  - name: "gcc"
-    command: "gcc"
-    args: ["--version"]
-    expectedOutput: ["^gcc \\(Alpine 12\\."]
+- name: gcc
+  command: gcc
+  args:
+  - --version
+  expectedOutput:
+  - ^gcc \(Alpine 12\.
 
-  - name: "checkov"
-    command: "checkov"
-    args: ["--version"]
-    expectedOutput: ["^([0-9]+\\.){2}[0-9]+\\n$"]
+- name: checkov
+  command: checkov
+  args:
+  - --version
+  expectedOutput:
+  - ^([0-9]+\.){2}[0-9]+\n$
 
-  - name: "infracost"
-    command: "infracost"
-    args: ["--version"]
-    expectedOutput: ["^Infracost v([0-9]+\\.){2}[0-9]+"]
+- name: infracost
+  command: infracost
+  args:
+  - --version
+  expectedOutput:
+  - ^Infracost v([0-9]+\.){2}[0-9]+
 
-  - name: "opentofu"
-    command: "tofu"
-    args: ["-version"]
-    expectedOutput: ["^OpenTofu v([0-9]+\\.){2}[0-9]+\\n"]
+- name: opentofu
+  command: tofu
+  args:
+  - --version
+  expectedOutput:
+  - ^OpenTofu v([0-9]+\.){2}[0-9]+\n
 
-  - name: "terraform"
-    command: "terraform"
-    args: ["-version"]
-    expectedOutput: ["^Terraform v([0-9]+\\.){2}[0-9]+\\n"]
+- name: terraform
+  command: terraform
+  args:
+  - --version
+  expectedOutput:
+  - ^Terraform v([0-9]+\.){2}[0-9]+\n
 
-  - name: "terraform-docs"
-    command: "terraform-docs"
-    args: ["--version"]
-    expectedOutput: ["^terraform-docs version v([0-9]+\\.){2}[0-9]+ [a-z0-9]+"]
+- name: terraform-docs
+  command: terraform-docs
+  args:
+  - --version
+  expectedOutput:
+  - ^terraform-docs version v([0-9]+\.){2}[0-9]+ [a-z0-9]+
 
-  - name: "terragrunt"
-    command: "terragrunt"
-    args: ["--version"]
-    expectedOutput: ["^terragrunt version v([0-9]+\\.){2}[0-9]+\\n$"]
+- name: terragrunt
+  command: terragrunt
+  args:
+  - --version
+  expectedOutput:
+  - ^terragrunt version v([0-9]+\.){2}[0-9]+\n$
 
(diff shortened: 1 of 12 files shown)


webknjaz:
@yermulnik this is the yamllint config I was talking about earlier.

webknjaz:
Looks like the CI is having problems accessing DockerHub. It's not related to this PR and somebody should restart the CI jobs once GH fixes their platform issues.

.pre-commit-config.yaml
```diff
@@ -1,44 +1,57 @@
...
-        ^tests/[^_].*\.py$
+  # Non-modifying checks:
+  - id: name-tests-test
+    files: >-
+      ^tests/[^_].*\.py$
+
+- repo: https://github.com/adrienverge/yamllint.git
+  rev: v1.35.1
+  hooks:
+  - id: yamllint
```
MaxymVlasov, line 49: It not format yaml locally, that's is unfortunate.


Better use 
```yaml
  - repo: https://github.com/jumanjihouse/pre-commit-hook-yamlfmt
    rev: 0.2.3
    hooks:
      - id: yamlfmt
        args: [
          --mapping, '2',
          --sequence, '4',
          --offset, '2',
          --width, '100',
          --implicit_start,
          --preserve_null,
        ]
```

which actually format code, and enforce style which should be used here anyway

MaxymVlasov: They actually can be combined by

```
    - repo: https://github.com/adrienverge/yamllint
      rev: v1.35.1
      hooks:
          - id: yamllint
            types:
                - file
                - yaml
            args: [--format, parsable, --strict]

    - repo: https://github.com/jumanjihouse/pre-commit-hook-yamlfmt
      rev: 0.2.3
      hooks:
          - id: yamlfmt
```

Just need to change `.yamllint` a bit

webknjaz: This is a linter, not a formatter. I'm yet to see a formatter working well. But FWIW this is out of the scope here. This PR is about a specific linter and anything else is an undesired scope creep that can be debated elsewhere.

webknjaz: > They actually can be combined by

This is a bad idea as formatters should always go before linters to prevent double errors.

MaxymVlasov: In this case, `yamllint` is redundant as `yamlfmt` will fix all these issues
If you want to have it for some unknown to me reason, I can adjust `yamllint` config

There just only issue which I see - such type of comment are not handled properly by `yamlfmt`, but they always can be moved or above, or just get rid of `>-` in simple constructions
image

webknjaz: Formatting is out of the scope. So that's irrelevant. But to answer the question about a separate line — no, it's not normally okay to put comments on a separate line due to context drifting. This is important in any code-like files, including but not limited to YAML. The comment is there intentionally and must not be moved.

Just to reiterate, the scope of this PR is solely adding one specific linter and nothing else. Any other attempts to change the scope indefinitely are to be rejected with prejudice.

.yamllint
```diff
@@ -0,0 +1,20 @@
+---
+
+extends: default
+
+rules:
+  indentation:
+    level: error
+    indent-sequences: false
```
MaxymVlasov, line 8:
```suggestion
  indentation:
    level: error
    indent-sequences: true
  line-length:
    max: 100
  document-start:
    present: false
  comments:
    min-spaces-from-content: 1
```

.pre-commit-config.yaml
MaxymVlasov: If you want to limit just to linter, here we go then.

.github/.container-structure-test-config.yaml
MaxymVlasov, line 1: And revert/fix changes for `terrascan` and `hcledit` as you broke them
https://github.com/antonbabenko/pre-commit-terraform/actions/runs/12676686555/job/35330392386?pr=747

yermulnik: @webknjaz Those two expect a subcommand rather than an arg: `--version` should be `version`.

.yamllint
webknjaz: I think the line length setting is unreasonably long. Also, double whitespace before a comment is an accepted convention. A single space seems dumb with this being required just about everywhere.

.github/.container-structure-test-config.yaml
yermulnik: So it's not about reverting, but rather about adjusting the change.

.github/.container-structure-test-config.yaml
```diff
@@ -1,101 +1,145 @@
...
+  command: terragrunt
+  args:
+  - --version
+  expectedOutput:
+  - ^terragrunt version v([0-9]+\.){2}[0-9]+\n$
+
+- name: terrascan
+  command: terrascan
+  args:
+  - --version
```
webknjaz, line 72:
```suggestion
  - version
```

.github/.container-structure-test-config.yaml
```diff
@@ -1,101 +1,145 @@
...
+  command: tfupdate
+  args:
+  - --version
+  expectedOutput:
+  - ([0-9]+\.){2}[0-9]+\n$
+
+- name: hcledit
+  command: hcledit
+  args:
+  - --version
```
webknjaz, line 109:
```suggestion
  - version
```

.github/.container-structure-test-config.yaml
webknjaz: Should be fixed. Marking as resolved.

.yamllint
MaxymVlasov: These "Unnecessary sequence indentation" exist in the same reason why Python have 4 space indentation, not 2 - it less error-prone.

It is much simpler to figure out where part of array belong to in next construction
```yaml
- repo: https://github.com/adrienverge/yamllint.git
  rev: v1.35.1
  hooks:
    - id: yamllint
      types:
        - file
        - yaml
     args:
        - --strict
```

than in 
```yaml
- repo: https://github.com/adrienverge/yamllint.git
  rev: v1.35.1
  hooks:
  - id: yamllint
    types:
    - file
    - yaml
   args:
    - --strict
```

just because you can't have list and K/V on the same level.

This is a common and very annoying problem, and the right solution to where indent should be is not covered by linters or formatters, so the only way to deal with it if there is no such indents and no `jsonschema` check - manually go and check documentation each time for tool or run this tool.

And just noticed "Unnecessary sequence indentation"  - is default setting of yamllint itself, and YAML formatter built-in in VS Code which, btw, ignores `.yamllint` config, but respects `.editorconfig`

More complex solution to it could be `indentation.spaces=4`, but it looks much worse

MaxymVlasov: > I think the line length setting is unreasonably long

Second vertical line is 80 chars, third - 100 chars
Screenshot from 2025-01-09 15-09-08

<details><summary>You added <b>18 ignores for line-length</b> in such simple project. And after that you said that it "unreasonably long"?</summary>

Screenshot from 2025-01-09 15-08-06

</details>

MaxymVlasov: About
```
  document-start:
    present: false
  comments:
    min-spaces-from-content: 1
```
First is just my neat-peak to skip unnecessary syntax, second - how `yamlfmt` formats comments and that, sadly not configurable, but I love to have automation which able to maintain "good enough" style, than go and fix issues manually to the best possible style.
But yes, they both can be skipped for now

webknjaz:
@yermulnik could you restart the cancelled job?

MaxymVlasov (approved):
Will change it later

antonbabenko:

PR #748, opened 2025-01-08
ci: Disable `no-commit-to-branch` @ GHA

webknjaz:
This turns off the "is push to master" in GitHub Actions CI/CD as it does not make sense in this environment, being local-only [[1]].

[1]: https://github.com/pre-commit/pre-commit-hooks/issues/1124

- [ ] This PR fixes a bug.
- [x] Internal maintenance, no effect on end-users or contributors.

This is a follow-up for #742. One of its jobs runs `pre-commit` and doesn't filter out this hook. The hook itself would only fail on pushes to `master` which is not a condition that can happen in PR workflow runs.

### How can we test changes

You can't. This will only be visible on `master`.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -387,8 +387,14 @@ jobs:
         - pre-commit
         - metadata-validation
         environment-variables:
+        # `no-commit-to-branch` is skipped because it does not make sense
+        # in the CI, only locally.
+        # Ref: https://github.com/pre-commit/pre-commit-hooks/issues/1124
         - >-  # only affects pre-commit, set for all for simplicity:
-          SKIP=hadolint,shfmt
+          SKIP=
+          hadolint,
+          no-commit-to-branch,
+          shfmt,
         tox-run-posargs:
         - ''
         xfail:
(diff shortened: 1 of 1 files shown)


MaxymVlasov (approved):

.github/workflows/ci-cd.yml
```diff
@@ -387,8 +387,14 @@ jobs:
...
         environment-variables:
+        # `no-commit-to-branch` is skipped because it does not make sense
+        # in the CI, only locally.
+        # Ref: https://github.com/pre-commit/pre-commit-hooks/issues/1124
         - >-  # only affects pre-commit, set for all for simplicity:
-          SKIP=hadolint,shfmt
+          SKIP=
+          hadolint,
+          no-commit-to-branch,
+          shfmt,
```
MaxymVlasov, line 397: just curios, can we have this?
```suggestion
          SKIP=
            hadolint,
            no-commit-to-branch,
            shfmt,
```

IMO, that will make it more readable, as it specify visually that skipped values are actually values for SKIP key

webknjaz: Sometimes, we can, sometimes we can't. I think I've seen some corner cases when using `>-`, which is why I usually don't add those indents.

But yes, you're right, it's more readable. So let's try it and see whether it breaks.

webknjaz:
@MaxymVlasov it breaks in a different place. I reverted it.

MaxymVlasov:
sad emoji

PR #749, opened 2025-01-09
chore: Rename 'py' to 'pytest' in GHA to avoid confusion

MaxymVlasov:

When you see `py` in tests it not clear what exactly going on. Until now I thought that it represents "python", but actually this part run `pytest`
image

So, for avoiding confusion and add ability to explicitly specify `tox r -e pytest`, this PR was made

### How can we test changes

Wait till tests pass on this PR

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -452,7 +452,7 @@ jobs:
         - macos-13
         - windows-2025
         toxenv:
-        - py
+        - pytest
         xfail:
         - false
 
--- a/.github/workflows/reusable-tox.yml
+++ b/.github/workflows/reusable-tox.yml
@@ -262,7 +262,7 @@ jobs:
       shell: bash
     - name: Download all the dists
       if: >-
-        contains(fromJSON('["metadata-validation", "py"]'), inputs.toxenv)
+        contains(fromJSON('["metadata-validation", "pytest"]'), inputs.toxenv)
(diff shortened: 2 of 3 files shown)


yermulnik (approved):

yermulnik:
Wow, 56 tests 😲 

I also think we should ask @webknjaz to also review and approve Py-based PRs here to get his knowledge work for us 😄

MaxymVlasov:
Rename at least in GHA view works fine, but some tests failed, idk why. Rerun doesn't help 
@webknjaz I see that these test runs on ubuntu 24.04, when prev were on 22.04. Could it be the reason why it failed? In other 2 PRs for test also used ubuntu 24.04, but their tests passed  :thinking:

webknjaz:
The "Download all the dists" step got skipped and the following job couldn't install the expected wheel artifact for testing, crashing the job. The rest were auto-cancelled due to fail-fast in the matrix.

webknjaz:
It's possible their past platform issues haven't been fully resolved: https://www.githubstatus.com/history. Give it time and restart in the morning.

webknjaz:
The root cause is that this part hasn't yet been made reusable: https://github.com/antonbabenko/pre-commit-terraform/blob/5181957/.github/workflows/reusable-tox.yml#L265C52-L265C56. You'll need to add the new name to the hardcoded list there, for now.

antonbabenko:

PR #751, opened 2025-01-09
chore(mypy): Fix type annotation

MaxymVlasov:

MyPy found that `dirs` is not annotated. This PR fix it.

### How can we test changes

coverage improved

### DIFF
--- a/src/pre_commit_terraform/terraform_docs_replace.py
+++ b/src/pre_commit_terraform/terraform_docs_replace.py
@@ -49,7 +49,7 @@ def invoke_cli_app(parsed_cli_args: Namespace) -> ReturnCodeType:
         category=UserWarning,
     )
 
-    dirs = []
+    dirs: list[str] = []
     for filename in cast_to(list[str], parsed_cli_args.filenames):
         if (os.path.realpath(filename) not in dirs and
                 (filename.endswith(".tf") or filename.endswith(".tfvars"))):


src/pre_commit_terraform/terraform_docs_replace.py
```diff
@@ -49,7 +49,7 @@ def invoke_cli_app(parsed_cli_args: Namespace) -> ReturnCodeType:
         category=UserWarning,
     )
 
-    dirs = []
+    dirs: list[str] = []
```
yermulnik, line 52: Too pythonish for me 🤣

yermulnik (approved):

src/pre_commit_terraform/terraform_docs_replace.py
MaxymVlasov:
```
<variable_name>[: <variable_type>] = <variable_value>
```

Actually, I think that `= <variable_value>` is optional too, but I can't remember when I saw such usage last time. It could be forbidden by most basic linters, tbh.

webknjaz: It's not forbidden. You can declare a type of a variable before you actually initialize it. For most simple types, the type checkers like MyPy would infer the type of the data intended to be stored there. But when it's an empty container, like a list, it's ambiguous. `dirs = []` made MyPy *think* the type was `list[typing.Any]` while a more precise type should be `list[str]`.

When declaring typing, the guidance is to use a narrow type when returning things from functions but wide types for the incoming function arguments. It allows passing more shapes of data into functions while being accurate regarding what they return. Individual variables should also be as accurate as possible.

antonbabenko:

PR #752, opened 2025-01-14
feat(`terraform_docs`): Add support for custom markers to better support other formats than Markdown

tarfu:

- [ ] This PR fixes a bug.

This adds support to pass custom markers.
This helps to improve support for other formats like asciidoc which you can use if you use a config for terraform-doc.

### How can we test changes

Add your custom marker to a README and run the hook with the options. You may also want to change the template used by terraform-doc

```
terraform_docs.sh --hook-config=--path-to-file=README.adoc --hook-config="--custom-marker-begin=// BEGIN_TF_DOCS" --hook-config="--custom-marker-end=// END_TF_DOCS" --args=--config=.terraform-docs.yml ./path/to/file.tf
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -610,7 +610,7 @@ Unlike most other hooks, this hook triggers once if there are any changed files 
         - --hook-config=--use-standard-markers=true     # Boolean. Defaults to true (v1.93+), false (<v1.93). Set to true for compatibility with terraform-docs
           # The following two options "--custom-marker-begin" and "--custom-marker-end" are ignored if "--use-standard-markers" is set to false
         - --hook-config=--custom-marker-begin           # String. Defaults to "<!-- BEGIN_TF_DOCS -->" (v1.93+), "<!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->" (<v1.93).
-                                                        # Set to use custom marker which helps you with using other formats like asciidoc. 
+                                                        # Set to use custom marker which helps you with using other formats like asciidoc.
                                                         # For Asciidoc this could be "--hook-config=--custom-marker-begin=// BEGIN_TF_DOCS"
         - --hook-config=--custom-marker-end             # String. Defaults to "<!-- END_TF_DOCS -->" (v1.93+), "<!-- END OF PRE-COMMIT-TERRAFORM DOCS HOOK -->" (<v1.93).
                                                         # Set to use custom marker which helps you with using other formats like asciidoc.
--- a/README.md
+++ b/README.md
@@ -609,13 +609,13 @@ Unlike most other hooks, this hook triggers once if there are any changed files 
         - --hook-config=--create-file-if-not-exist=true # Boolean. true or false
         - --hook-config=--use-standard-markers=true     # Boolean. Defaults to true (v1.93+), false (<v1.93). Set to true for compatibility with terraform-docs
           # The following two options "--custom-marker-begin" and "--custom-marker-end" are ignored if "--use-standard-markers" is set to false
-        - --hook-config=--custom-marker-begin           # String. Defaults to "<!-- BEGIN_TF_DOCS -->" (v1.93+), "<!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->" (<v1.93).
+        - --hook-config=--custom-marker-begin=<!-- BEGIN_TF_DOCS -->  # String.
                                                         # Set to use custom marker which helps you with using other formats like asciidoc.
                                                         # For Asciidoc this could be "--hook-config=--custom-marker-begin=// BEGIN_TF_DOCS"
-        - --hook-config=--custom-marker-end             # String. Defaults to "<!-- END_TF_DOCS -->" (v1.93+), "<!-- END OF PRE-COMMIT-TERRAFORM DOCS HOOK -->" (<v1.93).
+        - --hook-config=--custom-marker-end=<!-- END_TF_DOCS -->  # String.
                                                         # Set to use custom marker which helps you with using other formats like asciidoc.
                                                         # For Asciidoc this could be "--hook-config=--custom-marker-end=// END_TF_DOCS"
-        - --hook-config=--custom-doc-header             # String. Defaults to "# "
+        - --hook-config=--custom-doc-header="# "        # String. Defaults to "# "
                                                         # Set to use custom marker which helps you with using other formats like asciidoc.
                                                         # For Asciidoc this could be "--hook-config=--custom-marker-end=\= "
     ```
--- a/README.md
+++ b/README.md
@@ -615,6 +615,9 @@ Unlike most other hooks, this hook triggers once if there are any changed files 
         - --hook-config=--custom-marker-end             # String. Defaults to "<!-- END_TF_DOCS -->" (v1.93+), "<!-- END OF PRE-COMMIT-TERRAFORM DOCS HOOK -->" (<v1.93).
                                                         # Set to use custom marker which helps you with using other formats like asciidoc.
                                                         # For Asciidoc this could be "--hook-config=--custom-marker-end=// END_TF_DOCS"
+        - --hook-config=--custom-doc-header             # String. Defaults to "# "
+                                                        # Set to use custom marker which helps you with using other formats like asciidoc.
+                                                        # For Asciidoc this could be "--hook-config=--custom-marker-end=\= "
     ```
 
 4. If you want to use a terraform-docs config file, you must supply the path to the file, relative to the git repo root path:
--- a/README.md
+++ b/README.md
@@ -608,8 +608,13 @@ Unlike most other hooks, this hook triggers once if there are any changed files 
         - --hook-config=--add-to-existing-file=true     # Boolean. true or false
         - --hook-config=--create-file-if-not-exist=true # Boolean. true or false
         - --hook-config=--use-standard-markers=true     # Boolean. Defaults to true (v1.93+), false (<v1.93). Set to true for compatibility with terraform-docs
-        - --hook-config=--custom-marker-begin     # String. Defaults to "<!-- BEGIN_TF_DOCS -->" (v1.93+), "<!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->" (<v1.93). Set to use custom marker which helps you with using other formats like asciidoc
-        - --hook-config=--custom-marker-end     # String. Defaults to "<!-- END_TF_DOCS -->" (v1.93+), "<!-- END OF PRE-COMMIT-TERRAFORM DOCS HOOK -->" (<v1.93). Set to use custom marker which helps you with using other formats like asciidoc
+          # The following two options "--custom-marker-begin" and "--custom-marker-end" are ignored if "--use-standard-markers" is set to false
+        - --hook-config=--custom-marker-begin           # String. Defaults to "<!-- BEGIN_TF_DOCS -->" (v1.93+), "<!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->" (<v1.93).
+                                                        # Set to use custom marker which helps you with using other formats like asciidoc. 
+                                                        # For Asciidoc this could be "--hook-config=--custom-marker-begin=// BEGIN_TF_DOCS"
+        - --hook-config=--custom-marker-end             # String. Defaults to "<!-- END_TF_DOCS -->" (v1.93+), "<!-- END OF PRE-COMMIT-TERRAFORM DOCS HOOK -->" (<v1.93).
+                                                        # Set to use custom marker which helps you with using other formats like asciidoc.
+                                                        # For Asciidoc this could be "--hook-config=--custom-marker-end=// END_TF_DOCS"
     ```
 
 4. If you want to use a terraform-docs config file, you must supply the path to the file, relative to the git repo root path:
--- a/README.md
+++ b/README.md
@@ -608,6 +608,8 @@ Unlike most other hooks, this hook triggers once if there are any changed files 
         - --hook-config=--add-to-existing-file=true     # Boolean. true or false
         - --hook-config=--create-file-if-not-exist=true # Boolean. true or false
         - --hook-config=--use-standard-markers=true     # Boolean. Defaults to true (v1.93+), false (<v1.93). Set to true for compatibility with terraform-docs
+        - --hook-config=--custom-marker-begin     # String. Defaults to "<!-- BEGIN_TF_DOCS -->" (v1.93+), "<!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->" (<v1.93). Set to use custom marker which helps you with using other formats like asciidoc
+        - --hook-config=--custom-marker-end     # String. Defaults to "<!-- END_TF_DOCS -->" (v1.93+), "<!-- END OF PRE-COMMIT-TERRAFORM DOCS HOOK -->" (<v1.93). Set to use custom marker which helps you with using other formats like asciidoc
     ```
 
 4. If you want to use a terraform-docs config file, you must supply the path to the file, relative to the git repo root path:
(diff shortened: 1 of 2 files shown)


yermulnik (commented):
Thanks for your contribution. Please have a look into comments I added.
I'd also think of adding this as a single configuration parameter to avoid use of mix of standard and custom markers. Something like `--custom-markers="// BEGIN;// END"`. This may decrease clarity and readability and separator needs a more thorough think though.

yermulnik (commented):
Looks good. @MaxymVlasov What do you think?

hooks/terraform_docs.sh
```diff
@@ -42,8 +42,8 @@ function replace_old_markers {
 
   # Determine the appropriate sed command based on the operating system (GNU sed or BSD sed)
   sed --version &> /dev/null && SED_CMD=(sed -i) || SED_CMD=(sed -i '')
-  "${SED_CMD[@]}" -e "s/^${old_insertion_marker_begin}$/${insertion_marker_begin}/" "$file"
-  "${SED_CMD[@]}" -e "s/^${old_insertion_marker_end}$/${insertion_marker_end}/" "$file"
+  "${SED_CMD[@]}" -e "s/^${old_insertion_marker_begin}$/${insertion_marker_begin//\//\\/}/" "$file"
+  "${SED_CMD[@]}" -e "s/^${old_insertion_marker_end}$/${insertion_marker_begin//\//\\/}/" "$file"
```
MaxymVlasov, line 46: Did you mean just add `//\//\\/` as for L45?

```suggestion
  "${SED_CMD[@]}" -e "s/^${old_insertion_marker_end}$/${insertion_marker_end//\//\\/}/" "$file"
```

tarfu: Yeah, that is a copy fail.

MaxymVlasov:
Hmm, you probably had `// BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK` and `// END OF PRE-COMMIT-TERRAFORM DOCS HOOK`?


```
trace: replace_old_markers terraform_docs main main
       terraform_docs.sh:45: sed -i -e 's/^<!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->$/\/\/ BEGIN_TF_DOCS/' README.adoc

trace: replace_old_markers terraform_docs main main
       terraform_docs.sh:46: sed -i -e 's/^<!-- END OF PRE-COMMIT-TERRAFORM DOCS HOOK -->$/\/\/ END_TF_DOCS/' README.adoc
```

MaxymVlasov:
Also, I found that 

```yaml
formatter: asciidoc table
output:
  file: README.adoc
  mode: inject
  template: |-
    // BEGIN_TF_DOCS
    {{ .Content }}
    // END_TF_DOCS
```

```yaml
repos:
  rev: fc53d0a10b7846fa8571ddd504660cbe5c6110d6
    - id: terraform_docs
      args:
      #- --hook-config=--custom-marker-begin=// BEGIN_TF_DOCS
      #- --hook-config=--custom-marker-end=// END_TF_DOCS
```

When you have empty `README.adoc`, works fine without `custom-marker-begin` and `custom-marker-end` options.

And it produces identical result as for `rev: 1.96.3`

Can you please elaborate on what exactly you tried to fix/add by this PR?

tarfu:
If you got a non-empty file, it won't work for me without the custom markers.

```
= Terraform Module
This is a short description of the module and below this should be the terraform docs

// BEGIN_TF_DOCS

// END_TF_DOCS
```

With this snippet in an existing file, I won't get any output with the latest release/master.
If you add the standard markers below (which are then shown in the rendered asciidoc as it is not parsing them as comments), you'll get the output.


```
= Terraform Module
This is a short description of the module and below this should be the terraform docs

// BEGIN_TF_DOCS

// END_TF_DOCS

<!-- BEGIN_TF_DOCS -->
<!-- END_TF_DOCS -->
```

This will result in:


```
= Hub Management

This folder contains the Terraform code for deployment of the Hub resources, including Virtual WAN and Intermediate Hubs.

// BEGIN_TF_DOCS
=== Requirements

[cols="a,a",options="header,autowidth"]
|===
|Name |Version
|[[requirement_terraform]] <<requirement_terraform,terraform>> |~> 1.7
|===

=== Providers

No providers.

=== Modules

No modules.

=== Resources

No resources.

=== Inputs

No inputs.

=== Outputs

No outputs.
// END_TF_DOCS

<!-- BEGIN_TF_DOCS -->
<!-- END_TF_DOCS -->
```

This also works if you only have a begin marker, as it is the only thing the grep looks for

MaxymVlasov:
I see what you're talking about

Interesting, that specified by you example of

```
= Terraform Module
This is a short description of the module and below this should be the terraform docs

// BEGIN_TF_DOCS

// END_TF_DOCS
```

work well with

```yaml
formatter: asciidoc table
output:
  file: README.adoc
  mode: inject
  template: |-
    // BEGIN_TF_DOCS
    {{ .Content }}
    // END_TF_DOCS
```
```yaml
repos:
  rev: v1.96.3
  hooks:
    - id: terraform_docs
      args:
      - --args=--config=.terraform-docs.yml
```
but not with `- --hook-config=--add-to-existing-file=false` which is default

and 

```yaml
      args:
      - --args=--config=.terraform-docs.yml
```
produce

```
# dir-name

<!-- BEGIN_TF_DOCS -->
<!-- END_TF_DOCS -->

// BEGIN_TF_DOCS
== Requirements
...
// END_TF_DOCS
```

---

So if `--hook-config=--add-to-existing-file=true` able to use right templating in 1.96.3 from `.terraform-docs.yml`, I'm curious why default can't do the same without extra configs to hook? There could be a bug somewhere :thinking: => there should be a better way than proposed by this PR.

If you have time, please investigate what happens, using `PCT_LOG=trace pre-comit run` command and let us know if you'll find anything weird and try to fix it if possible. If not, I'll check later by my own

tarfu:
`--hook-config=--add-to-existing-file=true` skips the marker check, so it will always run tf-docs if it is set to false, it will check if the marker we expect is present and only if it is we will run tf-docs.
As we don't parse the config and try to extract the marker from there, I think the behavior is expected. If it is desired like that is another question.

You could try to parse the template in the config, grab the first line of it, check if it is not `{{ .Content }}` and hope everyone is using the YAML string style we expect. Which leads to other issues.

MaxymVlasov:
Am I right that you anyway need to specify `template:` in `.terraform-docs.yml`  if you use not `<!-- BEGIN_TF_DOCS -->` markers, which looks like are set by default?
https://terraform-docs.io/user-guide/configuration/

So, we actually need to find all  [`formatter:` options](https://terraform-docs.io/user-guide/configuration/formatter/) that not support HTML comments, and add check that if such `formater` provided - run `terraform_docs` same way as for `--hook-config=--add-to-existing-file=true`

tarfu:
Yeah, it is set up defaulting to Markdown and HTML comments, but there would be other options in MD as well.

If you run it like it was `--hook-config=--add-to-existing-file=true` you would lose the ability to skip files without the marker. Which is funnily enough the behavior I want/need :D

MaxymVlasov:
>You would lose the ability to skip files without the marker 

Ah, yes. And parse it from `template` will be problematic. 
Your PR looks perfectly fine then.

Just let make file creation header (`#`) be customizable too then for `- --hook-config=--create-file-if-not-exist=true`, in

https://github.com/antonbabenko/pre-commit-terraform/blob/e01b4dec5af830e25240f5de9ae81021f5d33dbd/hooks/terraform_docs.sh#L199-L204

as 

```
# dir-name-here

// BEGIN_TF_DOCS
== Requirements
...
// END_TF_DOCS
```

will be annoying to fix.
There cloud be sense to make `${PWD##*/}` customizable as well, but it out of scope for sure (and maybe not needed at all).

So, let's reiterate on same question by for different use-case - can we just get header data from `formatter` like `doc_header_prefix[${formatter}]` or we need to hardcore it as well via `--hook-config=--custom-doc-header`?

tarfu:
I think `--hook-config=--custom-doc-header` would be the way to go, as different formatters have multiple options to go with and would be the easiest way to be customized.

Another way could be just, letting terraform-docs create the file would be an option as well, which should work if there is a `tf` file.
This would enable us to throw away one marker and only test against the beginning marker.

MaxymVlasov:
Please add `--hook-config=--custom-doc-header` then, with default to `#`
Don't forget docs section as well.

> Another way could be just, letting terraform-docs create the file would be an option as well, which should work if there is a `tf` file. This would enable us to throw away one marker and only test against the beginning marker.

Interesting solution, but let someone comes with feature request, as I don't see who could use it :)

MaxymVlasov (approved):

README.md
```diff
@@ -608,6 +608,16 @@ Unlike most other hooks, this hook triggers once if there are any changed files
...
+          # The following two options "--custom-marker-begin" and "--custom-marker-end" are ignored if "--use-standard-markers" is set to false
+        - --hook-config=--custom-marker-begin           # String. Defaults to "<!-- BEGIN_TF_DOCS -->" (v1.93+), "<!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->" (<v1.93).
+                                                        # Set to use custom marker which helps you with using other formats like asciidoc.
+                                                        # For Asciidoc this could be "--hook-config=--custom-marker-begin=// BEGIN_TF_DOCS"
+        - --hook-config=--custom-marker-end             # String. Defaults to "<!-- END_TF_DOCS -->" (v1.93+), "<!-- END OF PRE-COMMIT-TERRAFORM DOCS HOOK -->" (<v1.93).
+                                                        # Set to use custom marker which helps you with using other formats like asciidoc.
+                                                        # For Asciidoc this could be "--hook-config=--custom-marker-end=// END_TF_DOCS"
```
yermulnik, line 617: Maybe have usage example clarified a bit? Because from the example it is not straightforward that it should have a value
```suggestion
        - --hook-config=--custom-marker-begin="[...]"   # String. Defaults to "<!-- BEGIN_TF_DOCS -->" (v1.93+), "<!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->" (<v1.93).
                                                        # Set to use custom marker which helps you with using other formats like asciidoc.
                                                        # For Asciidoc this could be "--hook-config=--custom-marker-begin=// BEGIN_TF_DOCS"
        - --hook-config=--custom-marker-end="[...]"     # String. Defaults to "<!-- END_TF_DOCS -->" (v1.93+), "<!-- END OF PRE-COMMIT-TERRAFORM DOCS HOOK -->" (<v1.93).
                                                        # Set to use custom marker which helps you with using other formats like asciidoc.
                                                        # For Asciidoc this could be "--hook-config=--custom-marker-end=// END_TF_DOCS"
```

README.md
```diff
@@ -608,6 +608,16 @@ Unlike most other hooks, this hook triggers once if there are any changed files
...
+        - --hook-config=--custom-doc-header             # String. Defaults to "# "
+                                                        # Set to use custom marker which helps you with using other formats like asciidoc.
+                                                        # For Asciidoc this could be "--hook-config=--custom-marker-end=\= "
```
yermulnik, line 620: To be honest use case of this arg isn't clear to me as a reader of this usage example =(

hooks/terraform_docs.sh
```diff
@@ -115,6 +116,18 @@ function terraform_docs {
...
+        insertion_marker_begin=$value
+        common::colorify "yellow" "INFO: --custom-marker-begin is used and the marker is set to \"$value\"."
+        ;;
+      --custom-marker-end)
+        insertion_marker_end=$value
+        common::colorify "yellow" "INFO: --custom-marker-end is used and the marker is set to \"$value\"."
+        ;;
+      --custom-doc-header)
+        doc_header=$value
+        common::colorify "yellow" "INFO: --custom-doc-header is used and the marker is set to \"$value\"."
```
yermulnik, line 129:
```suggestion
        common::colorify "yellow" "INFO: --custom-doc-header is used and the doc header is set to \"$value\"."
```

README.md
yermulnik: also clarify usage example
```suggestion
        - --hook-config=--custom-doc-header="[...]"     # String. Defaults to "# "
                                                        # Set to use custom marker which helps you with using other formats like asciidoc.
                                                        # For Asciidoc this could be "--hook-config=--custom-marker-end=\= "
```

yermulnik: also clarify usage example
```suggestion
        - --hook-config=--custom-doc-header="[...]"     # String. Defaults to "# "
                                                        # Set to use custom marker which helps you with using other formats like asciidoc.
                                                        # For Asciidoc this could be "--hook-config=--custom-marker-end=\= "
```

hooks/terraform_docs.sh
MaxymVlasov: ah yes. Let's make it "green"

hooks/terraform_docs.sh
```diff
@@ -115,6 +116,18 @@ function terraform_docs {
         common::colorify "yellow" "WARNING: --use-standard-markers is deprecated and will be removed in the future."
         common::colorify "yellow" "         All needed changes already done by the hook, feel free to remove --use-standard-markers setting from your pre-commit config"
         ;;
+      --custom-marker-begin)
+        insertion_marker_begin=$value
+        common::colorify "yellow" "INFO: --custom-marker-begin is used and the marker is set to \"$value\"."
+        ;;
+      --custom-marker-end)
+        insertion_marker_end=$value
+        common::colorify "yellow" "INFO: --custom-marker-end is used and the marker is set to \"$value\"."
```
MaxymVlasov, line 125:
```suggestion
        common::colorify "green" "INFO: --custom-marker-begin is used and the marker is set to \"$value\"."
        ;;
      --custom-marker-end)
        insertion_marker_end=$value
        common::colorify "green" "INFO: --custom-marker-end is used and the marker is set to \"$value\"."
```

hooks/terraform_docs.sh
tarfu: I'm just on my phone today and maybe tomorrow, please feel free to just commit your changes if you want to 🙂

README.md
MaxymVlasov: Previously we in examples we set it to defaults. So, let's do the same here
```suggestion
        - --hook-config=--custom-marker-begin=<!-- BEGIN_TF_DOCS -->  # String. Defaults to "<!-- BEGIN_TF_DOCS -->".
                                                        # Set to use custom marker which helps you with using other formats like asciidoc.
                                                        # For Asciidoc this could be "--hook-config=--custom-marker-begin=// BEGIN_TF_DOCS"
        - --hook-config=--custom-marker-end=<!-- END_TF_DOCS -->      # String. Defaults to "<!-- END_TF_DOCS -->".
                                                        # Set to use custom marker which helps you with using other formats like asciidoc.
                                                        # For Asciidoc this could be "--hook-config=--custom-marker-end=// END_TF_DOCS"
```

MaxymVlasov: Also, we can skip `Defaults to "<!-- END_TF_DOCS -->" (v1.93+), "<!-- END OF PRE-COMMIT-TERRAFORM DOCS HOOK -->" (<v1.93)` as it duplication of what we have paragraph above.

Plus, I don't think that it should be highlighted much, as migration happens w/o any user interaction

MaxymVlasov:
Amazing job, thank you.
Btw, as we are here, just a "quick interview": What would be easier for you to contribute: hooks on bash or on Python? If there are any preferences, ofc.
Structure for hooks will be the same, in terms that all hook-specific code will be located in the same file, similar to what we have in bash right now.

hooks/terraform_docs.sh
MaxymVlasov: Will do then :)

tarfu:
I think python would be more comfortable and would enable to parse yaml for example to have less redundant things in config and arguments of tools.

yermulnik (approved):

antonbabenko:

PR #769, opened 2025-01-14
ci: Avoid reaching GH limits on MacOS runners

MaxymVlasov:
We are hit limits for how many runs on MacOS runners can be run, just by creation 10 renovate PRs and fix twise issues in half of them ~= 20 new PRs/commits for check

Each commit trigger 10 MacOS runs, which means that limit is aproximatly ~200 MacOS runs per hour.

I didn't find such limits in offical docs, btw
https://docs.github.com/en/actions/administering-github-actions/usage-limits-billing-and-administration#usage-limits

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -3,4 +3,6 @@
   extends: ["local>SpotOnInc/renovate-config"],
   // To make happy 'Validate PR title' GHA
   commitMessageLowerCase: "never",
+  // Disable auto-rebase on every commit to avoid reaching Github limits on macos runners
+  rebaseWhen: "conflicted",
 }


yermulnik (approved):

webknjaz:
A better solution would be setting up `concurrency:` like it's done in `ci-cd.yml` so that older workflow runs would be auto-cancelled upon rebase or new commits pushed into the non-master branches.

Additionally, you should really exclude those branches in https://github.com/antonbabenko/pre-commit-terraform/blob/e01b4de/.github/workflows/ci-cd.yml#L8C5-L14C71 since every renovate commit generates 2 events — `push` and `pull_request`, and each triggers a separate workflow run. You're basically letting it run the CI twice for no reason.

yermulnik:
> A better solution would be setting up `concurrency:`

Good point! My bad that I missed to think of it despite I have concurrency configured myself for GHA workflows at my current job 🤦🏻

MaxymVlasov:
@webknjaz Renovate is a Github App, not GH workflow. There are no ability to specify `concurrency:` AFAIK.

>Additionally, you should really exclude those branches in [e01b4de/.github/workflows/ci-cd.yml#L8C5-L14C71](https://github.com/antonbabenko/pre-commit-terraform/blob/e01b4de/.github/workflows/ci-cd.yml?rgh-link-date=2025-01-15T15%3A13%3A46Z#L8C5-L14C71) since every renovate commit generates 2 events — push and pull_request, and each triggers a separate workflow run. You're basically letting it run the CI twice for no reason.

Will do, tnx. Btw, why it runs on commits which are not in PR nor `master` branch? What case it covers?
Usually you don't want to run checks till you're ready to create PR, as you can just push changes to remote just in case your laptop will burn down soon

webknjaz:
Usually, in well-established projects, their upstream Git trees aren't polluted with short-living topic branches so it doesn't matter that pushes to other branches are processed since they are all in forks.
Though, regardless of that, it'd typically a good idea to have CI run on things that don't yet have PRs created. This is also useful to any contributors from forks — they'll call their branches however they like, not master, and the CI should be able to run there.
Another event subtype that falls under this category is tag pushes.

antonbabenko:

PR #776, opened 2025-01-22
ci: Fix edge case with Nightly builds right after new release

MaxymVlasov:

Each failed nightly build occurs right when new tag released and tag point to latest commit.

To avoid that, we need to drop tag to HEAD in CI
image
image

### How can we test changes

1. Fork repo with all branches
2. Create PR from this branch and merge to fork `master`
3. Manually make new release
4. Go to Actions -> Enable :hourglass:  action
5. Trigger :hourglass: action manually
https://github.com/MaxymVlasov/pre-commit-terraform-PR776/actions/runs/12916272093

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -334,6 +334,15 @@ jobs:
         --skip-missing-interpreters false
         --notest
 
+    - name: Drop Git tags from HEAD for non-tag-create events
+      if: >-
+        !fromJSON(needs.pre-setup.outputs.release-requested)
+      run: >-
+        git tag --points-at HEAD
+        |
+        xargs git tag --delete
+      shell: bash
+
     - name: Set static timestamp for dist build reproducibility
       # ... from the last Git commit since it's immutable
       run: >-
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -335,8 +335,6 @@ jobs:
         --notest
 
     - name: Drop Git tags from HEAD for non-tag-create events
-      if: >-
-        !fromJSON(needs.pre-setup.outputs.release-requested)
       run: >-
         git tag --points-at HEAD
         |


.github/workflows/ci-cd.yml
```diff
@@ -334,6 +334,15 @@ jobs:
         --skip-missing-interpreters false
         --notest
 
+    - name: Drop Git tags from HEAD for non-tag-create events
+      if: >-
+        !fromJSON(needs.pre-setup.outputs.release-requested)
```
MaxymVlasov, line 339:
```suggestion
```

.github/workflows/ci-cd.yml
```diff
@@ -334,6 +334,13 @@ jobs:
         --skip-missing-interpreters false
         --notest
 
+    - name: Drop Git tags from HEAD for non-tag-create events
+      run: >-
+        git tag --points-at HEAD
+        |
+        xargs git tag --delete
```
yermulnik, line 341: What's the point in wrapping this one-liner to multiline?
For me it looks odd when split to three lines 🤷🏻 
```suggestion
      run: git tag --points-at HEAD | xargs git tag --delete
```

yermulnik (approved):

.github/workflows/ci-cd.yml
webknjaz: Because it shows separate chained commands more clearly. At it's also synched from other projects. The desire to diverge from that boilerplate was what contributed to this being missed in the original integration.

I sometimes do 
```yaml
run: >-
  cmd1
  | cmd2
  | cmd3
```
and it works well.

But in this case, I've gone more for more lines.

I think that the synchronized changes should be kept as is, as much as possible, to avoid similar problems. It also allows running a `diff` against files in two repositories when synchronizing new changes. I often do an `nvim -d` to copy bits over. The more divergence there is, the harder it is to figure out what to sync and what should be disregarded.

So I suggest not reformatting this.

.github/workflows/ci-cd.yml
webknjaz: This might need to be resurrected in the future. But for now, the change is correct.

webknjaz (approved):

.github/workflows/ci-cd.yml
yermulnik: > Because it shows separate chained commands more clearly.

Not for me 🤷🏻 This bit of code made me stop by and think wth is this about.

> I sometimes do

This one is a way better I think. Even despite I see no value in wrapping small one-liners into multiline that distracts attention from what the command does. But that's my personal opinion and preference vs you two 😂

MaxymVlasov: I preffer oneliner, but webknjaz have plans to improve reusable workflow during 2025, so let it as is for now, for simplifying his future work. If improvements do not come, we will refactor it in a year to what we want :)

yermulnik: Works for me 🤝

antonbabenko:
This PR is included in version 1.97.1 :tada:

Issue #712, opened 2024-09-03
Improve OpenSSF score


nitrocode:
### What problem are you facing?

Adoption in a new organization

### How could pre-commit-terraform help solve your problem?

Renovatebot includes an openssf score on every PR update for this repo. Due to low scores, this can irk developers and management.

Please consider improving the OpenSSF score of this repo. Current score is 6.7 which is not and could be better. The higher the score, the more objective integrity the community will have towards the project.

https://github.com/ossf/scorecard

https://securityscorecards.dev/viewer/?uri=github.com/antonbabenko/pre-commit-terraform

Some small improvements
- Add OpenSSF Best Practices Badge
- Use hadolint and shellcheck to pin dependencies
- Token Permissions in `.github/workflows/*` would improve it a lot
- etc

Some big improvements
- Create official releases and sign them
- etc

MaxymVlasov:
@nitrocode can you please point me where you find such score in Renovate PRs for pre-commit hooks?

![image](https://github.com/user-attachments/assets/bb69209a-2bdc-45e7-8f35-dfc65020c84d)

MaxymVlasov:
We definitely want 9+/10, but firstly I need to understand how to enable such scores for Renovate, as I never disable it in https://github.com/SpotOnInc/renovate-config/blob/main/default.template.json5

nitrocode:
Also the results may be better by aadopting the GitHub action. This should get the branch protections

https://github.com/ossf/scorecard-action

MaxymVlasov:
This can be partially mitigated by https://app.stepsecurity.io/securerepo, as it can provision part of stuff automatically

nitrocode:
StepSecurity is very nice at quickly improving some areas that OpenSSF scorecard detects, so once the StepSecurity PR is merged, the OpenSSF score should also increase.

I think it would still be valuable to integrate the OpenSSF scorecard banner to showcase the score and add the action to help improve the score.

Since opening this, we have seen the score improve from 6.7 to 7.0 😄

nitrocode:
Oh Thanks Maxym, I just noticed your PRs #777 and #780.

yermulnik:
> Since opening this, we have seen the score improve from 6.7 to 7.0 😄

Should we try and close/re-open this more times to get score improved even further? 🤣

MaxymVlasov:
Lol, nah. I'll add a badge to README, that will report status every day and on each commit to `master`, to simplify tracking of it

[...]

PR #785, opened 2025-01-26
chore: Add Open Source Security Foundation best practices badge

MaxymVlasov:
Relates #712


yermulnik (dismissed):

yermulnik (approved):
LGTM.
A side note: I'm wondering why the links and their references (marked with red arrows) are placed not nearby but are separated by an block of text (enclosed in a red rectangle)? 🤔 
image

MaxymVlasov:
> LGTM. A side note: I'm wondering why the links and their references (marked with red arrows) are placed not nearby but are separated by an block of text (enclosed in a red rectangle)? 🤔 image

In theory, it makes sense to move these links right to the bottom of file, but I thought that it will be too much, so I left them in the end of "intro" block

yermulnik:
> > LGTM. A side note: I'm wondering why the links and their references (marked with red arrows) are placed not nearby but are separated by an block of text (enclosed in a red rectangle)? 🤔
> 
> In theory, it makes sense to move these links right to the bottom of file, but I thought that it will be too much, so I left them in the end of "intro" block

Why don't we put them right under where they are referred at? As in move the red rectangle down under the list of links to images for banners?

yermulnik:
Do we really need to split banners from their images?
That should just work w/o quirks like this: [Stand With Ukraine](https://stand-with-ukraine.pp.ua)
```
[![Stand With Ukraine](https://raw.githubusercontent.com/vshymanskyy/StandWithUkraine/main/badges/StandWithUkraine.svg)](https://stand-with-ukraine.pp.ua)
```

MaxymVlasov:
Readability matters, even here

yermulnik:
> Readability matters, even here

Well, while I do agree with the statement, the current placement of the related links (separated by an arbitrary block of data) and the approach as a whole (separation of links and images) isn't readable to me 🤷🏻

PR #799, opened 2025-01-30
fix: Parallelism CPU calculation inside Kubernetes and Docker with limits

daniel-sampliner:
The value of /sys/fs/cgroup/cpu/cpu.cfs_quota_us is not in milliseconds and cannot be simply divided by 1000 to determine the CPU limit. As per kernel documentation[^1], the cpu limit can be determined by dividing that value by /sys/fs/cgroup/cpu/cpu.cfs_period_us.

[^1]: https://docs.kernel.org/scheduler/sched-bwc.html

- [X] This PR fixes a bug.

Fixes the max CPU calculation when run inside kubernetes.

### How can we test changes

1. Run pre-commit inside kubernetes pod with CPU limit N.
2. Without specifying `--parallelism-limit`, confirm that pre-commit is only spawning N-1 concurrent processes.

## Summary by CodeRabbit

- **Bug Fixes**
    - Improved CPU core detection mechanism for more accurate resource allocation.
    - Enhanced compatibility with cgroup-based environments.
    - Updated logic for calculating CPU cores based on new variables for quota and period.

### DIFF
--- a/hooks/_common.sh
+++ b/hooks/_common.sh
@@ -204,8 +204,9 @@ function common::get_cpu_num {
     ! -f /proc/sys/fs/binfmt_misc/WSLInterop ]]; then # WSL have cfs_quota_us, but WSL should be checked as usual Linux host
     # Inside K8s pod or DinD in K8s
     cpu_quota=$(< /sys/fs/cgroup/cpu/cpu.cfs_quota_us)
+    cpu_period=$(cat /sys/fs/cgroup/cpu/cpu.cfs_period_us || echo "$cpu_quota")
 
-    if [[ $cpu_quota -eq -1 ]]; then
+    if [[ $cpu_quota -eq -1 || $cpu_period -lt 1 ]]; then
       # K8s no limits or in DinD
       if [[ -n $parallelism_ci_cpu_cores ]]; then
         if [[ ! $parallelism_ci_cpu_cores =~ ^[[:digit:]]+$ ]]; then
@@ -234,7 +235,6 @@ function common::get_cpu_num {
       return
     fi
 
-    cpu_period=$(< /sys/fs/cgroup/cpu/cpu.cfs_period_us)
     cpu_num=$((cpu_quota / cpu_period))
     [[ $cpu_num -lt 1 ]] && echo 1 || echo $cpu_num
     return
--- a/hooks/_common.sh
+++ b/hooks/_common.sh
@@ -188,6 +188,11 @@ function common::is_hook_run_on_whole_repo {
 
 #######################################################################
 # Get the number of CPU logical cores available for pre-commit to use
+
+# CPU quota should be calculated as `cpu.cfs_quota_us / cpu.cfs_period_us`
+# For K8s see: https://docs.kernel.org/scheduler/sched-bwc.html
+# For Docker see: https://docs.docker.com/engine/containers/resource_constraints/#configure-the-default-cfs-scheduler
+#
 # Arguments:
 #  parallelism_ci_cpu_cores (string) Used in edge cases when number of
 #    CPU cores can't be derived automatically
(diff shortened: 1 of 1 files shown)


hooks/_common.sh
```diff
@@ -233,7 +234,13 @@ function common::get_cpu_num {
...
     fi
 
-    echo $((millicpu / 1000))
+    cpu_period=$(< /sys/fs/cgroup/cpu/cpu.cfs_period_us)
+    cpu_num=$((cpu_quota / cpu_period))
+    if ((cpu_num < 1)); then
+      echo 1
+    else
+      echo $cpu_num
+    fi
```
yermulnik, line 243:
```suggestion
    [[ $cpu_num -lt 1 ]] && echo 1 || echo $cpu_num
```

yermulnik, line 243: @daniel-sampliner Do you think that there's no need for `echo $((millicpu / 1000))` as e.g. a fallback if `/sys/fs/cgroup/cpu/cpu.cfs_period_us` does not exist (in case it may not exist at all)? 🤔

daniel-sampliner: Good question. I don't believe it's possible for `cpu.cfs_period_us` to not exist if `cpu.cfs_quota_us` does exist and is > 0. But I also don't know for certain.

It definitely can't hurt to fall back to something if `cpu.cfs_period_us` doesn't exist, but I don't think `millicpu / 1000` would be the right value. Perhaps like this to default to 1 CPU when `cpu.cfs_period_us` doesn't exist

```suggestion
    cpu_period=$(< /sys/fs/cgroup/cpu/cpu.cfs_period_us || echo -1)
    cpu_num=$((cpu_quota / cpu_period))
    [[ $cpu_num -lt 1 ]] && echo 1 || echo $cpu_num
```

yermulnik: Yep, makes sense to me.
@MaxymVlasov WDYT?

hooks/_common.sh
```diff
@@ -197,14 +197,16 @@ function common::is_hook_run_on_whole_repo {
...
 
+  local cpu_quota cpu_period cpu_num
   local millicpu
 
   if [[ -f /sys/fs/cgroup/cpu/cpu.cfs_quota_us &&
     ! -f /proc/sys/fs/binfmt_misc/WSLInterop ]]; then # WSL have cfs_quota_us, but WSL should be checked as usual Linux host
     # Inside K8s pod or DinD in K8s
-    millicpu=$(< /sys/fs/cgroup/cpu/cpu.cfs_quota_us)
+    cpu_quota=$(< /sys/fs/cgroup/cpu/cpu.cfs_quota_us)
+    cpu_period=$(cat /sys/fs/cgroup/cpu/cpu.cfs_period_us || echo "$cpu_quota")
```
yermulnik, line 207: It makes sense to suppress error
```suggestion
    cpu_period=$(cat /sys/fs/cgroup/cpu/cpu.cfs_period_us 2> /dev/null || echo "$cpu_quota")
```

yermulnik (commented):
LGTM. @MaxymVlasov WDYT?

hooks/_common.sh
```diff
@@ -233,12 +235,14 @@ function common::get_cpu_num {
...
+    [[ $cpu_num -lt 1 ]] && echo 1 || echo $cpu_num
     return
   fi
 
   if [[ -f /sys/fs/cgroup/cpu.max ]]; then
     # Inside Linux (Docker?) container
+    local millicpu
     millicpu=$(cut -d' ' -f1 /sys/fs/cgroup/cpu.max)
 
     if [[ $millicpu == max ]]; then
```
MaxymVlasov, line 248: Regarding thread above about fallbacking to millicpu
In docker we have next picture:
https://docs.docker.com/engine/containers/resource_constraints/#configure-the-default-cfs-scheduler

```
➜ docker run -ti --cpus="1.5" alpine sh
/ # grep ^ /dev/null /sys/fs/cgroup/cpu*
/sys/fs/cgroup/cpu.idle:0
/sys/fs/cgroup/cpu.max:150000 100000
```

Where 150000 is `cpu-quota` and 100000 is `cpu-period`

`150000 / 1000 (millicpu) != 1.5`, so, I'll fix it in this PR too

MaxymVlasov (approved):
Great findings

hooks/_common.sh
```diff
@@ -188,6 +188,11 @@ function common::is_hook_run_on_whole_repo {
 
 #######################################################################
 # Get the number of CPU logical cores available for pre-commit to use
+
```
MaxymVlasov, line 191:
```suggestion
#
```

yermulnik (approved):

antonbabenko:
This PR is included in version 1.97.1 :tada:

PR #818, opened 2025-02-14
ci: Integrate zizmor checks to repo

MaxymVlasov:
That PR integrate one more static analysis tool, as CodeQL is good, but it not cover all cases
Here is an article about zizmor value - https://blog.yossarian.net/2024/12/06/zizmor-ultralytics-injection
zizmor repo - https://github.com/woodruffw/zizmor


MaxymVlasov:
Oh, I see, there a bunch of issues in workflows to be fixed first

yermulnik (dismissed):
Hope at some point we won't end up running several dozens of GH actions over the handful of scripts that we maintain here 🤣

.github/workflows/ci-cd.yml
```diff
@@ -218,13 +221,13 @@ jobs:
                 mode=FILE_APPEND_MODE,
         ) as outputs_file:
             print(
-                "tag=v${{
-                    steps.scm-version.outputs.dist-version
-                }}",
+                f"tag=v{environ['DIST_VERSION']}",
```
yermulnik: When version gets unconditionally added a `v` prefix, I always suggest to unconditionally trim `v` prefix from source string (the `environ['DIST_VERSION']` in this case) to just always stay on the safe side and to not end up with `vvX.X.X` in any case. JFYI and up to you to consider.

yermulnik: Sort of this:
```suggestion
                f"tag=v{environ['DIST_VERSION'].removeprefix('v')}",
```

.github/workflows/pre-commit.yaml
```diff
@@ -14,17 +14,20 @@ jobs:
...
       id: file_changes
+      env:
+        BASE_REF: ${{ github.base_ref }}
+        SHA: ${{ github.sha }}
       run: |
-        export DIFF=$(git diff --name-only origin/${{ github.base_ref }} ${{
-          github.sha
-        }})
-        echo "Diff between ${{ github.base_ref }} and ${{ github.sha }}"
+        export DIFF=$(git diff --name-only origin/$BASE_REF $SHA)
```
yermulnik: It is better to wrap vars in double quotes to prevent globbing and word splitting (if any):
```suggestion
        export DIFF=$(git diff --name-only "origin/$BASE_REF" "$SHA")
```

.github/workflows/pre-commit.yaml
```diff
@@ -48,10 +51,11 @@ jobs:
         )"
         > hadolint
         && chmod +x hadolint && sudo mv hadolint /usr/bin/
-     # Needed for pre-commit fix push to succeed
+      # Needed for pre-commit fix push to succeed
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
```
yermulnik: I might be wrong, though I though it should be aligned at start of the string rather than at start of the word
```suggestion
    # Needed for pre-commit fix push to succeed
    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
```

.github/workflows/reusable-tox.yml
```diff
@@ -272,19 +274,25 @@ jobs:
...
-        --parallel auto
-        --parallel-live
-        --skip-missing-interpreters false
-        ${{
-          inputs.built-wheel-names != ''
-          && format('--installpkg dist/{0}', inputs.built-wheel-names)
-          || ''
-        }}
-        --notest
+        Pre-populate tox envs: ${{ env.TOXENV }}
```
yermulnik: Wrapping in backticks would make that string be shown as monospaced text in output to make it different from the rest of the regular text on this line. Ain't it good? 🤔

.github/workflows/reusable-tox.yml
```diff
@@ -272,19 +274,25 @@ jobs:
...
+      shell: bash
+      env:
+        INPUT_BUILT_WHEEL_NAMES: ${{ inputs.built-wheel-names }}
+      run: |-
+        tox_common_args=(
+          --parallel auto
+          --parallel-live
+          --skip-missing-interpreters false
+        )
+        if [ -n "$INPUT_BUILT_WHEEL_NAMES" ]; then
```
yermulnik: With `shell: bash` this probably can use Bashism:
```suggestion
        if [[ $INPUT_BUILT_WHEEL_NAMES ]]; then
```
Else there should be not reason for Bash specifically (POSIX compatible `/bin/sh` should be okay).

.github/workflows/reusable-tox.yml
```diff
@@ -352,13 +365,15 @@ jobs:
         && steps.tox-run.outputs.test-result-files == ''
         && steps.tox-run.outputs.codecov-flags != 'MyPy'
       run: >-
-        cat code-coverage-results.md >> "${GITHUB_STEP_SUMMARY}"
+        cat code-coverage-results.md >> $GITHUB_STEP_SUMMARY
```
yermulnik: While curly brackets are indeed redundant here (while shell formatting tools still do recommend them as a general rule), wrapping var in double quotes to avoid globbing and word splitting in shell commands is good thing (despite GH docs don't care about this).
Up to you to consider as I think GH probably ensures GH-specific vars don't have globbing or splitting chars, and hence no commit suggestion 🤷🏻

.pre-commit-config.yaml
```diff
@@ -47,6 +47,12 @@ repos:
   hooks:
   - id: gitleaks
 
+# Github Action static analysis tool
+- repo: https://github.com/woodruffw/zizmor-pre-commit
+  rev: v1.9.0
```
yermulnik, line 52: Tag and not commit hash? I thought you tend to trust commits rather than tags 🤔

yermulnik (approved):
Approved as technically there seem to be no issues.
Please consider comments and commit suggestions as you think they worth it.

.github/workflows/reusable-tox.yml
MaxymVlasov:
```suggestion
        cat code-coverage-results.md >> "$GITHUB_STEP_SUMMARY"
```

.github/workflows/reusable-tox.yml
MaxymVlasov:
```suggestion
        Pre-populate tox envs: `${{ env.TOXENV }}`
```

.github/workflows/reusable-tox.yml
```diff
@@ -299,21 +307,26 @@ jobs:
     # But only for 'pytest' env in 'tox'.
     # For details: ../../tox.ini '[testenv:pytest]' 'commands_post'
     - name: >-
-        Run tox envs: `${{ env.TOXENV }}`
+        Run tox envs: ${{ env.TOXENV }}
```
MaxymVlasov:
```suggestion
        Run tox envs: `${{ env.TOXENV }}`
```

.github/workflows/reusable-tox.yml
MaxymVlasov: I debugged one warning and forget to restore this behavior. Good catch

.github/workflows/ci-cd.yml
MaxymVlasov: Good suggestion. While I'm pretty sure that there will be no `v` prefix at all, better to have guardrails than hope that they will be never needed

.pre-commit-config.yaml
MaxymVlasov: We need to switch out from Renovate to pre-commit-freeze then.

I'll do it as separate PR

.github/workflows/reusable-tox.yml
```diff
@@ -352,13 +365,15 @@ jobs:
         && steps.tox-run.outputs.test-result-files == ''
         && steps.tox-run.outputs.codecov-flags != 'MyPy'
       run: >-
-        cat code-coverage-results.md >> "${GITHUB_STEP_SUMMARY}"
+        cat code-coverage-results.md >> "$GITHUB_STEP_SUMMARY"
```
yermulnik, line 368: I'd say it's best to either wrap in double quotes all such GH vars or none. Otherwise those not wrapped look unsafe (while most probably they won't have any special chars ever I guess, but still unsafe 🤪) =)

yermulnik (approved):
A mug of weird black magic for me 🤣 Overall looks good to me in those bits that I was able to understand =)))

MaxymVlasov:
GitHub failed to run workflows or what -_-


Ah, I see
image

antonbabenko:
This PR is included in version 1.99.2 :tada:

.pre-commit-config.yaml
webknjaz: I think native `pre-commit autoupdate` can do the same.

.github/workflows/reusable-tox.yml
webknjaz: I'm sure Max got confused here — this doesn't look like something Zizmor would complain about.

.pre-commit-config.yaml
webknjaz, line 1: @MaxymVlasov FYI, I got a reusable workflow that integrates with GitHub's security reporting interfaces natively: https://github.com/zizmorcore/workflow/blob/3bb5e95068d0f44b6d2f3f7e91379bed1d2f96a8/.github/workflows/reusable-zizmor.yml. If you add it, you'll get inline alerts right inside PRs.

PR #819, opened 2025-02-14
chore: Set emojis that better represent what happens in GH workflows

MaxymVlasov:


yermulnik (approved):
I'm not entirely happy with the purple'ish background color on CI/CD sign emoji 🤣

MaxymVlasov:
It's blue on Firefox on Ubuntu

But I see it as purple on Android -__-

MaxymVlasov:
Like this one for twitter
https://emojipedia.org/infinity#designs

Need to check is there a way to force specific style

MaxymVlasov:
We can try:
* ∞
* ᨖ
* ♾
* ᪲᪲᪲

Just need to test how it looks like on different OS and browsers

yermulnik:
Chrome 🤮
image
FF 😍
image

> Just need to test how it looks like on different OS and browsers

The very last looks broken on both browsers for me:
image

The 1st and 2nd look ok =)

MaxymVlasov:
FF on Android
Screenshot_20250215-020127

GH mobile (probably chrome-based)
Screenshot_20250215-020344

antonbabenko:

PR #822, opened 2025-02-20
feat: Include .tfbackend files in hooks

thiagowfx:
Blocked as there Upstream FR / bug: https://github.com/hashicorp/terraform/issues/36564

---

- [ ] This PR fixes a bug.

As per the official docs https://developer.hashicorp.com/terraform/language/backend#file:

> A backend configuration file has the contents of the backend block as top-level attributes, without the need to wrap it in another terraform or backend block:
>
> ```
> address = "demo.consul.io"
> path    = "example_app/terraform_state"
> scheme  = "https"
> ```
>
> *.backendname.tfbackend (e.g. config.consul.tfbackend) is the recommended naming pattern. Terraform will not prevent you from using other names but following this convention will help your editor understand the content and likely provide better editing experience as a result.

### How can we test changes

Add `foo.tfbackend` (a simple HCL file, without blocks, similar to `.tfvars`), run e.g. `pre-commit run --all-files terraform_fmt`.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -13,7 +13,7 @@
     Rewrites all Terraform configuration files to a canonical format.
   entry: hooks/terraform_fmt.sh
   language: script
-  files: (\.tf|\.tfvars)$
+  files: (\.tf|\.tfbackend|\.tfvars)$
   exclude: \.terraform/.*$
 
 - id: terraform_docs
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -4,7 +4,7 @@
   entry: hooks/infracost_breakdown.sh
   language: script
   require_serial: true
-  files: \.(tf(vars)?|hcl)$
+  files: \.(tf(backend|vars)?|hcl)$
   exclude: \.terraform/.*$
 
 - id: terraform_fmt
@@ -53,7 +53,7 @@
   require_serial: true
   entry: hooks/terraform_validate.sh
   language: script
-  files: \.(tf(vars)?|terraform\.lock\.hcl)$
+  files: \.(tf(backend|vars)?|terraform\.lock\.hcl)$
   exclude: \.terraform/.*$
 
 - id: terraform_providers_lock
@@ -71,7 +71,7 @@
   require_serial: true
   entry: hooks/terraform_tflint.sh
   language: script
-  files: (\.tf|\.tfvars)$
+  files: \.tf(backend|vars)?$
   exclude: \.terraform/.*$
 
 - id: terragrunt_fmt
@@ -114,7 +114,7 @@
     Static analysis of Terraform templates to spot potential security issues.
   require_serial: true
   entry: hooks/terraform_tfsec.sh
-  files: \.tf(vars)?$
+  files: \.tf(backend|vars)?$
   language: script
 
 - id: terraform_trivy
@@ -123,7 +123,7 @@
     Static analysis of Terraform templates to spot potential security issues.
   require_serial: true
   entry: hooks/terraform_trivy.sh
-  files: \.tf(vars)?$
+  files: \.tf(backend|vars)?$
   language: script
 
 - id: checkov
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -13,7 +13,7 @@
     Rewrites all Terraform configuration files to a canonical format.
   entry: hooks/terraform_fmt.sh
   language: script
-  files: (\.tf|\.tfbackend|\.tfvars)$
+  files: \.tf(backend|vars)?$
   exclude: \.terraform/.*$
 
 - id: terraform_docs
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -13,7 +13,7 @@
     Rewrites all Terraform configuration files to a canonical format.
   entry: hooks/terraform_fmt.sh
   language: script
-  files: \.tf(backend|vars)?$
+  files: \.tf(vars)?$
   exclude: \.terraform/.*$
 
 - id: terraform_docs


MaxymVlasov:
Hi, so it is a new possibility in terraform to define backend. Interesting. 

There is a chance that the same changes need to be included in these hooks:
* terraform_docs
* terraform_validate
* terraform_tflint
* terraform_tfsec
* terraform_trivy
* terraform_checkov
* terrascan

But need to test it. Would much appreciate it if you'll handle it. If not - I'll do it later by myself.
Need to check if addition of that file somehow affect behavior of these hooks and if affects - add `.tfbackend` to it too

thiagowfx:
My understanding is that `.tfbackend` files are supposed to have simple key-value pairs:

```
a = "1"
b = "2"
```

...etc.

Formatting them makes sense (`terraform_fmt` likes to align the equal signs vertically).

I am not completely sure whether the other hooks make sense, would need to look deeper.

I would be happy to send more PRs later on if it turns out they make sense.
For now, I changed only `fmt` because I am confident it's applicable.

What do you think?

.pre-commit-hooks.yaml
```diff
@@ -13,7 +13,7 @@
     Rewrites all Terraform configuration files to a canonical format.
   entry: hooks/terraform_fmt.sh
   language: script
-  files: (\.tf|\.tfvars)$
+  files: (\.tf|\.tfbackend|\.tfvars)$
```
yermulnik, line 16: Just a nit to improve regex overall while we're here:
```suggestion
  files: \.tf(backend|vars)?$
```

yermulnik (approved):
* TIL about `.tfbackend` files 😲 
* LGTM

yermulnik:
> There is a chance that the same changes need to be included in these hooks:

Do the other hooks take TF backend configuration into account at all? 🤔 If yes, then, yep, we'd better off adding it to those that are aware of this type of files 👍🏻

thiagowfx:
I was thinking about this. `.tfbackend` isn't too different from `.tfvars` (https://spacelift.io/blog/terraform-tfvars).

Perhaps it makes sense to add it to all hooks that have `.tfvars`.

thiagowfx:
Updated the file + PR title + PR description.

yermulnik:
> I was thinking about this. `.tfbackend` isn't too different from `.tfvars` ([spacelift.io/blog/terraform-tfvars](https://spacelift.io/blog/terraform-tfvars)).
> 
> Perhaps it makes sense to add it to all hooks that have `.tfvars`.

Underlying tools infer values for variables from tfvars, whereas I can't say for sure how they will behave encountering "variables" in tfbackend files as these "variables" ain't declared by means of `variable` block definitions. This is my concern and hesitation.

As @MaxymVlasov already inquired: would you be able to add change to `.pre-commit-hooks.yaml` locally and test to see whether none tool get confused with the vars in tfbackend file(s)? Thank you in advance 👍🏻

thiagowfx:
I am only currently using `terraform_fmt` and `terraform_tflint` in our projects. I can test these two and report back (well, `terraform_fmt` is already tested, so there's just `tflint` remaining).

For the other 4 hooks, are there existing samples / tests somewhere?

thiagowfx:
Oh wait, I take it back for `terraform_fmt`:

```
$ terraform fmt a.tfbackend
│ Error: Only .tf, .tfvars, and .tftest.hcl files can be processed with terraform fmt
```

It should be possible to do so.
I'll file an upstream bug.

thiagowfx:
Upstream FR: https://github.com/hashicorp/terraform/issues/36564

yermulnik:
> I am only currently using `terraform_fmt` and `terraform_tflint` in our projects. I can test these two and report back (well, ~`terraform_fmt` is already tested~[1](#user-content-fn-1-8eada0e9d4666c0e0abeddbbf03272c6), so there's just `terraform_tflint` remaining).
> 
> For the other 4 hooks, are there existing samples / tests somewhere I could use for testing?

It should be as simple as adding each of those hooks to your `.pre-commit-config.yaml` file, like this:
```yaml
repos:
  rev: 837b9ae2b24688c9cc3727fcf717c8494c421ee6 # commit that you want to test against
  hooks:
    - id: terraform_fmt
    - id: terraform_validate
    - id: terraform_tflint
    - id: terraform_docs
    - id: terraform_tfsec
    - id: terraform_trivy
    #- id: infracost_breakdown # I don't have a API key :shrugging:
```
So for me none fails (and even `terraform_fmt` with commit id `0dad06983ea30273c52be9c3c443ad5970de0a90`) as they all seem to ignore that type of file and as they operate on a dir rather than on a specific files, hence none throws error.
```shell
# running with `0dad06983ea30273c52be9c3c443ad5970de0a90`
> pre-commit run -a
Terraform fmt...........................................................................Passed
Terraform validate......................................................................Passed
Terraform validate with tflint..........................................................Passed
Terraform docs......................................................(no files to check)Skipped
Terraform validate with tfsec (deprecated, use "terraform_trivy").......................Passed
Terraform validate with trivy...........................................................Passed
```
I'm not sure whether we can enforce each hook to run over specific file(s) instead. @MaxymVlasov can assist perceiving this. Else we should probably be safe with this PR (including TF fmt hook).
I'm deferring the decision to @MaxymVlasov as a main dev here.

MaxymVlasov:
About hook examples and how to test it. 
That 100% out of scope of this PR, but thats all about tests and their automation. I don't think that unit tests (hello python) or fuzzing tests (hello OSSF, and maybe, python too) will help with that much. Probably we will need add some e2e tests by create separate repo with bunch of code samples and bunch isolated hooks runs.

Good news that we can do it right now (and I'll probably do it anyway, glad you asked, as I tired to test changes somehow locally with hope that I cover everything - that's far from ideal)
Bad news - could  be hard to automate it right now, but at least it will be a good starting point.

I can't say ETA when I'll do that, but it will not take more than 1 workday, just need to find that time

PR #824, opened 2025-02-23
chore(linters): Add yamlfmt

MaxymVlasov:
`yamlfmt` does not support inline comments in `key: >-  # comment` construction - it removes such comments.

That's only one incontinence which I see for that linter, so it's good enough to force its use


yermulnik (approved):

antonbabenko:

PR #830, opened 2025-02-26
fix(docker image security): Improve dependency pinning and disable ability to build image from different than specified in dockefile tag

MaxymVlasov:
Relates #712 
Based on https://github.com/ossf/scorecard/issues/3684#issuecomment-2683942964

I am not sure that pining python deps to major version will make any difference for OSSF, but at least we will specify what major versions we currently rely on

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -185,7 +185,7 @@ jobs:
         python -m
         pip install
         --user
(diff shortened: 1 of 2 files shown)


antonbabenko (approved):

yermulnik (approved):

antonbabenko:

PR #832, opened 2025-03-03
chore(linters): Reorder linters, make hadolint ignores more specific

MaxymVlasov:
Split from https://github.com/antonbabenko/pre-commit-terraform/pull/831

Hadolint supports its own config file, and that make able 3rd-party tools that utilize hadolint also known about specified config options, not pre-commit only

### DIFF
--- a/.hadolint.yaml
+++ b/.hadolint.yaml
@@ -1,0 +1,9 @@
+ignored:
+- DL3007  # Using latest
+- DL3013  # Pin versions in pip
+- DL3027  # Do not use apt
+- DL3059  # Docker `RUN`s shouldn't be consolidated here
+- DL4006  # Not related to alpine
+- SC1091  # Useless check
+- SC2015  # Useless check
+- SC3037  # Not related to alpine
--- a/.hadolint.yaml
+++ b/.hadolint.yaml
@@ -1,9 +1,0 @@
-ignored:
-- DL3007  # Using latest
-- DL3013  # Pin versions in pip
-- DL3027  # Do not use apt
-- DL3059  # Docker `RUN`s shouldn't be consolidated here
-- DL4006  # Not related to alpine
-- SC1091  # Useless check
-- SC2015  # Useless check
-- SC3037  # Not related to alpine
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -47,9 +47,13 @@ repos:
   hooks:
   - id: gitleaks
 
-#
-# YAML Linters
-#
+# Dockerfile
+- repo: https://github.com/hadolint/hadolint
+  rev: v2.12.1-beta
+  hooks:
+  - id: hadolint
+
+# YAML
 - repo: https://github.com/jumanjihouse/pre-commit-hook-yamlfmt
   rev: 0.2.3
   hooks:
@@ -71,7 +75,7 @@ repos:
     args:
     - --strict
 
-# JSON5 Linter
+# JSON5
 - repo: https://github.com/pre-commit/mirrors-prettier
   rev: v3.1.0
   hooks:
@@ -79,6 +83,21 @@ repos:
     # https://prettier.io/docs/en/options.html#parser
     files: .json5$
 
+# Bash
+- repo: https://github.com/jumanjihouse/pre-commit-hooks
+  rev: 3.0.0
+  hooks:
+  - id: shfmt
+    args:
+    - -l
+    - -i
+    - '2'
+    - -ci
+    - -sr
+    - -w
+  - id: shellcheck
+
+# Python
 - repo: https://github.com/pre-commit/mirrors-mypy.git
   rev: v1.15.0
   hooks:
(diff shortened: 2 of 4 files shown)


.hadolint.yaml
```diff
@@ -0,0 +1,9 @@
+ignored:
+- DL3007  # Using latest
+- DL3013  # Pin versions in pip
+- DL3027  # Do not use apt
+- DL3059  # Docker `RUN`s shouldn't be consolidated here
+- DL4006  # Not related to alpine
+- SC1091  # Useless check
+- SC2015  # Useless check
+- SC3037  # Not related to alpine
```
yermulnik, line 9: tbh such comments require reader to google what the check is about. would be great to have a more clear comments (either the brief description of the check, or a more clear rationale to ignore it **globally**).
I think it's also best to use inline exclusions per each Dockefile rather than exclude checks globally.

.pre-commit-config.yaml
```diff
@@ -47,9 +47,13 @@ repos:
   hooks:
   - id: gitleaks
 
-#
-# YAML Linters
-#
+# Dockerfile
+- repo: https://github.com/hadolint/hadolint
+  rev: v2.12.1-beta
```
yermulnik, line 52: I might be missing some gist of this, but using beta...

.hadolint.yaml
webknjaz: +1
Comments are for justification, not just naming things. Although, in case of magic numbers, naming is also useful in addition to justification.

.pre-commit-config.yaml
webknjaz: `pre-commit autoupdate` grabs whatever tag is the latest in Git. It does not understand versioning semantics.

MaxymVlasov: And it's "good enough" to be utilized, as it contains only deps updates
image

hadolint is maintained, but there were no releases for 2 years, idk why
Feel free to bump this issue - https://github.com/hadolint/hadolint/issues/1081

tests/Dockerfile
```diff
@@ -1,11 +1,9 @@
...
 FROM pre-commit-terraform:latest
 
-RUN apt update && \
-    apt install -y \
-        datamash \
-        time && \
-    # Cleanup
-    rm -rf /var/lib/apt/lists/*
+RUN apk add --no-cache \
+    datamash=~1.8
```
MaxymVlasov, line 6: I found that we changed base image to alpine a long time ago :)

.pre-commit-config.yaml
webknjaz: No, thanks. Spamming maintainers is unhelpful at best.

tests/Dockerfile
```diff
@@ -1,11 +1,9 @@
+# We use `latest` tag for tests proposes
```
webknjaz, line 1: Tests also deserve to be stable FWIW.

webknjaz (changes requested):

tests/Dockerfile
MaxymVlasov: You need to test your local image. Which you build locally

MaxymVlasov: Hmm, wait a minute, it could be used not for image testing, let me double check

MaxymVlasov: Double-checked - it used just in case if used don't want to install all that toolset locally, but it still requires local image build (by default)
https://github.com/antonbabenko/pre-commit-terraform/blob/5a022c914233bda538ec3c2b8fe8b82f3fe74fa8/.github/CONTRIBUTING.md#L60-L86

.pre-commit-config.yaml
yermulnik: Given the age of both beta and its parent, it makes no point is using beta in favour of its parent 🤪

yermulnik (dismissed):

Dockerfile
```diff
@@ -81,6 +83,10 @@ RUN /install/trivy.sh
 
 
 # Checking binaries versions and write it to debug file
+
+# SC3037 - We do not use `echo` flags here, so it's false-positive
+# DL4006 - Not Applicable for for /bin/sh in alpine images. Disable, as recommended by check itself
```
yermulnik, line 88:
```suggestion
# DL4006 - Not Applicable for /bin/sh in alpine images. Disable, as recommended by check itself
```

Dockerfile
```diff
@@ -81,6 +83,10 @@ RUN /install/trivy.sh
 
 
 # Checking binaries versions and write it to debug file
+
+# SC3037 - We do not use `echo` flags here, so it's false-positive
```
yermulnik, line 87: JFYI:
image

tests/Dockerfile
```diff
@@ -1,11 +1,9 @@
...
+# hadolint ignore=DL3007
 FROM pre-commit-terraform:latest
 
-RUN apt update && \
-    apt install -y \
-        datamash \
-        time && \
-    # Cleanup
-    rm -rf /var/lib/apt/lists/*
```
yermulnik, line 8: Maybe not just drop, but adjust for alpine instead? 🤔 
```suggestion
    # Cleanup
    rm -rf /var/cache/apk/*
```

MaxymVlasov: --no-cache

yermulnik: I meant a double check. Though, well, you're right, probably it makes no sense 👍🏻

Dockerfile
```diff
@@ -81,12 +83,15 @@ RUN /install/trivy.sh
 
 
 # Checking binaries versions and write it to debug file
+
+# SC2086 - We do not need to quote "$F" variable, because it's not contain spaces
```
yermulnik, line 87: It doesn't hurt to wrap in double quotes though and the benefit would be to not skip this check is we, by any chance, will add any new var into this code block.

Dockerfile
yermulnik: Still double "for".

yermulnik (approved):

Dockerfile
MaxymVlasov: If we add new vars here, it should be moved to separate .sh file, as it will become too complicated

yermulnik (approved):

Dockerfile
webknjaz: It's always a good idea to use the `"${VAR}"` syntax for consistency and to prevent unexpected behavior happening due to changes in seemingly unrelated places not having been caught because of some forgotten linting suppression.

yermulnik: Despite the use of curly brackets is also encouraged by at least some shell linters I guess, I'd argue against it as curly brackets denote variable expansion syntax, which is pointless where no expansion is meant or expected.

antonbabenko:
This PR is included in version 1.98.0 :tada:

PR #861, opened 2025-03-25
chore(linters): Apply ruff-format

MaxymVlasov:
Subset of https://github.com/antonbabenko/pre-commit-terraform/pull/831

Result of run

```yaml
repos:
- repo: https://github.com/astral-sh/ruff-pre-commit
  rev: v0.8.4
  hooks:
  - id: ruff-format
```

with same `ruff.toml` as in #831, exept 2 last lines which enable ignore of deprecated hooks (https://github.com/antonbabenko/pre-commit-terraform/blob/948a36d52d36270c329028bd27f7a3e999c899ff/ruff.toml)

PR #862, opened 2025-03-26
chore: Ignore ruff-format apply in git blame

MaxymVlasov:
Follow-up to https://github.com/antonbabenko/pre-commit-terraform/pull/861

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,70 @@
+# `git blame` master ignore list.
+#
+# This file contains a list of git hashes of revisions to be ignored
+# by `git blame`. These revisions are considered "unimportant" in
+# that they are unlikely to be what you are interested in when blaming.
+# They are typically expected to be formatting-only changes.
+#
+# It can be used for `git blame` using `--ignore-revs-file` or by
+# setting `blame.ignoreRevsFile` in the `git config`[1].
+#
+# Ignore these commits when reporting with blame. Calling
+#
+#   git blame --ignore-revs-file .git-blame-ignore-revs
+#
+# will tell `git blame` to ignore changes made by these revisions when
+# assigning blame, as if the change never happened.
+#
+# You can enable this as a default for your local repository by
+# running
+#
+#   git config blame.ignoreRevsFile .git-blame-ignore-revs
+#
+# This will probably be automatically picked by your IDE
+# (VSCode+GitLens and JetBrains products are confirmed to do this).
+#
+# Important: if you are switching to a branch without this file,
+# `git blame` will fail with an error.
+#
+# GitHub also excludes the commits listed below from its "Blame"
+# views[2][3].
+#
+# [1]: https://git-scm.com/docs/git-blame#Documentation/git-blame.txt-blameignoreRevsFile
+# [2]: https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/
+# [3]: https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view
+#
+# Guidelines:
+# - Only large (generally automated) reformatting or renaming PRs
+#   should be added to this list. Do not put things here just because
+#   you feel they are trivial or unimportant. If in doubt, do not put
+#   it on this list.
+# - When adding a single revision, use inline comment to link relevant
+#   issue/PR. Alternatively, paste the commit title instead.
+#   Example:
+#     d4a8b7307acc2dc8a8833ccfa65426ad28b3ffc9  # https://github.com/sanitizers/octomachinery/issues/1
+# - When adding multiple revisions (like a bulk of work over many
+#   commits), organize them in blocks. Precede each such block with a
+#   comment starting with the word "START", followed by a link to the
+#   relevant issue or PR. Add a similar comment after the last block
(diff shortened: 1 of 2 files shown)


webknjaz (approved):
You may want to add a mention on configuring `blame.ignoreRevsFile` in Git to the contributing doc. But this is fine to go in otherwise.

.git-blame-ignore-revs
MaxymVlasov, line 1: @yermulnik wdyt about disabling Copilot till 2026? I can't remember any useful finding from its side

webknjaz (approved):

.git-blame-ignore-revs
yermulnik: Up to you. In the other project it helped me to rectify several oversights and mistakes. So I'm fine with it adding its reviews as they don't hurt I guess.

.git-blame-ignore-revs
```diff
@@ -0,0 +1,70 @@
...
+#   or any other revision reference).
+# - Append to the bottom of the file, regardless of the chronological
+#   order of the revisions. Revisions within blocks should be in
+#   chronological order from oldest to newest.
+# - Because you must use a hash, you need to append to this list in a
+#   follow-up PR to the actual reformatting PR that you are trying to
+#   ignore. This approach helps avoid issues with arbitrary rebases
+#   and squashes while the pull request is in progress.
+
+23928fbf8511697c915c3231977ee254bd3fa0c2  # chore(linters): Apply ruff-format
```
yermulnik, line 70: Maybe add link to corresponding PR as they recommend it in the comment above?
I mean have both: a textual comment and link to PR 🤔

yermulnik (approved):

.git-blame-ignore-revs
MaxymVlasov: I'll leave link in merge commit msg, so it will be simple to check PR too

webknjaz: I compiled those guidelines above from several sources and polished them a bit. I think it's enough to just visually show a commit message + maybe justification. Everything's in Git. Viewing PRs requires internet access. So it's something on top. But they are findable via GH from the commit page, so adding a link on top is probably unnecessary.

yermulnik: My point was to reduce friction: open file, copy link, see what it hidden from git blame VS open file, copy commit id, run `git show commit_id`, copy PR id, go to GH, switch to PRs tab, find PR or similar =)

webknjaz: Not my choice to disallow proper natural merge commits :man_shrugging:

yermulnik: Not entirely sure how this relates to what I mean as my point is not about Git but about people =)

ps: "proper natural merge commits" — you mean non-squashed merges?

webknjaz: > Not entirely sure how this relates to what I mean as my point is not about Git but about people =)
> 

It's a problem of culture of maintenance. Git should contain all the context necessary for offline inspection. And inventing hacks like linking external pointers only deepens this problem. Instead, the context should be included directly.

> ps: "proper natural merge commits" — you mean non-squashed merges?

Yes, `--no-ff` merges that preserve commits with their messages and context. This works when the Git tree is maintained well.

yermulnik: > > Not entirely sure how this relates to what I mean as my point is not about Git but about people =)
> 
> It's a problem of culture of maintenance. Git should contain all the context necessary for offline inspection. And inventing hacks like linking external pointers only deepens this problem. Instead, the context should be included directly.

Erm, you seem to be talking about something very different and unrelated.
This was only about improving user experience when exploring that file.
I'd also admit that, given it's you who authored that file (right?), it was specifically you who suggested to link relevant issue or PR. Or am I misreading it? 😕 
image

> > ps: "proper natural merge commits" — you mean non-squashed merges?
> 
> Yes, `--no-ff` merges that preserve commits with their messages and context. This works when the Git tree is maintained well.

Well, under certain conditions you're absolutely right 👍🏻

yermulnik: > Yeah, maybe it was me.

Maybe? 😕 
image

> Though, I don't think that there's value in links in every single case. OTOH, this is just a boilerplate. It's up to you to decide in each case if you want to add a link. Personally, I think that inline context might be more useful.

What I'm hinting at is that (apart from improved usability where I can just copy URL, if I need to, from the same file I'm looking at instead of digging the same from the git logs) the discrepancy between suggested file format and the format of the entries is **odd** and adds confusion. Why the heck we need all that huge block with guidelines if we don't give a heck to follow it. Empty that and then we follow «_It's up to you to decide in each case_» rule. Else it's just an good example of bad manner and the principle of neglecting standards that we create for ourselves.

webknjaz: Okay, fair.

yermulnik: 🤝

yermulnik: Jesus, mate, you do think that VSCode is the sole **file viewer** in the world (do you indeed mean one must always use VSCode or similar and can't use e.g. `less` to view files?) or that one must click around to just get — ONCE AGAIN — what is recommended by the whole almost 70(sic!) lines preamble in that file? 🤦🏻 
Descriptions are for humans, not for machines! Period.

antonbabenko:

PR #863, opened 2025-03-26
chore(linters): Apply ruff linter semi-automatic linting fixes

MaxymVlasov:
Subset of #831 

Apply next for each violation:
```yaml
repos:
- repo: https://github.com/astral-sh/ruff-pre-commit
  rev: v0.8.4
  hooks:
  - id: ruff
    args:
      - --fix
      - --unsafe-fixes
      - --select=<rulename>
  - id: ruff-format
```
and commit as separate commit

with same `ruff.toml` as in #831, except 2 last lines which enable ignore of deprecated hooks (https://github.com/antonbabenko/pre-commit-terraform/blob/948a36d52d36270c329028bd27f7a3e999c899ff/ruff.toml)

### DIFF
--- a/src/pre_commit_terraform/terraform_docs_replace.py
+++ b/src/pre_commit_terraform/terraform_docs_replace.py
@@ -76,7 +76,7 @@ def invoke_cli_app(parsed_cli_args: Namespace) -> ReturnCodeType:
             procArgs.append('>')
             procArgs.append(
                 './{dir}/{dest}'.format(
-                    dir=dir, dest=cast_to(bool, parsed_cli_args.dest)
+                    dir=dir, dest=cast_to(bool, parsed_cli_args.dest),
                 ),
             )
             subprocess.check_call(' '.join(procArgs), shell=True)
--- a/src/pre_commit_terraform/terraform_docs_replace.py
+++ b/src/pre_commit_terraform/terraform_docs_replace.py
@@ -74,7 +74,7 @@ def invoke_cli_app(parsed_cli_args: Namespace) -> ReturnCodeType:
             procArgs.extend(
                 (
                     'md',
-                    './{dir}'.format(dir=dir),
+                    f'./{dir}',
                     '>',
                     './{dir}/{dest}'.format(
                         dir=dir,
--- a/src/pre_commit_terraform/terraform_docs_replace.py
+++ b/src/pre_commit_terraform/terraform_docs_replace.py
@@ -53,6 +53,7 @@ def invoke_cli_app(parsed_cli_args: Namespace) -> ReturnCodeType:
         'https://github.com/antonbabenko/pre-commit-terraform/issues/248'
         '#issuecomment-1290829226',
         category=UserWarning,
+        stacklevel=2,
     )
 
     dirs: list[str] = []
--- a/src/pre_commit_terraform/terraform_docs_replace.py
+++ b/src/pre_commit_terraform/terraform_docs_replace.py
@@ -59,7 +59,7 @@ def invoke_cli_app(parsed_cli_args: Namespace) -> ReturnCodeType:
     dirs: list[str] = []
     for filename in cast_to(list[str], parsed_cli_args.filenames):
         if os.path.realpath(filename) not in dirs and (
-            filename.endswith('.tf') or filename.endswith('.tfvars')
+            filename.endswith(('.tf', '.tfvars'))
         ):
             dirs.append(os.path.dirname(filename))
 
--- a/src/pre_commit_terraform/terraform_docs_replace.py
+++ b/src/pre_commit_terraform/terraform_docs_replace.py
@@ -53,7 +53,7 @@ def invoke_cli_app(parsed_cli_args: Namespace) -> ReturnCodeType:
         'https://github.com/antonbabenko/pre-commit-terraform/issues/248'
         '#issuecomment-1290829226',
         category=UserWarning,
-        stacklevel=1,
+        stacklevel=1,  # It's should be 2, but tests are failing w/ values >1. As it's deprecated hook, it's safe to leave it as is w/o fixing it later.
     )
 
     dirs: list[str] = []
--- a/src/pre_commit_terraform/terraform_docs_replace.py
+++ b/src/pre_commit_terraform/terraform_docs_replace.py
@@ -53,7 +53,7 @@ def invoke_cli_app(parsed_cli_args: Namespace) -> ReturnCodeType:
         'https://github.com/antonbabenko/pre-commit-terraform/issues/248'
         '#issuecomment-1290829226',
         category=UserWarning,
-        stacklevel=2,
+        stacklevel=1,
     )
 
     dirs: list[str] = []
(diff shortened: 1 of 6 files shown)


webknjaz (approved):

yermulnik (dismissed):

yermulnik:
I guess @webknjaz suggests (even insists) we make no manual changes to such PRs as their merge commits are meant to go into blame ignore file, right?
Maybe it is worth of mentioning this in PR description so that I don't even think of suggesting to rectify that incorrect casting of string to bool 😄

webknjaz:
@yermulnik no, these changes must not go into blame ignore at all. It's not formatting. The fact that these changes were generated by ruff does not mean that they are any less of linting fixes with runtime impact more often than not. Such patches shouldn't be hidden. Only real mass formatting should be.

webknjaz:
> Maybe it is worth of mentioning this in PR description so that I don't even think of suggesting to rectify that incorrect casting of string to bool 😄

FWIW, once in the f-string, that entire `cast_to()` call might not even be necessary at all. Although, I haven't checked what MyPy reports.

webknjaz:
In general, it's okay to add fixes here that relate directly to fixing the rule violations. Especially, since one 55009e6 broke the CI and mandates an additional change in the tests to make it work.

tests/pytest/_cli_test.py
```diff
@@ -89,7 +89,8 @@ def populate_argument_parser(
             return None
 
         def invoke_cli_app(self, parsed_cli_args: Namespace) -> ReturnCodeType:
-            raise PreCommitTerraformExit('sentinel')
+            msg = 'sentinel'
+            raise PreCommitTerraformExit(msg)
```
webknjaz, line 93: Wait, was this automatic? This seems manual. Drop “autofixes” from the CI title. Just say that you're fixing linting violations. It doesn't really matter if it's semiautomatic or not.

MaxymVlasov: Its automatic
 EM101 - Exception must not use a string literal, assign to variable first
[`93898c9` (#863)](https://github.com/antonbabenko/pre-commit-terraform/pull/863/commits/93898c98853448dd2b83353fd1b8f3589fe77b08)

webknjaz: In this case, we can probably just reuse another sentinel var:
```suggestion
            raise PreCommitTerraformExit(self.CLI_SUBCOMMAND_NAME)
```

But in a more generic case, I'd rename `msg` to something meaningful. Like `program_exit_msg`.

src/pre_commit_terraform/terraform_docs_replace.py
```diff
@@ -53,12 +53,13 @@ def invoke_cli_app(parsed_cli_args: Namespace) -> ReturnCodeType:
         'https://github.com/antonbabenko/pre-commit-terraform/issues/248'
         '#issuecomment-1290829226',
         category=UserWarning,
+        stacklevel=2,
```
webknjaz, line 56:
```suggestion
        stacklevel=1,  # FIXME: figure out how to suppress in tests
```

MaxymVlasov, line 56: That's breaks test. And because it's deprecated hook, no sense to fix it later, so just set it to 1 and forget about it
```suggestion
        stacklevel=1,
```

src/pre_commit_terraform/terraform_docs_replace.py
```diff
@@ -53,12 +53,13 @@ def invoke_cli_app(parsed_cli_args: Namespace) -> ReturnCodeType:
         'https://github.com/antonbabenko/pre-commit-terraform/issues/248'
         '#issuecomment-1290829226',
         category=UserWarning,
+        stacklevel=1,
```
webknjaz, line 56: Document the reasoning
```suggestion
        stacklevel=1,  # FIXME: figure out how to suppress in tests
```

MaxymVlasov: Documented. No need to figure out that.
[`3132f17` (#863)](https://github.com/antonbabenko/pre-commit-terraform/pull/863/commits/3132f1719de112b7f5354eea7001d399a6b5d37f)

webknjaz: It might be useful to know for any future occurrences.

webknjaz (approved):

src/pre_commit_terraform/terraform_docs_replace.py
```diff
@@ -68,14 +69,17 @@ def invoke_cli_app(parsed_cli_args: Namespace) -> ReturnCodeType:
...
-                './{dir}/{dest}'.format(
-                    dir=dir, dest=cast_to(bool, parsed_cli_args.dest)
+            procArgs.extend(
+                (
+                    'md',
+                    f'./{dir}',
+                    '>',
+                    './{dir}/{dest}'.format(
+                        dir=dir,
+                        dest=cast_to('bool', parsed_cli_args.dest),
```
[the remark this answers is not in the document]
yermulnik: Can we have this rectified please? I mean accepting coderabbit's code suggestion?

MaxymVlasov:
> > In general, it's okay to add fixes here that relate directly to fixing the rule violations. Especially, since one [55009e6](https://github.com/antonbabenko/pre-commit-terraform/commit/55009e6c2825aea2210bc239b2b68a2a6209d854) broke the CI and mandates an additional change in the tests to make it work.
> 
> @MaxymVlasov Let's get this fixed then.

Already fixed

src/pre_commit_terraform/terraform_docs_replace.py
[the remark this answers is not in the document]
MaxymVlasov: @webknjaz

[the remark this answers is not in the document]
webknjaz: > Can we have this rectified please? I mean accepting coderabbit's code suggestion?

FWIW, this is not really related to the linting violations that are being fixed in this PR. It might be better to do this separately.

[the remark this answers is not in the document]
yermulnik: > FWIW, this is not really related to the linting violations that are being fixed in this PR. It might be better to do this separately.

Let's do it one way or another. Coderabbit is crying and I also don't think the confusing expression is worth of keeping in code once we've found it.

yermulnik:
> Let's not contribute to unnecessary merge delays.

I'm ok to approve and will do it now, though the delay is because we can't agree on the most greatest way to fix this single bit of obviously wrong code bit 🤷🏻 Get this fixed and, while you both are on it, you catch if any part of the code relies upon this incorrect bit of code.

webknjaz:
> > 
> 
> For me it's still there 😕

It's not the same thing you're quoting.

yermulnik:
> > For me it's still there 😕
> 
> It's not the same thing you're quoting.

I'm not following you. My point is about that odd cast from string to bool. I'm talking about it and only it. So I just can't be quoting something else 🤷🏻 

> > Get this fixed and, while you both are on it, you catch if any part of the code relies upon this incorrect bit of code.
> 
> That cast only influences MyPy runs. It exists to make MyPy interpret things as certain types, as a hack. It doesn't have any effect in runtime whatsoever.

Honestly, I don't really care what that cast is for. It is confusing and I'd like this fixed.
Nothing else. Just a fix while you both are working on this.

yermulnik (approved):

MaxymVlasov:
@yermulnik please create follow-up PR that will fix this cast. Tnx

yermulnik:
> @yermulnik please create follow-up PR that will fix this cast. Tnx

https://github.com/antonbabenko/pre-commit-terraform/pull/864

antonbabenko:

PR #864, opened 2025-03-26
chore: Streamline `cast_to()` into correct type

yermulnik:
Follow-up for #863 to correct `cast_to()` cast type for the string arg from `bool` to `str`.

### DIFF
--- a/src/pre_commit_terraform/terraform_docs_replace.py
+++ b/src/pre_commit_terraform/terraform_docs_replace.py
@@ -78,7 +78,7 @@ def invoke_cli_app(parsed_cli_args: Namespace) -> ReturnCodeType:
                     '>',
                     './{dir}/{dest}'.format(
                         dir=dir,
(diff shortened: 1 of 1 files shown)


src/pre_commit_terraform/terraform_docs_replace.py
```diff
@@ -78,7 +78,7 @@ def invoke_cli_app(parsed_cli_args: Namespace) -> ReturnCodeType:
                     '>',
                     './{dir}/{dest}'.format(
                         dir=dir,
-                        dest=cast_to('bool', parsed_cli_args.dest),
+                        dest=cast_to('str', parsed_cli_args.dest),
```
MaxymVlasov, line 81: They are both not covered by tests, so we can ignore codecov violations, as it's a deprecated hook anyway

image

MaxymVlasov (approved):

PR #872, opened 2025-04-06
fix(WSL): Fix parallelism support for WSL systems with enabled systemd

### LINKED ISSUE #871: Incorrectly detecting wsl when systemd is enable
https://github.com/antonbabenko/pre-commit-terraform/blob/e41252b0d1a25565a38a05e7fd6b02035260f277/hooks/_common.sh#L210

Hello! I'm encountering an issue in detecting wsl and setting the cores. When systemd is enable the file is WSLInterop-late.

yermulnik:

- [ ] This PR fixes a bug.

Add a version of `WSLInterop` file that exists in WSL with SystemD enabled.
Ref: https://github.com/Mozilla-Ocho/llamafile/issues/275

Fixes #871 

Or should we better use something like `grep -qi Microsoft /proc/version` (ref: https://github.com/microsoft/WSL/issues/4071) — @MaxymVlasov What do you think? 🤔

### DIFF
--- a/hooks/_common.sh
+++ b/hooks/_common.sh
@@ -206,8 +206,13 @@ function common::get_cpu_num {
 
   local cpu_quota cpu_period cpu_num
 
+  local wslinterop_path = "/proc/sys/fs/binfmt_misc/WSLInterop"
+
   if [[ -f /sys/fs/cgroup/cpu/cpu.cfs_quota_us &&
-    ! -f /proc/sys/fs/binfmt_misc/WSLInterop ]]; then # WSL have cfs_quota_us, but WSL should be checked as usual Linux host
+    ( # WSL has cfs_quota_us, but WSL should be checked as usual Linux host
+      ! -f "${wslinterop_path}" &&
+      ! -f "${wslinterop_path}-late"
+    ) ]]; then
     # Inside K8s pod or DinD in K8s
     cpu_quota=$(< /sys/fs/cgroup/cpu/cpu.cfs_quota_us)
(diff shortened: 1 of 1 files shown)


hooks/_common.sh
```diff
@@ -206,8 +206,10 @@ function common::get_cpu_num {
 
   local cpu_quota cpu_period cpu_num
 
+  local wslinterop_path="/proc/sys/fs/binfmt_misc/WSLInterop"
```
MaxymVlasov, line 209: Let's make it read-only 


```suggestion
  local -r wslinterop_path="/proc/sys/fs/binfmt_misc/WSLInterop"
```

yermulnik: Makes sense. Thanks.

yermulnik:
> > Or should we better use something like `grep -qi Microsoft /proc/version`
> 
> Sounds good to me too. I not sure what is faster or more relaible solution for our case. And there WSL and WSL2 - they could have different behavior

Here's how the Snapd does it: https://github.com/canonical/snapd/blob/master/release/release.go#L176-L190 — so I'm leaning to let it check whether any of these not exist: `WSLInterop`, `WSLInterop-late`, and additionally `/run/WSL` (I'm adding this third file to the bundle now).

MaxymVlasov (commented):
LGTM, just let's wait confirmation from @rayppapa that it works for him

MaxymVlasov (approved):

raypappa:
To follow up here as well as on the original issue. This works great for me. I added some additional information on the original issue.

antonbabenko:

PR #903, opened 2025-06-05
fix(`infracost_breakdown`): Make hook compatible with bash 3.2 (macOS)

MaxymVlasov:
pull-back from fork

- [x] This PR fixes a bug.

This fixes the error `line 77: -1: substring expression < 0` and replaces mapfile with something compatible for bash 3.x

-    Replace mapfile with while-read loop for bash 3.2 compatibility
-    Fix substring expressions to avoid negative indices: (${check: -1} => ${check:$((${#check}-1)):1})
-    Replace negative array indexing with explicit length calculation
-    Maintain identical functionality while supporting macOS default bash version

### How can we test changes

I only tested my own use case which is to get a failed check if `.totalMonthlyCost` crosses a threshold:

```yaml
    hooks:
      - id: infracost_breakdown
        args:
          - --args=--path=./
          # Fail if cost is higher than USD 4
          - --hook-config='.totalMonthlyCost|tonumber < 4'
          - --hook-config='.currency == "USD"'
```

### DIFF
--- a/hooks/infracost_breakdown.sh
+++ b/hooks/infracost_breakdown.sh
@@ -70,19 +70,24 @@ function infracost_breakdown_ {
     # -h .totalHourlyCost > 0.1
     # --hook-config=.currency == "USD"
     first_char=${check:0:1}
-    last_char=${check: -1}
+    last_char=${check:$((${#check} - 1)):1}
     if [ "$first_char" == "$last_char" ] && {
       [ "$first_char" == '"' ] || [ "$first_char" == "'" ]
     }; then
-      check="${check:1:-1}"
+      check="${check:1:$((${#check} - 2))}"
     fi
 
-    mapfile -t operations < <(echo "$check" | grep -oE '[!<>=]{1,2}')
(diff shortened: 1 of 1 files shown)


yermulnik (approved):

MaxymVlasov:
Double-checked on

```yaml
    - id: infracost_breakdown
      args:
        # - --args=--config-file=test/infracost.yml
        # - --args=--path=./environment/qa
        # - --args=--terraform-plan-flags="-var-file=../../test.tfvars"
        # - --args=--terraform-plan-flags="-var-file=../../test2.tfvars"
        # - --hook-config='.totalHourlyCost >= 0.1'
        # - --hook-config=".totalHourlyCost|tonumber > 1"
        # - -h ".projects[].diff.totalMonthlyCost|tonumber!=10000"
        # - -h [.projects[].diff.totalMonthlyCost | select (.!=null) | tonumber] | add > 1000
      verbose: true # Always show costs
```

using 1 `--hook-config` at once - works fine

antonbabenko:
This PR is included in version 1.99.2 :tada:

PR #905, opened 2025-06-06
chore(checks): Init wemake-python-styleguide configuration

MaxymVlasov:
* Integrate wemake-python-styleguide and ruff
* Add ignores for deprecated hook
* Make initial setup of wemake-python-styleguide, by disabling a few too strict rules


setup.cfg
MaxymVlasov, line 1: List of not fixed in this PR issues:
```
src/pre_commit_terraform/_cli_subcommands.py:7:1: WPS407 Found mutable module constant

src/pre_commit_terraform/_structs.py:12:5: WPS115 Found upper-case constant in a class: OK
src/pre_commit_terraform/_structs.py:13:5: WPS115 Found upper-case constant in a class: ERROR
src/pre_commit_terraform/_types.py:17:5: WPS115 Found upper-case constant in a class: CLI_SUBCOMMAND_NAME
tests/pytest/_cli_test.py:52:9: WPS115 Found upper-case constant in a class: CLI_SUBCOMMAND_NAME
tests/pytest/_cli_test.py:82:9: WPS115 Found upper-case constant in a class: CLI_SUBCOMMAND_NAME

tests/pytest/_cli_test.py:51:5: WPS431 Found nested class: CustomCmdStub
tests/pytest/_cli_test.py:81:5: WPS431 Found nested class: CustomCmdStub

tests/pytest/_cli_test.py:54:9: WPS324 Found inconsistent `return` statement
tests/pytest/_cli_test.py:58:13: WPS324 Found inconsistent `return` statement
tests/pytest/_cli_test.py:84:9: WPS324 Found inconsistent `return` statement
tests/pytest/_cli_test.py:88:13: WPS324 Found inconsistent `return` statement
```
https://results.pre-commit.ci/run/github/69382485/1749246048.bYxekzxPQ4aC8LwApZczqw
https://github.com/antonbabenko/pre-commit-terraform/actions/runs/15500283856/job/43646441309?pr=905


@webknjaz I would be happy to hear what exactly to do with them.

webknjaz: Each rule is explained @ https://wemake-python-styleguide.readthedocs.io/en/latest/pages/usage/violations/index.html

.flake8
```diff
@@ -0,0 +1,108 @@
...
+  WPS428,
+  # WPS462: "Wrong multiline string usage" -- false-positives with
+  # attribute docstrings. Ref:
+  # https://github.com/wemake-services/wemake-python-styleguide/issues/3056
+  WPS462,
+  # WPS300: "Forbid imports relative to the current folder" -- we use relative imports
+  WPS300,
+
+# WPS410: "Forbid some module-level variables" -- __all__ is a legitimate use case
+allowed-module-metadata=__all__
```
webknjaz: Move plugin-specific settings to the end.

.flake8
```diff
@@ -0,0 +1,108 @@
...
+
+# https://wemake-python-styleguide.readthedocs.io/en/latest/pages/usage/formatter.html
+#format = wemake
```
webknjaz: Uncomment this

src/pre_commit_terraform/_types.py
```diff
@@ -14,7 +14,10 @@
 class CLISubcommandModuleProtocol(Protocol):
     """A protocol for the subcommand-implementing module shape."""
 
-    CLI_SUBCOMMAND_NAME: str
+    # WPS115: "Require snake_case for naming class attributes".
+    # It's a valid use case as then it's used as constants:
```
webknjaz, line 19: This is actually because this protocol describes module shapes and not regular classes.

src/pre_commit_terraform/terraform_docs_replace.py
```diff
@@ -58,7 +58,7 @@ def populate_argument_parser(subcommand_parser: ArgumentParser) -> None:
     )
 
 
-def invoke_cli_app(parsed_cli_args: Namespace) -> ReturnCodeType:
+def invoke_cli_app(parsed_cli_args: Namespace) -> ReturnCodeType:  # noqa: WPS231
```
webknjaz, line 63: Why?

src/pre_commit_terraform/terraform_docs_replace.py
```diff
@@ -107,8 +107,8 @@ def invoke_cli_app(parsed_cli_args: Namespace) -> ReturnCodeType:
             subprocess.check_call(' '.join(proc_args), shell=True)  # noqa: S602
         # PERF203 - try-except shouldn't be in a loop, but it's deprecated
         # hook, so leave as is
-        except subprocess.CalledProcessError as e:  # noqa: PERF203
+        except subprocess.CalledProcessError as err:  # noqa: PERF203
             # T201 - Leave print statement as is, as this is deprecated hook
-            print(e)  # noqa: T201
+            print(err)  # noqa: T201,WPS421
```
webknjaz: Why?

MaxymVlasov: https://wemake-python-styleguide.readthedocs.io/en/latest/pages/usage/violations/naming.html#wemake_python_styleguide.violations.naming.TooShortNameViolation

.flake8
MaxymVlasov: I checked this previously - it does not provide you clicakble link to point you to line with error, which is annoying. It's pretty but useless

src/pre_commit_terraform/terraform_docs_replace.py
MaxymVlasov: Added comment

.flake8
webknjaz: It does output this. There's per-line references + a separate summary.

.flake8
```diff
@@ -0,0 +1,108 @@
...
+pytest-parametrize-values-type = tuple
+pytest-parametrize-values-row-type = tuple
+# PT023:
+pytest-mark-no-parentheses = true
+
+# wemake-python-styleguide
+show-source = true
+
+# WPS410: "Forbid some module-level variables" -- __all__ is a legitimate use case
+allowed-module-metadata=__all__
```
webknjaz: Sort within the section + consistent whitespaces 
```suggestion
# WPS410: "Forbid some module-level variables" -- __all__ is a legitimate use case
allowed-module-metadata = __all__

show-source = true 
```

src/pre_commit_terraform/terraform_docs_replace.py
webknjaz: It's best to follow the same variable naming practices as in other places tho.

yermulnik: Yeah, while I agree with "_too short name_" style violation, I lean to support consistency more too 👍🏻

MaxymVlasov: So, are we voting for disabling TooShortNameViolation (WPS111) globally as disruptive?

Btw, there are no other `e` entries in codebase, so it's consistent, AFAIK

.flake8
MaxymVlasov: it's not clickable. You need to manually go to line inside file

[Screencast from 13.06.25 14:24:39.webm](https://github.com/user-attachments/assets/d2305d91-12e5-40ec-bb2d-49b4b422ac96)

src/pre_commit_terraform/terraform_docs_replace.py
yermulnik: > So, are we voting for disabling TooShortNameViolation (WPS111) globally as disruptive?

Disruptive is to violate this rule.

I also can't get why the `WPS421` in code and `WPS111` in your comment above 😕 
I **really** wish you leave comments about suppressed checks for clarity so that others do not need to google-search them. Appreciate in advance =)

> Btw, there are no other `e` entries in codebase, so it's consistent, AFAIK

If there are no other too short names, why would we not rectify this here instead of suppressing the check? 😕

.flake8
yermulnik: I guess the idea is not clickability which effectively is provided by your IDE or terminal rather than by linter output, but about convinient reference to file-and-line-and-row which is useful and handy. I personally prefer this better than default behavior of putting path to file before the list of per-line violations (but it's my personal pref though).

A side comment from the linked docu page: the show-violation-links option looks cool =)

yermulnik (approved):

.flake8
webknjaz: Yes, that's my impression too. One can change the format when calling flake8 from an editor that wants it to look certain way but the default should optimize for better generic experience, not for an arbitrary editor.

src/pre_commit_terraform/terraform_docs_replace.py
webknjaz: I still maintain that `err` is almost just as bad as `e` or `z`. This is usually an indication that whoever wrote the code in the first place did not really understand what's inside either.

.flake8
MaxymVlasov:
```suggestion
format = wemake
```

src/pre_commit_terraform/terraform_docs_replace.py
MaxymVlasov: >I also can't get why the WPS421 in code and WPS111 in your comment above 

That's different violations. Here description in place where it will be actually fixed

https://github.com/antonbabenko/pre-commit-terraform/blob/6687da970dcead31d6fa36931384b821708db27a/src/pre_commit_terraform/_cli.py#L39-L40
https://wemake-python-styleguide.readthedocs.io/en/latest/pages/usage/violations/best_practices.html#wemake_python_styleguide.violations.best_practices.WrongFunctionCallViolation

> I still maintain that err is almost just as bad as e or z. This is usually an indication that whoever wrote the code in the first place did not really understand what's inside either.

Agree. But this one is deprecated hook, so we're choosing here between `err`, `error` and `e # noqa: WPS111`, as I will not spend time trying to fix hook which exist only because I follow SemVer, and we don't have much stuff for v2.0 yet, otherwise this hook would have been removed ~4 years ago

yermulnik: > > I also can't get why the WPS421 in code and WPS111 in your comment above
> 
> That's different violations. Here description in place where it will be actually fixed

And this is exactly what's confusing: the thread is about `WPS421` while you're mentioning `WPS111`:
image

> But this one is deprecated hook

I personally am fine with `print()` and non-descriptive var names in deprecated hook if check suppression is accompanied with a comment similarly as it is done for `T201`:
image
Thanks

yermulnik: _(I'm approving the PR to avoid blocking development and since technically PR is good I guess)_

yermulnik (dismissed):

src/pre_commit_terraform/terraform_docs_replace.py
MaxymVlasov: Thread about WPS111, about renaming e to err

yermulnik: Might be.

webknjaz: It's probably fine to ignore more things in the deprecated code. But "fixing" incorrectly is something that I'm uneasy about.

antonbabenko:
This PR is included in version 1.99.5 :tada:

PR #911, opened 2025-07-08
feat(`terraform_fmt`): Add support for `.tftest.hcl` and `.tfmock.hcl` formatting

nikolaik:
This adds `.tftest.hcl` and `.tfmock.hcl` to the default `files` attribute for the `terraform_fmt` hook config.

Support for formatting these files were added in recent terraform versions

---
Notes by @MaxymVlasov:
* `terraform fmt` definitely supports `.tfmock.hcl` , because not supporting of it was reported as a bug, and now is resolved by addition of such support - https://github.com/hashicorp/terraform/issues/34577 
* And `.tftest.hcl` also mentioned in `fmtSupportedExts` - https://github.com/hashicorp/terraform/blob/0c63fb2b097edcd5cb1a91322765a414206fbea2/internal/command/fmt.go#L30-L35

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -13,7 +13,7 @@
     Rewrites all Terraform configuration files to a canonical format.
   entry: hooks/terraform_fmt.sh
   language: script
-  files: \.(tf|tofu|tfvars)$
+  files: \.(tf|tofu|tfvars|tftest\.hcl|tfmock\.hcl)$
   exclude: \.terraform/.*$
 
 - id: terraform_docs
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -13,6 +13,8 @@
     Rewrites all Terraform configuration files to a canonical format.
   entry: hooks/terraform_fmt.sh
   language: script
+  # Supported extensions by Terraform specified in
+  # https://github.com/hashicorp/terraform/blob/0c63fb2b097edcd5cb1a91322765a414206fbea2/internal/command/fmt.go#L30-L35
   files: \.(tf|tofu|tfvars|tftest\.hcl|tfmock\.hcl)$
   exclude: \.terraform/.*$
 


.pre-commit-hooks.yaml
```diff
@@ -13,7 +13,7 @@
     Rewrites all Terraform configuration files to a canonical format.
   entry: hooks/terraform_fmt.sh
   language: script
-  files: \.(tf|tofu|tfvars)$
+  files: \.(tf|tofu|tfvars|tftest\.hcl|tfmock\.hcl)$
```
yermulnik, line 18: Optimize regex (may worsen readability though):
```suggestion
  files: \.(tf(vars|(mock|test)\.hcl)?|tofu)$
```

yermulnik (dismissed):
LGTM, though I'd like @MaxymVlasov's review/approve too.

UPD: also maybe it is worth to implement the same for `terraform_validate` and exclude from `terragrunt_fmt` and `terragrunt_validate` hooks? 🤔

.pre-commit-hooks.yaml
MaxymVlasov: Let's leave as is for readability

MaxymVlasov, line 18: Did you test that it works for these formats?

We had recently PR for .tfbackend - https://github.com/antonbabenko/pre-commit-terraform/pull/822, but it actually not supported by `t fmt`

Feel free to create a GH repo just for testing this new functional - I'll fork it to org and one day will include it to test suite.

And I will much apreciate if you'll include related links about support these extensions by TF and support of HCL2 formatter for them in this PR

MaxymVlasov, line 18:
```suggestion
  # Supported extensions by Terraform specified in
  # https://github.com/hashicorp/terraform/blob/0c63fb2b097edcd5cb1a91322765a414206fbea2/internal/command/fmt.go#L30-L35
  files: \.(tf|tofu|tfvars|tftest\.hcl|tfmock\.hcl)$
```

yermulnik (approved):

antonbabenko:
This PR is included in version 1.100.0 :tada:

nikolaik:
Thanks for checking actual support and adding references! 👏

PR #930, opened 2025-09-01
ci: Add zizimor as workflow to get GH-native alerts

MaxymVlasov:

Another bit of security improvements.

Workflow body: https://github.com/zizmorcore/workflow/blob/main/.github/workflows/reusable-zizmor.yml

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -506,6 +506,13 @@ jobs:
     secrets:
       codecov-token: ${{ secrets.CODECOV_TOKEN }}
 
+  lint-github-actions:
+    name: Lint GitHub Actions
+    permissions:
+      security-events: write
+    # yamllint disable-line rule:line-length
+    uses: zizmorcore/workflow/.github/workflows/reusable-zizmor.yml@3bb5e95068d0f44b6d2f3f7e91379bed1d2f96a8
+
   check:  # This job does nothing and is only used for the branch protection
 
     # Separate 'pull_request' check from other checks to avoid confusion in


webknjaz (approved):

.github/workflows/ci-cd.yml
```diff
@@ -506,6 +506,13 @@ jobs:
     secrets:
       codecov-token: ${{ secrets.CODECOV_TOKEN }}
 
+  lint-github-actions:
+    name: Lint GitHub Actions
+    permissions:
+      security-events: write
+    # yamllint disable-line rule:line-length
+    uses: zizmorcore/workflow/.github/workflows/reusable-zizmor.yml@3bb5e95068d0f44b6d2f3f7e91379bed1d2f96a8
```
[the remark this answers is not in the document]
webknjaz: No, you're drunk.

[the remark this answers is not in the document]
webknjaz: (that privilege would only be needed for private repos)

.github/workflows/ci-cd.yml
```diff
@@ -506,6 +506,13 @@ jobs:
     secrets:
       codecov-token: ${{ secrets.CODECOV_TOKEN }}
 
+  lint-github-actions:
+    name: Lint GitHub Actions
+    permissions:
+      security-events: write
+    # yamllint disable-line rule:line-length
```
yermulnik, line 513: JFYI: https://yamllint.readthedocs.io/en/latest/rules.html#module-yamllint.rules.line_length — `allow-non-breakable-words: true`

MaxymVlasov: Yeah, but it's `true` by default, and it wasn't disabled in https://github.com/antonbabenko/pre-commit-terraform/blob/master/.yamllint

I suppose it dumb enough to see space in 
```yaml
uses: zizmorcore/workflow/.github/workflows/reusable-zizmor.yml@3bb5e95068d0f44b6d2f3f7e91379bed1d2f96a8
```
as something that can be replaced to new line like

```yaml
uses:
  zizmorcore/workflow/.github/workflows/reusable-zizmor.yml@3bb5e95068d0f44b6d2f3f7e91379bed1d2f96a8
```

even if it's not make much sense

yermulnik: > Yeah, but it's `true` by default

What I've seen multiple times is that when you craft your own yamllint config it does not include default settings, so if you customize yamllint configuration, you should provide at the very minimum all those from https://github.com/adrienverge/yamllint/blob/master/yamllint/conf/default.yaml to ensure the default values are in place.

For me an explicit `allow-non-breakable-words: true` works for lines line `yaml_key: a,very,log,line,without,whitespaces`.

MaxymVlasov: That's... unfortunate

MaxymVlasov: It doesn't work for me
image

yermulnik: That's odd 😕 Is it indeed picking that yamllint config? 🤔

MaxymVlasov: yes, otherwise it will say that line longer than 80 symbols, not 100

yermulnik: Hmm, I double checked and it indeed doesn't work for me locally anymore 😲 
Apologies for the noise 😕

antonbabenko:
This PR is included in version 1.100.1 :tada:

PR #937, opened 2025-10-06
fix(`terraform_wrapper_module_for_each`): Prevent locale-dependent sorting inconsistencies

zahorniak:

- [x] This PR fixes a bug.

I discovered that the hook `terraform_wrapper_module_for_each` produces different outputs on different machines due to locale-dependent sorting behavior and differences between `glibc` versions. And especially for symbols like `_` that can be de-prioritized depending on the locale of the `glibc` version.
I did some digging and found out that the `sort` command uses values from `LC_*` variables to determine the character ordering.

It appears that adding `LC_ALL=C` will prevent any inconsistencies in the future for all users. 

Here is an example of when my local pre-commit run moved `task_tags` to the bottom of the list and pre-commit failed on Github Runner because it tried to put it back: https://github.com/terraform-aws-modules/terraform-aws-ecs/actions/runs/18283659979/job/52056232720?pr=361

Here is more info about my current machine:
```sh
$ getconf GNU_LIBC_VERSION
glibc 2.42

$ sort --version
sort (GNU coreutils) 9.8

$ hcledit version
0.2.17

$ locale
LANG=en_US.UTF-8
LC_CTYPE="en_US.UTF-8"
LC_NUMERIC="en_US.UTF-8"
LC_TIME="en_US.UTF-8"
LC_COLLATE="en_US.UTF-8"
LC_MONETARY="en_US.UTF-8"
LC_MESSAGES="en_US.UTF-8"
LC_PAPER="en_US.UTF-8"
LC_NAME="en_US.UTF-8"
LC_ADDRESS="en_US.UTF-8"
LC_TELEPHONE="en_US.UTF-8"
LC_MEASUREMENT="en_US.UTF-8"
LC_IDENTIFICATION="en_US.UTF-8"
LC_ALL=
```

### How can we test changes

I personally tested my changes on the https://github.com/terraform-aws-modules/terraform-aws-ecs repo. Currently, it meets all the requirements for this edge case. 

`LC_ALL=C pre-commit run -a` and `LC_ALL=en_US.UTF-8 pre-commit run -a` produces different outputs for me in this repo. Where the `LC_ALL=C pre-commit run -a` matches the current state of the repo.

### DIFF
--- a/hooks/terraform_wrapper_module_for_each.sh
+++ b/hooks/terraform_wrapper_module_for_each.sh
@@ -321,7 +321,7 @@ EOF
 
     # Get names of module variables in all terraform files
     # shellcheck disable=SC2207
-    module_vars=($(echo "$all_tf_content" | hcledit block list | { grep "^variable\." | cut -d'.' -f 2 | sort || true; }))
+    module_vars=($(echo "$all_tf_content" | hcledit block list | { grep "^variable\." | cut -d'.' -f 2 | LC_ALL=C sort || true; }))
 
     # Get names of module outputs in all terraform files
     # shellcheck disable=SC2207


yermulnik:
@zahorniak Thank you for the contribution.
The touched topic is a bit fragile and I personally am hesitant on enforcing some specific locale for all users.
I can see that GNU `sort` from `coreutils` relies upon `LC_ALL`, whereas BSD `sort` takes into account multiple `LC_*` vars, which affect not just sorting order, but also, for instance, the language of output messages and numeric/time format (with `LC_ALL` overriding all of the `LC_*` vars).
While I see rationale behind your suggestion, I also see how this might be a bit too invasive to override user's preference for locale setting by following the suggested approach.

It is also not entirely correct to add this as an ad-hoc fix in a specific place for a specific utility, rather than globally e.g. by means of [/hooks/_common.sh](https://github.com/antonbabenko/pre-commit-terraform/blob/master/hooks/_common.sh).

However, I'd suggest you to look into:
* Passing `LC_ALL` env var on a per-hook basis: https://github.com/antonbabenko/pre-commit-terraform#all-hooks-set-env-vars-inside-hook-at-runtime
* Exporting `LC_ALL` env var outside of `pre-commit` framework **globally** for a system, or for a user (or a team), or even for a GHA workflow.
* Employing any of env management tools, like [direnv](https://github.com/direnv/direnv), [asdf](https://github.com/asdf-vm/asdf), [mise](https://github.com/jdx/mise) or similar to define `LC_ALL` per project or per directory.

Let me know if you think this PR should better be kept open. Thank you once again for looking into improving this project ❤️

PR #947, opened 2025-11-16
feat(docker): Allow authenticated calls to GitHub API during `docker build`

wI2L:
Accept build arg `GITHUB_TOKEN` to authenticate calls made to GitHub API in `common::install_from_gh_release` function.

Closes #946

Put an `x` into the box if that apply:

- [ ] This PR introduces breaking change.
- [ ] This PR fixes a bug.
- [x] This PR adds new functionality.
- [ ] This PR enhances existing functionality.

### Description of your changes

Accept build arg `GITHUB_TOKEN` to authenticate calls made to GitHub API in `common::install_from_gh_release` function.

### How can we test changes

Generate a GitHub PAT and export it in your env.
Run `docker build -t pre-commit-terraform --build-arg GITHUB_TOKEN --build-arg INSTALL_ALL=true .` to build the `pre-commit`, using the PAT during build time to authenticate calls.

PR #948, opened 2025-11-18
ci: Drop EOL Python 3.9 and macos-13

MaxymVlasov:
Deal with issues highlighted in https://github.com/antonbabenko/pre-commit-terraform/pull/947

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -1,6 +1,3 @@
-# Assume Python 3.10
-target-version = "py310"
-
 line-length = 79 # To decrease PR diff size
 
 namespace-packages = ["src/pre_commit_terraform/", "tests/pytest/"]


MaxymVlasov:
>==> Could not verify signature. Please contact Codecov if problem continues
    Exiting...
    
Must be a Cloudflare issue. Let's wait a day

.github/workflows/ci-cd.yml
```diff
@@ -453,17 +453,17 @@ jobs:
...
+        - 3.14
         # str
         - >-
           3.10
+        - 3.13
         - 3.12
         - 3.11
         runner-vm-os:
         - ubuntu-24.04
         - macos-14
```
webknjaz: Also bump this to 15 as GH is deprecating older VMs.

ruff.toml
```diff
@@ -1,5 +1,5 @@
-# Assume Python 3.9
-target-version = "py39"
+# Assume Python 3.10
+target-version = "py310"
```
webknjaz: Isn't this inherited from `pyproject.toml`'s `[project].requires-python`?

MaxymVlasov: No idea

.github/workflows/ci-cd.yml
MaxymVlasov: Actually, no

image

MaxymVlasov: but I can bump it now anyway

MaxymVlasov:
```suggestion
        - macos-15
```

.pre-commit-config.yaml
```diff
@@ -121,53 +121,53 @@ repos:
...
+    - --txt-report=.tox/.tmp/.test-results/mypy--py-3.12
     pass_filenames: false
   - id: mypy
-    alias: mypy-py39
-    name: MyPy, for Python 3.9
+    alias: mypy-py310
+    name: MyPy, for Python 3.10
     additional_dependencies:
     - lxml  # dep of `--txt-report`, `--cobertura-xml-report` & `--html-report`
     - pytest
```
webknjaz, line 162: It might be a good idea to start pinning the additional deps in here. Both these and the repo version contribute to the cache key in pre-commit. And the cache in https://pre-commit.ci is effectively immportal. So when a new version of a dep like `pytest` is out, `pre-commit.ci` won't start using it until the cache key changes. That can happen when the repo version above changes or when something in this list of strings changes. This results in updating the repo version suddenly starting to use a different `pytest` version way later than its actual release happens. Having pins in here gives you better control over *when* the new versions are being picked up and so you'd see failures when you bump things in here instead of seeing failures at some point later, when the repo is getting bumped by the bot.

One project that can manage these is https://pypi.org/p/upadup.

ruff.toml
webknjaz: https://docs.astral.sh/ruff/configuration/#inferring-the-python-version

ruff.toml
```diff
@@ -1,5 +1,5 @@
...
```
webknjaz:
```suggestion
```

webknjaz:
> Must be a Cloudflare issue. Let's wait a day

Yep, that's happening to all my CI runs. All the hosts behind their CDN return HTTP 500: https://cloudflarestatus.com.

antonbabenko:
This PR is included in version 1.104.0 :tada:

PR #967, opened 2026-05-25
feat(`terraform_wrapper_module_for_each`): Strip `provider_meta` blocks from `versions.tf`

### LINKED ISSUE #954: Review the wrapper hook to make sure it works correctly with provider_meta
If anyone wants to take a look...

> FYI - @antonbabenko might be worthwhile to update the wrapper hook to not copy over the provider meta because technically all of the wrappers are duplicating the same user agent tag twice.

_Originally posted by @bryantbiggs in https://github.com/terraform-aws-modules/terraform-aws-iam/pull/632#discussion_r2673107849_

pen-pal:
- [x] This PR fixes a bug.

The hook copies the source module's `versions.tf` into the wrapper verbatim. If the source declares a `provider_meta` block (used by terraform-aws-modules to tag the provider's user-agent), the wrapper inherits it. The wrapper makes no provider calls itself, so that propagation is redundant and skews module-usage telemetry.

Uses `hcledit` (already a hook dependency, see `check_dependencies` at `hooks/terraform_wrapper_module_for_each.sh:430`) to remove `user_agent` and `module_name` for the provider configured via `--module-repo-provider`. The edit is structural:

- Modules with no `provider_meta` are untouched.
- Non-AWS `provider_meta` blocks are untouched.
- A future nested block inside `provider_meta` (e.g. if the AWS provider extends the schema) is preserved.

**Cosmetic caveat**: the (possibly now-empty) `provider_meta` block itself remains, because `hcledit` cannot remove nested blocks. Functionally harmless — no attributes ⇒ no user-agent contribution.

Fixes #954

### How can we test changes

Run the hook against a module whose `versions.tf` has a `provider_meta` block (e.g. any `terraform-aws-modules` repo):

```sh
bash hooks/terraform_wrapper_module_for_each.sh \
  --args=--root-dir=<path> --args=--module-dir=. \
  --args=--module-repo-shortname=<name> <path>/main.tf

Check <path>/wrappers/versions.tf — provider_meta should be gone, everything else intact.
```

Tested against hypothetical value: 

# input
```
terraform {
  provider_meta "aws" {
    user_agent = ["github.com/terraform-aws-modules/terraform-aws-s3-bucket"]

    hypothetical_nested_block {
      key = "value"
    }
  }
}
```
# after the hook
```
terraform {
  provider_meta "aws" {

    hypothetical_nested_block {
      key = "value"
    }
  }
}
```

user_agent cleanly removed, nested block intact.

### DIFF
--- a/hooks/terraform_wrapper_module_for_each.sh
+++ b/hooks/terraform_wrapper_module_for_each.sh
@@ -398,10 +398,10 @@ EOF
       if [[ -f "${full_module_dir}/versions.tf" ]]; then
         cp "${full_module_dir}/versions.tf" "${output_dir}/versions.tf"
         # Don't propagate redundant `provider_meta` user-agent attrs (#954).
-        hcledit attribute rm "terraform.provider_meta.${module_repo_provider}.user_agent" \
-          -f "${output_dir}/versions.tf" -u
-        hcledit attribute rm "terraform.provider_meta.${module_repo_provider}.module_name" \
-          -f "${output_dir}/versions.tf" -u
+        # sepcific to aws cloud only
+        hcledit attribute rm "terraform.provider_meta.${module_repo_provider}.user_agent" -f "${output_dir}/versions.tf" -u
+        # specific to gcp cloud only
+        hcledit attribute rm "terraform.provider_meta.${module_repo_provider}.module_name" -f "${output_dir}/versions.tf" -u
       else
         echo "$CONTENT_VERSIONS_TF" > "${output_dir}/versions.tf"
       fi
--- a/hooks/terraform_wrapper_module_for_each.sh
+++ b/hooks/terraform_wrapper_module_for_each.sh
@@ -398,9 +398,9 @@ EOF
       if [[ -f "${full_module_dir}/versions.tf" ]]; then
         cp "${full_module_dir}/versions.tf" "${output_dir}/versions.tf"
         # Don't propagate redundant `provider_meta` attributes
-        # AWS specific
+        # AWS-provider specific
         hcledit attribute rm "terraform.provider_meta.${module_repo_provider}.user_agent" -f "${output_dir}/versions.tf" -u
-        # GCP specific
+        # GCP-provider specific
         hcledit attribute rm "terraform.provider_meta.${module_repo_provider}.module_name" -f "${output_dir}/versions.tf" -u
       else
         echo "$CONTENT_VERSIONS_TF" > "${output_dir}/versions.tf"
--- a/hooks/terraform_wrapper_module_for_each.sh
+++ b/hooks/terraform_wrapper_module_for_each.sh
@@ -395,14 +395,13 @@ EOF
       echo "$CONTENT_VARIABLES_TF" > "${output_dir}/variables.tf"
 
       # If the root module has a versions.tf, use that; otherwise, create it.
-      # Strip `provider_meta` blocks so wrappers don't duplicate the module's
-      # user-agent tag on every provider call (see #954).
       if [[ -f "${full_module_dir}/versions.tf" ]]; then
-        awk '
-          /^[[:space:]]*provider_meta[[:space:]]+"[^"]+"[[:space:]]*\{[[:space:]]*$/ { in_block = 1; next }
-          in_block && /^[[:space:]]*\}[[:space:]]*$/ { in_block = 0; next }
-          !in_block { print }
-        ' "${full_module_dir}/versions.tf" > "${output_dir}/versions.tf"
+        cp "${full_module_dir}/versions.tf" "${output_dir}/versions.tf"
+        # Don't propagate redundant `provider_meta` user-agent attrs (#954).
+        hcledit attribute rm "terraform.provider_meta.${module_repo_provider}.user_agent" \
+          -f "${output_dir}/versions.tf" -u
+        hcledit attribute rm "terraform.provider_meta.${module_repo_provider}.module_name" \
+          -f "${output_dir}/versions.tf" -u
       else
         echo "$CONTENT_VERSIONS_TF" > "${output_dir}/versions.tf"
       fi
--- a/hooks/terraform_wrapper_module_for_each.sh
+++ b/hooks/terraform_wrapper_module_for_each.sh
@@ -394,9 +394,15 @@ EOF
 
       echo "$CONTENT_VARIABLES_TF" > "${output_dir}/variables.tf"
 
-      # If the root module has a versions.tf, use that; otherwise, create it
+      # If the root module has a versions.tf, use that; otherwise, create it.
+      # Strip `provider_meta` blocks so wrappers don't duplicate the module's
+      # user-agent tag on every provider call (see #954).
       if [[ -f "${full_module_dir}/versions.tf" ]]; then
-        cp "${full_module_dir}/versions.tf" "${output_dir}/versions.tf"
+        awk '
+          /^[[:space:]]*provider_meta[[:space:]]+"[^"]+"[[:space:]]*\{[[:space:]]*$/ { in_block = 1; next }
+          in_block && /^[[:space:]]*\}[[:space:]]*$/ { in_block = 0; next }
+          !in_block { print }
+        ' "${full_module_dir}/versions.tf" > "${output_dir}/versions.tf"
       else
         echo "$CONTENT_VERSIONS_TF" > "${output_dir}/versions.tf"
       fi
--- a/hooks/terraform_wrapper_module_for_each.sh
+++ b/hooks/terraform_wrapper_module_for_each.sh
@@ -397,10 +397,10 @@ EOF
       # If the root module has a versions.tf, use that; otherwise, create it.
       if [[ -f "${full_module_dir}/versions.tf" ]]; then
         cp "${full_module_dir}/versions.tf" "${output_dir}/versions.tf"
-        # Don't propagate redundant `provider_meta` user-agent attrs (#954).
-        # sepcific to aws cloud only
+        # Don't propagate redundant `provider_meta` attributes
+        # AWS specific
         hcledit attribute rm "terraform.provider_meta.${module_repo_provider}.user_agent" -f "${output_dir}/versions.tf" -u
-        # specific to gcp cloud only
+        # GCP specific
         hcledit attribute rm "terraform.provider_meta.${module_repo_provider}.module_name" -f "${output_dir}/versions.tf" -u
       else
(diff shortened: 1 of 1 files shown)


yermulnik:
@pen-pal Thanks for the contribution.

> so the provider's user-agent tag gets sent twice on every call.

What's wrong with such duplication? Does it impact/break something? What's the drawback of leaving `provider_meta "aws" {` block as is in ` terraform {}` inside wrapped configuration?

Also given that `awk` is going to be yet another dependency for the wrapper, it's probably best to convert it to `sed`, which is already in use by the wrapper. WDYT?

Apart from that, I guess it's best to strip the `provider_meta` off of the exact provider, rather than from any `provider_meta` block which might potentially exist and be used with non-AWS providers.

Also (and it's probably my main concern) if by any reason the AWS provider's `provider_meta` nested block receives an update that introduces its own nested block, then the proposed solution would produce incorrect `terraform {}` configuration, which would result in non-obvious syntax error and might complicate usage and diagnostics for end users.
The below is what I mean with the hypothetical configuration:
```bash
> cat ~/tmp/versions.tf
terraform {
  provider_meta "aws" {
    user_agent = [
      "github.com/terraform-aws-modules/terraform-aws-s3-bucket"
    ]

    hypthetical_nested_block {
      key = "value"
    }
  }
}

> cat ~/tmp/versions.tf | awk '/^[[:space:]]*provider_meta[[:space:]]+"[^"]+"[[:space:]]*\{[[:space:]]*$/ { in_block = 1; next }; in_block && /^[[:space:]]*\}[[:space:]]*$/ { in_block = 0; next }; !in_block { print }'
terraform {
  }
}
```

To avoid such — though still hypothetical — confusion, we're better off scaffolding some other solution I reckon.
Off the top of my head, I see three immediate options:
* Update the suggested code to become more intelligent to catch such conditions. This option is weird obviously, as it would introduce unnecessary complexity.
* Since `hcledit` cannot be used to strip the `provider_meta` block directly, we can use it to drop the `user_agent` from this block (`hcledit attribute rm terraform.provider_meta.aws.user_agent`), which (the `user_agent` key) seemingly is the direct "culprit" of the "issue" I guess — would this do the thing to eliminate duplication and would it not produce incorrect configuration? Could you please see if you've got a chance to test and validate this? 🤔
* Use e.g. https://github.com/tmccombs/hcl2json and `jq` to convert HCL to JSON, strip the `provider_meta "aws"` block, and convert back from JSON to HCL — this also would intro unnecessary complexity by means of additional dependencies.

pen-pal:
> @pen-pal Thanks for the contribution.
> 
> > so the provider's user-agent tag gets sent twice on every call.
> 
> What's wrong with such duplication? Does it impact/break something? What's the drawback of leaving `provider_meta "aws" {` block as is in ` terraform {}` inside wrapped configuration?
> 
> Also given that `awk` is going to be yet another dependency for the wrapper, it's probably best to convert it to `sed`, which is already in use by the wrapper. WDYT?
> 
> Apart from that, I guess it's best to strip the `provider_meta` off of the exact provider, rather than from any `provider_meta` block which might potentially exist and be used with non-AWS providers.
> 
> Also (and it's probably my main concern) if by any reason the AWS provider's `provider_meta` nested block receives an update that introduces its own nested block, then the proposed solution would produce incorrect `terraform {}` configuration, which would result in non-obvious syntax error and might complicate usage and diagnostics for end users. The below is what I mean with the hypothetical configuration:
> 
> ```shell
> > cat ~/tmp/versions.tf
> terraform {
>   provider_meta "aws" {
>     user_agent = [
>       "github.com/terraform-aws-modules/terraform-aws-s3-bucket"
>     ]
> 
>     hypthetical_nested_block {
>       key = "value"
>     }
>   }
> }
> 
> > cat ~/tmp/versions.tf | awk '/^[[:space:]]*provider_meta[[:space:]]+"[^"]+"[[:space:]]*\{[[:space:]]*$/ { in_block = 1; next }; in_block && /^[[:space:]]*\}[[:space:]]*$/ { in_block = 0; next }; !in_block { print }'
> terraform {
>   }
> }
> ```
> 
> To avoid such — though still hypothetical — confusion, we're better off scaffolding some other solution I reckon. Off the top of my head, I see three immediate options:
> 
> * Update the suggested code to become more intelligent to catch such conditions. This option is weird obviously, as it would introduce unnecessary complexity.
> * Since `hcledit` cannot be used to strip the `provider_meta` block directly, we can use it to drop the `user_agent` from this block (`hcledit attribute rm terraform.provider_meta.aws.user_agent`), which (the `user_agent` key) seemingly is the direct "culprit" of the "issue" I guess — would this do the thing to eliminate duplication and would it not produce incorrect configuration? Could you please see if you've got a chance to test and validate this? 🤔
> * Use e.g. https://github.com/tmccombs/hcl2json and `jq` to convert HCL to JSON, strip the `provider_meta "aws"` block, and convert back from JSON to HCL — this also would intro unnecessary complexity by means of additional dependencies.

@yermulnik made some updates based on suggestion.
also updated the description

hooks/terraform_wrapper_module_for_each.sh
```diff
@@ -394,9 +394,14 @@ EOF
...
       echo "$CONTENT_VARIABLES_TF" > "${output_dir}/variables.tf"
 
-      # If the root module has a versions.tf, use that; otherwise, create it
+      # If the root module has a versions.tf, use that; otherwise, create it.
       if [[ -f "${full_module_dir}/versions.tf" ]]; then
         cp "${full_module_dir}/versions.tf" "${output_dir}/versions.tf"
+        # Don't propagate redundant `provider_meta` user-agent attrs (#954).
+        hcledit attribute rm "terraform.provider_meta.${module_repo_provider}.user_agent" \
+          -f "${output_dir}/versions.tf" -u
+        hcledit attribute rm "terraform.provider_meta.${module_repo_provider}.module_name" \
```
yermulnik: What `module_name` is? I can't locate it at the https://registry.terraform.io/providers/hashicorp/aws/latest/docs
Could you please provide some details about it? Thanks.

yermulnik:
> @yermulnik made some updates based on suggestion. also updated the description

Thank you.
LGTM, though I left a comment — please take a look.

hooks/terraform_wrapper_module_for_each.sh
yermulnik: @pen-pal A friendly reminder 🔔

pen-pal: > What `module_name` is? I can't locate it at the https://registry.terraform.io/providers/hashicorp/aws/latest/docs Could you please provide some details about it? Thanks.

`module_name` is the **Google** provider's `provider_meta` field — not an AWS one. AWS's equivalent is `user_agent`.

Evidence for `module_name`:

- **Schema** — [`terraform-provider-google` →
`google/provider/provider.go`](https://github.com/hashicorp/terraform-provider-google/blob/main/google/provider/provider.go):
  ```go
  ProviderMetaSchema: map[string]*schema.Schema{
      "module_name": {
          Type:     schema.TypeString,
          Optional: true,
      },
  },

yermulnik: Thanks for update 👍🏻

yermulnik (dismissed):
LGTM
@MaxymVlasov Could you please take a look too? Thanks

hooks/terraform_wrapper_module_for_each.sh
yermulnik: I guess next iteration would be to add attribute name as a comma-separated argument for the hook so that the list can be expanded or overridden on-demand.

antonbabenko: Lets add surrounding comments to which providers each `hcledit` command is related. LGTM otherwise

antonbabenko (commented):
Minor comment, LGTM

yermulnik (dismissed):

hooks/terraform_wrapper_module_for_each.sh
yermulnik: @antonbabenko Please see if 9ef469ab74afb2d2bc02b3bad83748393f80ecc3 looks good to you and approve if it does. Many thanks.

@MaxymVlasov I'd prefer your review/approve as well =)

hooks/terraform_wrapper_module_for_each.sh
```diff
@@ -394,9 +394,14 @@ EOF
...
       echo "$CONTENT_VARIABLES_TF" > "${output_dir}/variables.tf"
 
-      # If the root module has a versions.tf, use that; otherwise, create it
+      # If the root module has a versions.tf, use that; otherwise, create it.
       if [[ -f "${full_module_dir}/versions.tf" ]]; then
         cp "${full_module_dir}/versions.tf" "${output_dir}/versions.tf"
+        # Don't propagate redundant `provider_meta` user-agent attrs (#954).
+        # sepcific to aws cloud only
+        hcledit attribute rm "terraform.provider_meta.${module_repo_provider}.user_agent" -f "${output_dir}/versions.tf" -u
+        # specific to gcp cloud only
```
MaxymVlasov:
```suggestion
        # Don't propagate redundant `provider_meta` attributes
        # AWS specific
        hcledit attribute rm "terraform.provider_meta.${module_repo_provider}.user_agent" -f "${output_dir}/versions.tf" -u
        # GCP specific
```

hooks/terraform_wrapper_module_for_each.sh
```diff
@@ -394,9 +394,14 @@ EOF
...
       echo "$CONTENT_VARIABLES_TF" > "${output_dir}/variables.tf"
 
-      # If the root module has a versions.tf, use that; otherwise, create it
+      # If the root module has a versions.tf, use that; otherwise, create it.
       if [[ -f "${full_module_dir}/versions.tf" ]]; then
         cp "${full_module_dir}/versions.tf" "${output_dir}/versions.tf"
+        # Don't propagate redundant `provider_meta` attributes
+        # AWS specific
+        hcledit attribute rm "terraform.provider_meta.${module_repo_provider}.user_agent" -f "${output_dir}/versions.tf" -u
+        # GCP specific
```
MaxymVlasov:
```suggestion
        # AWS-provider specific
        hcledit attribute rm "terraform.provider_meta.${module_repo_provider}.user_agent" -f "${output_dir}/versions.tf" -u
        # GCP-provider specific
```

MaxymVlasov (approved):

antonbabenko:
This PR is included in version 1.106.0 :tada:

PR #993, opened 2026-06-10
feat(`terraform_validate`): Add support for running hook in `git worktree`

### LINKED ISSUE #992: terraform_validate corrupts worktree index when tofu init clones modules (GIT_INDEX_FILE leakage)
### Describe the bug

When running `git commit` from a linked git worktree (`git worktree add ...`) on a stack that references modules, the commit fails at tree-build time with:

```
error: invalid object 100644 <oid> for '<some-upstream-module-file>'
error: Error building trees
```

even though every pre-commit hook reports `Passed`.

**Root cause**: `terraform_validate` invokes `tofu init -backend=false`, which shells out to `git clone <module-source>` for each `module { source = ... }` reference. The child `git clone` inherits `GIT_INDEX_FILE` / `GIT_DIR` / `GIT_WORK_TREE` / `GIT_OBJECT_DIRECTORY` from the parent `git commit` env (which `git` sets when invoking hooks). When the parent is in a worktree, `GIT_INDEX_FILE` points at `.git/worktrees/<name>/index`, and the child `git clone` writes the cloned module's blob OIDs INTO that index — with OIDs that exist only in `.terraform/modules/<name>/.git/objects/`. The subsequent `git commit` then fails to resolve those OIDs in the parent worktree's object DB.

This matches the maintainer's diagnosis on [pre-commit/pre-commit#1849](https://github.com/pre-commit/pre-commit/issues/1849): pre-commit framework deliberately does NOT scrub `GIT_*` from hook subprocess env (`no_git_env()` is used only internally for pre-commit's own git calls). asottile's [documented punt](https://github.com/pre-commit/pre-commit/blob/main/pre_commit/git.py#L20-L38): *"hook authors that do git writes need the same code as our `no_git_env` helper"*.

`pre-commit-terraform`'s hook scripts currently do NOT scrub `GIT_*`. This bug fires consistently for any user running pre-commit from a git worktree on a stack with module references.

### How can we reproduce it?

<details><summary>Reproduction script (~30s on a fresh setup)</summary>

```bash
# Setup
mkdir /tmp/repro && cd /tmp/repro
git init && git commit --allow-empty -m init
mkdir stack
cat > stack/main.tf <<'TF'
module "test" {
  source  = "terraform-aws-modules/iam/aws"
  version = "6.6.1"
}
TF
cat > .pre-commit-config.yaml <<'YAML'
repos:
    rev: v1.106.0
    hooks:
      - id: terraform_validate
        args:
          - --tf-init-args=-backend=false
YAML
git add -A && git commit -m "baseline"

# Create the worktree (key step — the bug only fires from a worktree)
git worktree add ../repro-wt
cd ../repro-wt

# Install hooks + trigger
pre-commit install
echo "" >> stack/main.tf
git add stack/main.tf
git commit -m "test"

# Observe: hooks all pass, then:
#   error: invalid object 100644 <oid> for '<file from cloned module repo>'
#   error: Error building trees
git fsck --no-progress | head    # → invalid sha1 pointer in cache-tree of worktree index
```

</details>

The phantom blob OIDs in the worktree's index resolve to files inside the cloned module repo (`.terraform/modules/test/.git/objects/`), confirming `git clone`'s blob writes went to the wrong index.

### Environment information

* OS: macOS 15.5 (Darwin arm64) — but the bug is platform-independent; we've seen it on Linux too
* `uname -a` output:

```bash
Darwin <host> 25.5.0 Darwin Kernel Version 25.5.0: Mon Apr 27 20:41:12 PDT 2026; root:xnu-12377.121.6~2/RELEASE_ARM64_T6050 arm64
```

```bash
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin25)
pre-commit 4.6.0
OpenTofu v1.12.0
Python 3.9.6
TFLint version 0.63.1
+ ruleset.terraform (0.15.0-bundled)
trivy Version: 0.70.0
```

* `.pre-commit-config.yaml`:

```yaml
repos:
    rev: v1.106.0
    hooks:
      - id: terraform_validate
        args:
          - --tf-init-args=-backend=false
```

</details>

### Proposed fix

Add a `common::scrub_git_env` helper to `hooks/_common.sh` that mirrors pre-commit framework's own [`no_git_env`](https://github.com/pre-commit/pre-commit/blob/main/pre_commit/git.py#L20-L38) helper (allowlist-based: keep `GIT_SSH` / `GIT_ASKPASS` / `GIT_CONFIG_*` etc. so SSH-source modules still authenticate; strip everything else `GIT_*`). Call it from `main()` of every hook script that shells to git, principally `terraform_validate.sh`, `terraform_tflint.sh`, and `terraform_docs.sh`.

Patch sketch (bash):

```bash
# hooks/_common.sh
function common::scrub_git_env {
  local var
  while IFS= read -r var; do
    case "$var" in
      GIT_CONFIG_KEY_*|GIT_CONFIG_VALUE_*) ;;
      GIT_EXEC_PATH|GIT_SSH|GIT_SSH_COMMAND|GIT_SSL_CAINFO) ;;
      GIT_SSL_NO_VERIFY|GIT_CONFIG_COUNT|GIT_HTTP_PROXY_AUTHMETHOD) ;;
      GIT_ALLOW_PROTOCOL|GIT_ASKPASS) ;;
      GIT_*) unset "$var" ;;
    esac
  done < <(env | sed -n 's/^\(GIT_[^=]*\)=.*/\1/p')
}
```

```bash
# hooks/terraform_validate.sh (and similarly tflint, docs)
function main {
  common::initialize "$SCRIPT_DIR"
  common::parse_cmdline "$@"
  common::export_provided_env_vars "${ENV_VARS[@]}"
  common::scrub_git_env       # ← NEW
  common::per_dir_hook "$HOOK_ID" "${#ARGS[@]}" "${ARGS[@]}" "${FILES[@]}"
}
```

Net diff: ~15 lines `_common.sh` + 1 line per affected hook. No behavior change for non-worktree users (their `GIT_*` env is empty for these vars).

I'm happy to open a PR if the maintainer is receptive to this approach. Wanted to file the issue first for discussion since the alternative (Docker mode per [pre-commit#3492](https://github.com/pre-commit/pre-commit/issues/3492)) is the path the pre-commit author has suggested for this class of bug; happy to defer to the maintainer's preference between (a) in-script env scrub here, or (b) "use the Docker variant".

### Related upstream issues

- [pre-commit/pre-commit#1849](https://github.com/pre-commit/pre-commit/issues/1849) — the canonical issue; asottile recommends hook authors apply `no_git_env`-equivalent themselves
- [pre-commit/pre-commit#3492](https://github.com/pre-commit/pre-commit/issues/3492) — asottile's stance that Docker mode is the design fix for `GIT_INDEX_FILE` leakage (closed-no-fix)
- [pre-commit/pre-commit#3515](https://github.com/pre-commit/pre-commit/issues/3515) — Docker mode + worktrees (closed-and-locked-as-off-topic)
- [`pre_commit/git.py:20-38`](https://github.com/pre-commit/pre-commit/blob/main/pre_commit/git.py#L20-L38) — the reference `no_git_env` implementation

yuriipolishchuk:
- [ ] This PR fixes a bug.

Fixes #992.

#### What

- New `common::scrub_git_env` helper in `hooks/_common.sh`: unsets 4 GIT_* vars that leak the parent repo location into child Git processes
- `hooks/terraform_validate.sh`: one-line call in `main()` after env setup, before `common::per_dir_hook`
- Zero behavior change for non-worktree users — `unset` is a no-op when the var is absent

#### Why

Hooks that invoke `tofu init` / `terraform init` (`terraform_validate`) end up running `git clone <module>` for each registry/git-source module. When `git commit` runs from a linked worktree, the parent git sets `GIT_INDEX_FILE` (pointing at `.git/worktrees/<name>/index`) in the hook subprocess env. The child `git clone` inherits this and writes cloned module blob OIDs into the parent worktree's index. The next commit fails at tree-build:

```
error: invalid object 100644 <oid> for '<path>'
error: Error building trees
```

pre-commit framework deliberately does NOT scrub `GIT_*` from user hook subprocesses (only its own internal git calls) per the maintainer in [pre-commit/pre-commit#1849](https://github.com/pre-commit/pre-commit/issues/1849):

> *they need the same code as in our [`no_git_env`] helper if they are dealing with doing git writes*

asottile [reaffirmed this](https://github.com/pre-commit/pre-commit/issues/3492#issuecomment-3200773733) in [pre-commit/pre-commit#3492](https://github.com/pre-commit/pre-commit/issues/3492) — framework-level scrubbing breaks valid use cases (e.g., gitleaks-docker wants `GIT_INDEX_FILE` passed through). Per-hook-author opt-in is the design intent. This patch applies that recommendation.

This is a targeted denylist (4 vars), NOT a mirror of pre-commit's allowlist-based `no_git_env` helper. Only the vars that leak the parent repository's location into child Git processes are unset; SSH/TLS/config-related `GIT_*` vars are preserved.

### How can we test changes

Same `.pre-commit-config.yaml` shape, same staged `.tf` change, same `git commit` invocation. Switch `repo:` between `antonbabenko/pre-commit-terraform@v1.106.0` and `yuriipolishchuk/pre-commit-terraform@fix/scrub-git-env`:

| Run | Config | Result |
|---|---|---|
| Unpatched | `rev: v1.106.0` | ❌ Commit fails after 27.4s with `Error building trees`. 5 phantom blobs in worktree index. `git fsck` reports invalid sha1 in cache-tree. |
| Patched | `rev: fix/scrub-git-env` | ✅ Commit succeeds in 4.7s. 0 phantom blobs. `git fsck` clean. `Terraform validate: Passed`. |

<details><summary>Full reproduction script</summary>

```bash
mkdir /tmp/repro && cd /tmp/repro
git init && git commit --allow-empty -m init
mkdir stack
cat > stack/main.tf <<'TF'
module "test" {
  source  = "terraform-aws-modules/iam/aws"
  version = "5.x"
}
TF
cat > .pre-commit-config.yaml <<'YAML'
repos:
    rev: v1.106.0
    hooks:
      - id: terraform_validate
        args:
          - --tf-init-args=-backend=false
YAML
git add -A && git commit -m baseline
git worktree add ../wt && cd ../wt
pre-commit install
echo "" >> stack/main.tf
git add stack/main.tf
git commit -m test           # → Error building trees on v1.106.0
                             # → Passes when rev is fix/scrub-git-env
git fsck --no-progress       # invalid sha1 pointer pre-fix; clean post-fix
```

</details>

- [x] Reproduce bug on `master` (v1.106.0) from a worktree with a module-using stack → commit fails with `Error building trees`, index has phantom blobs
- [x] Apply this patch via fork rev → same scenario passes, commit succeeds, index clean
- [x] Confirm non-worktree (main checkout) behavior unchanged — `unset` is a no-op when env var absent
- [x] Confirm SSH-source modules still authenticate — `GIT_SSH`, `GIT_ASKPASS`, `GIT_SSL_*` etc. are NOT touched
- [x] `shellcheck` clean (no new warnings)

### Other notes

- Considered the allowlist mirror of `no_git_env` (~15 lines `case` block in bash) but went with the minimal denylist since only `GIT_INDEX_FILE` actually causes the observed failure; the others are included as documented defensive offenders. Happy to expand to allowlist if preferred.
- The unset only takes effect for the duration of the hook and its child processes. The parent `git commit` env is unaffected.
- Hooks that don't shell to git (`terraform_fmt`, `terraform_trivy`, `terraform_checkov`, etc.) are not touched.

### References

- [pre-commit/pre-commit#1849](https://github.com/pre-commit/pre-commit/issues/1849) — the canonical "hook authors must scrub themselves" issue
- [pre-commit/pre-commit#3492](https://github.com/pre-commit/pre-commit/issues/3492) — asottile reaffirms framework-side scrub is rejected
- [`pre_commit/git.py:20-38`](https://github.com/pre-commit/pre-commit/blob/main/pre_commit/git.py#L20-L38) — `no_git_env` source + docstring

### Assisted-by

Per [AI_POLICY.md](https://github.com/antonbabenko/pre-commit-terraform/blob/master/.github/AI_POLICY.md): the patch, docstring, and PR text were drafted with LLM assistance (Claude Code). The bug was hit, root-caused, and reproduced in my real infrastructure repo; every revision here was A/B-tested by me against a worktree reproduction before pushing (unpatched fails at tree-build, patched commits cleanly), including re-verification after scoping down to terraform_validate in 63ab2bc.


hooks/_common.sh
```diff
@@ -116,6 +116,52 @@ function common::parse_cmdline {
   done
 }
 
+#######################################################################
+# Scrub four dangerous GIT_* env vars before child git operations.
```
```suggestion
# Scrub some GIT_* env vars before child Git operations.
```

hooks/_common.sh
```diff
@@ -116,6 +116,52 @@ function common::parse_cmdline {
...
+# `no_git_env` (pre_commit/git.py) uses an ALLOWLIST that scrubs all
+# GIT_* except ~11 known-safe vars (GIT_EXEC_PATH, GIT_SSH,
+# GIT_SSH_COMMAND, GIT_ASKPASS, GIT_SSL_CAINFO, GIT_SSL_NO_VERIFY,
+# GIT_CONFIG_KEY_*, GIT_CONFIG_VALUE_*, GIT_CONFIG_COUNT,
+# GIT_HTTP_PROXY_AUTHMETHOD, GIT_ALLOW_PROTOCOL). It runs only on
+# pre-commit's INTERNAL git calls, not on user hook subprocesses
+# (per pre-commit/pre-commit#1849).
+#
+# We unset only the four vars that pre-commit's own docstring (git.py
+# lines 20-38) names as dangerous when leaked into child git:
```
yermulnik: * Please use real URLs instead of placeholders (e.g. `https://github.com/pre-commit/pre-commit/blob/main/pre_commit/git.py` instead of `pre_commit/git.py` and `https://github.com/pre-commit/pre-commit/issues/1849` instead of `per pre-commit/pre-commit#1849`) so that others can click URLs inside the IDE of their preference.
* The line range does not correctly point to the function in question. As this may happen when referred file is updated, I see no reason to point to line range. Better to point to a file and name the function so look up.
* I find no word "dangerous" in referred function. And that's correct as those env vars aren't dangerous. They're just, sort of, incompatible with `pre-commit`. Please reword.
* I also see no good reason to specify number of vars filtered out (unset) as we may shrink or expand the list over the time and the word "four" will eventually and inevitably become obsolete.

hooks/_common.sh
```diff
@@ -116,6 +116,52 @@ function common::parse_cmdline {
...
+#
+#   GIT_INDEX_FILE        proximate cause; "Causes 'error invalid
+#                         object ...' during commit" in git.py.
+#   GIT_DIR               "Causes git clone to clone wrong thing".
+#   GIT_WORK_TREE         defensive; paired with GIT_DIR.
+#   GIT_OBJECT_DIRECTORY  defensive; prevents child writes into the
+#                         parent object DB via non-clone git calls.
```
yermulnik: * The docstring in `no_git_env()` func lists only three vars. Where the `GIT_OBJECT_DIRECTORY` comes from?
* Also the ` in git.py.` bit is confusing. Please drop it.
* `Git` is a proper name. Please capitalize it where it is not used as name of the CLI tool.

hooks/_common.sh
```diff
@@ -116,6 +116,52 @@ function common::parse_cmdline {
...
+# Denylist over allowlist mirror because:
+#   1. Bash equivalent of the Python whitelist scan is ~15-20 lines
+#      versus a one-line `unset`.
+#   2. The observed bug is fully fixed by scrubbing GIT_INDEX_FILE
+#      alone; the other three are belt-and-braces.
+#   3. Narrower scrub minimizes risk of breaking hook authors who
+#      rely on inheritance of vars an aggressive allowlist might drop.
+#######################################################################
+function common::scrub_git_env {
+  unset GIT_INDEX_FILE GIT_DIR GIT_WORK_TREE GIT_OBJECT_DIRECTORY
```
yermulnik: * Please put vars in alphabetical order.
* Might it worth of handling `unset` failure? E.g. if by any reason the var was declared readonly outside of the hook? 🤔
* I'd also suggest to unset Git vars only when the hook is run from inside the Git worktree so that we don't mess up non-worktree envs.
  * Here's how I identify this locally:
  ```bash
  local _git_repo_git_file="$(git rev-parse --show-cdup).git"
  [[ -f $_git_repo_git_file ]] && $(grep -qE ^gitdir: "$_git_repo_git_file") && local _is_git_worktree="true"
  ```
* Would we want to let user know what's happening as it's sort of intrusion to user-controlled area? 🤔 Maybe not unconditionally, but e.g. when `PCT_LOG` is set to e.g. `debug`… @MaxymVlasov WDYT
* Would it make sense to apply `common::scrub_git_env()` to **all** hooks e.g. by making it part of `common::initialize()` (or being called from `common::initialize()`)? 🤔 This way we would not need to care of, for example, adding it to any new hooks (if any). @MaxymVlasov WDYT?

hooks/terraform_docs.sh
```diff
@@ -20,6 +20,7 @@ function main {
   common::parse_cmdline "$@"
   common::export_provided_env_vars "${ENV_VARS[@]}"
   common::parse_and_export_env_vars
+  common::scrub_git_env
```
yermulnik: Out of interest, does the `hooks/terraform_providers_lock.sh` not require Git env var scrubbing?

hooks/terraform_tflint.sh
```diff
@@ -13,6 +13,7 @@ function main {
   common::parse_cmdline "$@"
   common::export_provided_env_vars "${ENV_VARS[@]}"
   common::parse_and_export_env_vars
+  common::scrub_git_env
```
MaxymVlasov: I can't reproduce the worktree bug in this hook. 
```suggestion
```

hooks/terraform_docs.sh
MaxymVlasov: I can't reproduce the worktree bug in this hook. 
```suggestion
```

MaxymVlasov: There another thing that shows, both for `terraform_docs` and `terraform_fmt` hooks.

But it's not an issue at all.
```bash
18:57 repro-hoo git:(repro-hoo +) 
➜ gc       
Terraform docs...........................................................Passed
Terraform validate with tflint...........................................Passed
Terraform fmt............................................................Failed
- hook id: terraform_fmt

main.tf


18:57 repro-hoo git:(repro-hoo !+) 
✘1 ➜ gc
Terraform docs...........................................................Passed
Terraform validate with tflint...........................................Passed
Terraform fmt............................................................Passed
Terraform docs.......................................(no files to check)Skipped
Terraform validate with tflint.......................(no files to check)Skipped
Terraform fmt........................................(no files to check)Skipped
[repro-hoo 44cd287] fmt
 2 files changed, 2 insertions(+), 6 deletions(-)

18:58 repro-hoo git:(repro-hoo) took 9s 
➜ git fsck --no-progress | head
dangling tree bba236455e65937967031c5d25b02e11f5c252f5

18:58 repro-hoo git:(repro-hoo) 
➜ git fsck --no-progress | head
dangling tree bba236455e65937967031c5d25b02e11f5c252f5

18:58 repro-hoo git:(repro-hoo) 
➜ gc                           
- hook id: terraform_docs
Terraform validate with tflint...........................................Passed
Terraform fmt............................................................Passed

18:58 repro-hoo git:(repro-hoo !+) 
✘1 ➜ gc
Terraform docs...........................................................Passed
Terraform validate with tflint...........................................Passed
Terraform fmt............................................................Passed
Terraform docs.......................................(no files to check)Skipped
Terraform validate with tflint.......................(no files to check)Skipped
Terraform fmt........................................(no files to check)Skipped
[repro-hoo c115ce4] docs
 2 files changed, 2 insertions(+), 2 deletions(-)

18:58 repro-hoo git:(repro-hoo) took 6s 
➜ git fsck --no-progress | head
dangling tree a13f3782d35b8ce91a9d45f5b57acd390ea671a2
dangling tree bba236455e65937967031c5d25b02e11f5c252f5
```

MaxymVlasov (changes requested):

yuriipolishchuk:
Thanks both for the thorough review - pushed 63ab2bc addressing it.

### Scope (@MaxymVlasov)

Accepted both suggestions: `common::scrub_git_env` calls removed from `terraform_tflint.sh` and `terraform_docs.sh`. The helper is now called only from `terraform_validate.sh` - the confirmed reproduction path (`validate -> tofu/terraform init -> git clone`). I also re-verified on my side that the validate-only scrub fully fixes the original worktree failure (fresh module clone during commit, index stays clean, `git fsck` clean, follow-up commits healthy).

The dangling trees you observed with `terraform_docs`/`terraform_fmt` make sense as a separate, harmless artifact: pre-commit's own stash/restore cycle writes tree objects that end up unreferenced when the first commit attempt is rejected with "files were modified". No index corruption involved.

### Wording / references (@yermulnik)

All applied in 63ab2bc:

- Function summary now uses your suggested line verbatim.
- "dangerous" dropped; reworded around the parent-repository-location leak.
- Var count removed from the docstring (won't go stale if the list changes).
- Full clickable URLs instead of repo-relative paths and issue shorthand; line-range references dropped.
- "Git" capitalized as a proper name (kept lowercase only in literal commands).
- `unset` arguments alphabetized.
- `GIT_OBJECT_DIRECTORY` provenance corrected - you're right that pre-commit's `no_git_env` docstring names only three vars. It comes from Git's own repository-location environment variables (https://git-scm.com/docs/git#_environment_variables); the docstring now cites that source for all four. Happy to drop it down to the three if you'd rather keep the list minimal.

### Open design questions - deferring to you two

1. **Scrub in `common::initialize` for all hooks vs per-hook opt-in** - I see the appeal of central placement (no per-hook bookkeeping), but it would also apply the unset to hooks that demonstrably don't need it, and Maxym's reproduction results argue for the narrow version. Either shape is a small diff; I'll implement whichever you align on.
2. **Worktree-detection gating** - the `unset` is a no-op when the vars aren't set (non-worktree `git commit` doesn't export them into hooks in a way that breaks anything we've seen), so the gate would add code without changing behavior. But if you prefer the explicit guard for the "don't touch user env unless necessary" principle, I can add your `rev-parse --show-cdup` check.
3. **`PCT_LOG=debug` notice when scrubbing** - happy to add if you want it; suggest doing it together with whatever logging convention you settle on for (2).
4. **readonly `unset` failure** - a var marked readonly outside the hook would make `unset` error under `set -e`. I can add `2> /dev/null || true`, though a readonly `GIT_INDEX_FILE` in a pre-commit env would already be broken in stranger ways. Your call.
5. **`terraform_providers_lock`** - it runs `init`-adjacent provider fetches but not module `git clone` to my knowledge; I didn't have a failing reproduction there, so I left it out per the same evidence standard Maxym applied to tflint/docs.

yuriipolishchuk:
@MaxymVlasov @yermulnik gentle nudge - whenever you have a moment for a re-review.

Recap of what changed since the review round, all in `63ab2bc`:

- **Scope narrowed to `terraform_validate` only** - removed the `common::scrub_git_env` calls from `terraform_tflint` and `terraform_docs` per @MaxymVlasov (you couldn't reproduce there, and I confirmed validate-only fully fixes the original worktree failure).
- **Docstring reworded** per @yermulnik - dropped "dangerous" and the var count, capitalized Git, full clickable URLs instead of repo-relative paths / issue shorthand, no line-range references, corrected `GIT_OBJECT_DIRECTORY` provenance (Git's own env-var docs, not `no_git_env`), alphabetized the `unset` args.
- Branch updated to current `master`; all CI green.

I also left replies on the open design questions (central `common::initialize` placement, worktree-detection gating, `PCT_LOG` notice, readonly-`unset` handling, `terraform_providers_lock`) - happy to implement whichever direction you two prefer on those. No rush; just flagging it's ready when you are.

yermulnik (dismissed):
@yuriipolishchuk Thanks for a nudge. The code looks good to me. However I'd defer the final review/approval to @MaxymVlasov as main dev.

UPD: since the PR targets Git worktrees only, I'd prefer the worktree gating there so that the code changes user env only where the problem exists rather than everywhere. @MaxymVlasov What's your view on this in particular and the PR overall?

MaxymVlasov (dismissed):

yermulnik (dismissed):
@MaxymVlasov Code LGTM.
The two points which I'm concerned about, however I might be over-engineering these. @MaxymVlasov WDYT?
* Handle `unset` failure? E.g. if by any reason the var was declared readonly outside of the hook? 🤔 The `unset … || true` should be sufficient (not suppressing error output to let user know what's happening wrongly)
* I'd also suggest to unset Git vars only when the hook is run from inside the Git worktree so that we don't mess up non-worktree envs.
  * Here's how I identify this locally:
  ```bash
  local _git_repo_git_file="$(git rev-parse --show-cdup).git"
  [[ -f $_git_repo_git_file ]] && $(grep -qE ^gitdir: "$_git_repo_git_file") && local _is_git_worktree="true"
  ```

MaxymVlasov (dismissed):

hooks/_common.sh
```diff
@@ -116,6 +116,38 @@ function common::parse_cmdline {
...
+# the parent repository's location into child Git processes:
+#
+#   GIT_DIR               makes child Git operate on the parent repo
+#   GIT_INDEX_FILE        proximate cause of the failure above
+#   GIT_OBJECT_DIRECTORY  redirects child object writes into the
+#                         parent object database
+#   GIT_WORK_TREE         pairs with GIT_DIR
+#######################################################################
+function common::scrub_git_env {
+  local -r git_env_vars=(
```
yermulnik: Why readonly? And why not also `-a` to explicitly declare array type?

hooks/_common.sh
```diff
@@ -116,6 +116,38 @@ function common::parse_cmdline {
...
+    GIT_DIR
+    GIT_INDEX_FILE
+    GIT_OBJECT_DIRECTORY
+    GIT_WORK_TREE
+  )
+  local env_var
+  for env_var in "${git_env_vars[@]}"; do
+    unset "$env_var" 2> /dev/null || true
```
yermulnik: * IMHO there's no sense to use loop instead of multiple args.
* I also think we should not suppress error as this would complicate debugging.
* Maybe also add `-v` so that only shell vars are attempted to be unset 🤔

hooks/_common.sh
MaxymVlasov: As I don't expect that this array will be changed

yermulnik (approved):

antonbabenko:
This PR is included in version 1.107.0 :tada:

PR #1001, opened 2026-07-23
fix(docker): Downloader of `opentofu` to specific version was broken, only `latest` worked

MaxymVlasov:
- [x] This PR fixes a bug.

#### What

- Fixed `tools/install/opentofu.sh`'s `GH_RELEASE_REGEX_SPECIFIC_VERSION` regex, which never matched any real OpenTofu release asset. It expected the filename itself to start with a `v` glued to the version (`v${VERSION}_${TARGETOS}...`), but OpenTofu's actual asset filenames have no `v` prefix (e.g. `tofu_1.12.5_linux_amd64.tar.gz`) - the `v` only appears in the release tag's URL *path segment* (`.../download/v1.12.5/tofu_1.12.5_linux_amd64.tar.gz`), not the filename.
- Net effect before this fix: pinning any specific OpenTofu version (via `--build-arg OPENTOFU_VERSION=X.Y.Z` at Docker build time, or the `--tool-version=X.Y.Z` hook-config from #) silently failed to download - only `OPENTOFU_VERSION=latest` ever worked, since `GH_RELEASE_REGEX_LATEST` doesn't reference the version at all.

#### Why

Accidentally found  during work on a different PR.

### How can we test changes

```bash
TARGETOS=linux TARGETARCH=amd64 VERSION=1.12.5
curl -s https://api.github.com/repos/opentofu/opentofu/releases | grep -o -E -i -m 1 "https://.+?_${VERSION}_${TARGETOS}_${TARGETARCH}.tar.gz"
# before this fix: no output (regex never matches)
# after this fix:  https://github.com/opentofu/opentofu/releases/download/v1.12.5/tofu_1.12.5_linux_amd64.tar.gz
```

Also verified with a full end-to-end download/extract of the resulting URL - the extracted binary correctly reports `OpenTofu v1.12.5`.

### Assisted-by
- Assisted-by: Sisyphus:claude-sonnet-5 claude (da0828e)


yermulnik (approved):

antonbabenko:
This PR is included in version 1.108.1 :tada:

Issue #722, opened 2024-10-01
Automate dependencies


actuarysailor:
### What problem are you facing?

If you use the integrated[ pre-commit GitHub application](https://github.com/marketplace/pre-commit-ci), the fact that the docker images do not automatically download, install, or have the dependencies pre-defined causes issues.

Could you please update the hooks so that they either download or come pre-loaded with a version of the tools many of these hooks are dependent on?

### How could pre-commit-terraform help solve your problem?

Either add a parameter in the hooks to download a specific version of the tools, or pre-load with latest version of tools.

![image](https://github.com/user-attachments/assets/09f3a7bb-2609-45cd-b357-af7783b6cee5)

yermulnik:
I'm not familiar with https://pre-commit.ci platform, though under the [Github Actions](https://github.com/antonbabenko/pre-commit-terraform?tab=readme-ov-file#github-actions) section of `README` there's an example of how to use `pre-commit-terraform` Docker image, which has all required tools pre-installed — is it any helpful in regards to your use case? 🤔

webknjaz:
@yermulnik pre-commit.ci separates what it does into two stages.

The first one is preparing the testing environment, it installs the deps (including `additional_dependencies`), creates venvs and such. It has network access, and I'm pretty sure it uses the feature I requested years ago: https://github.com/pre-commit/pre-commit/issues/456#issuecomment-270398182. This environment is then cached.

The second step is running the checks. It does not provision anything. Moreover, it has networking disconnected from the env where it's being executed. This stage is not supposed to rely on network access in any way or attempt to provision things, since it's expected that the provisioned env is cached and never restarted until said cache is invalidated.

So the only way to address this is to provide the dependencies in a way that is native to pre-commit.

MaxymVlasov:
> This stage is not supposed to rely on network access in any way or attempt to provision things,

That's a problem for `tflint` and `terrascan` in some setups :thinking: 

tflint requires `tflint --init` for `.tflint.hcl` which includes plugins:

```hcl
plugin "aws" {
    enabled    = true
    version    = "0.37.0"
    source     = "github.com/terraform-linters/tflint-ruleset-aws"
}
```

In theory, we could init all available plugins and pack them for `pre-commit.ci` somewhere, but, that's will ot work if user uses outdated version or we not fast enough to update our build. In other words, that requires perfect sync from us and all `pre-commit.ci` users of `terraform_tflint` hook.
[...]

Issue #270, opened 2021-10-26
[3rd party: tflint] terraform_tflint.sh still throws "command not found" 


shoof1072:
### terraform_tflint.sh still throws "command not found" 

Expected: This behavior was expected to be fixed by PR #251  
Observed: When intentionally changing a terraform file to cause a linting failure observed the following
```
$ pre-commit
Terraform fmt............................................................Passed
Terraform validate with tflint...........................................Failed
- hook id: terraform_tflint
- exit code: 127


ERROR in terraform/:
$HOME/.cache/pre-commit/repohu82fesd/terraform_tflint.sh: line 70: tflint: command not found
```

### How can we reproduce it?

Well, interestingly this error is observed even when there is a formatting error detected by terraform_fmt.

The relevant section of the `.pre-commit-config.yaml` used when trying to add terraform_tflint is:
```
repos:
  - repo: https://github.com/antonbabenko/pre-commit-terraform
    rev: e33c654a4005cff364a42f965068160e4ba3e3c2  
    hooks:
      - id: terraform_fmt
      - id: terraform_tflint
```

So, for example, if I have the following terraform:
```
resource "aws_iam_role_policy" "some_iam_policy" {
  count  = my_count
  policy = my_policy_document
  role   = arbitrary_role
}
```

And I change the formatting to cause terraform_fmt to catch and reformat due to expressions no longer aligning on the equals operator, ex:

```
resource "aws_iam_role_policy" "some_iam_policy" {
  count  = my_count
  policy = my_policy_document
  role = arbitrary_role
}
```

Then the observed behavior is that terraform_fmt will catch and fix the formatting error, but the `terraform_tflint.sh` script will will return "file not found".

If I intentionally cause a linting error, for example changing the keyword `source` in a module to `source_`

```
module "my_module" {
  source_ = "git::ssh:/git@some_url"
  ...
  ...
}     
```

The same error is observed:
```
ERROR in terraform/:
$HOME/.cache/pre-commit/repohu82fesd/terraform_tflint.sh: line 70: tflint: command not found
```

### Environment information

* OS:  Ubuntu 21.10

* `uname -a` and/or `systeminfo | Select-String "^OS"` output:

```bash
$ uname -a
Linux shoofcrash01 5.13.0-20-generic #20-Ubuntu SMP Fri Oct 15 14:21:35 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```

* Tools availability and versions:

```bash
[...]

PR #1002, opened 2026-07-23
feat: Add ability to download specific tool versions used by hooks. TLDR: add `--hook-config=--tool-version=1.2.3` for pinning wrapped tool version. Works for all tools (even `terraform` and `tofu`) except `checkov`

### LINKED ISSUE #188: Allow to specify the version of terraform-docs for the hook to use

MaxymVlasov:
- [x] This PR fixes a bug.

#### What

- Adds `--hook-config=--tool-version=X.Y.Z` version pinning to every hook that wraps a downloadable release binary (`terraform_tflint`, `terraform_docs`, `terraform_tfsec`/`terraform_trivy`, `terrascan`, `infracost_breakdown`, `tfupdate`, `terragrunt_*`, and Terraform/OpenTofu via `terraform_validate`/`terraform_fmt`/`terraform_providers_lock`), resolved and cached on demand by invoking the existing `tools/install/<tool>.sh` installer scripts - opt-in, zero effect on hooks that don't set it.
- Adds `--hook-config=--tool-version-mode=strict|prefer-local` to control whether a pinned version always wins (`strict`, default) or an already-on-`$PATH` binary is preferred when present (`prefer-local`).
- On `terraform_validate`/`terraform_fmt`/`terraform_providers_lock`, `--tf-path` can now also be set to the literal value `terraform`, `opentofu`, or `tofu` (instead of an actual binary path) when combined with `--tool-version`, to explicitly pick which of the two to pin rather than relying on `$PATH`-based auto-detection.
- Adds an actionable error (naming the missing tool and the hook, suggesting `--hook-config=--tool-version=`) when a hook's tool is neither pinned nor found on `$PATH`, instead of deferring to a generic "command not found" failure later.
- `checkov`/`terraform_checkov` is explicitly excluded (pip-distributed, not a GitHub-release binary - forcing it through this path would fight its actual, already-working distribution model).
- New `README.md` section documenting usage, a cache-directory-mounting note for persisting across Docker container runs, and a Renovate `customManagers` recipe for keeping pins auto-bumped.
- Updates `.github/CONTRIBUTING.md`'s "Add new hook" checklist to mention the new runtime call site alongside the existing Docker build-time one.
- One small, backward-compatible change to `tools/install/_common.sh`: `source /.env` is now conditional on the file existing, so the installer scripts also run correctly when invoked directly at hook run-time (not just inside a Docker build).
- Fixes a bug where a failed tool resolution (`common::resolve_tool_path`) was silently swallowed by `local -r tool_path=$(...)`, letting `terraform_tflint`, `terraform_docs`, `infracost_breakdown`, and the shared `per_dir_hook` continue with an empty tool path instead of aborting with the resolver's own error message.
- Corrects stale `tool_path` doc comments left over from the `tf_path`->`tool_path` rename in 9 hooks, and fixes invalid YAML indentation in 3 `--tool-version` README examples.
- Fixes `terragrunt_validate`/`terragrunt_validate_inputs`/`terragrunt_fmt`/`terragrunt_providers_lock` picking their pre/post-0.78 CLI syntax off whatever `terragrunt` is on `$PATH` instead of the resolved/pinned `tool_path` binary that actually runs - a pinned version on the other side of the 0.78 boundary from `$PATH` picked the wrong syntax and broke.
- Fixes 5 bugs flagged by Copilot's automated review (#4992979471): `prefer-local` mode probed `opentofu` on `$PATH` instead of the real `tofu` binary so the pin always won regardless of mode; the published Docker image had no `curl`, breaking cache-miss downloads; `terrascan`'s installer used bash4-only `${TARGETOS^}`, breaking under macOS's bash 3.2; the cache key omitted OS/arch, so a cache dir shared across platforms (e.g. a Docker volume mount) could serve an incompatible binary as a false hit; and specific-version resolution only searched GitHub's unpaginated, 30-release page, so pinning anything older silently failed.
- Wires `hcledit` through the same resolver for `terraform_wrapper_module_for_each` - the last release-binary hook `--tool-version` didn't yet cover.
- Fixes the Docker cache-mount README example, which mounted under `/root` while also dropping to a non-root `USERID` (unreachable, permission denied).
- Silences a `hadolint` DL3064 false positive on the pre-existing `ARG GITHUB_TOKEN`/`ENV INFRACOST_API_KEY` (both empty-default placeholders, real values supplied externally, never hardcoded).
- Fixes CI Docker `buildx` builds failing with e.g. `tflint.sh ... exit code: 3`: `build-image.yaml`/`build-image-test.yaml` passed `GITHUB_TOKEN` via buildx `secrets:`, which nothing in the Dockerfile mounts - dead config left over from before `--build-arg`-based GitHub API auth was added, so every image build resolved tool versions via unauthenticated (60/hr, shared across all Actions runners) GitHub API calls. Now passed as a `GITHUB_TOKEN` build-arg, matching what `tools/install/_common.sh` already expects.

#### Why

This enables multi-version testing and the ability to pin the same versions of underlying tools by users, which is a decent improvement towards reproducibility.

Fix: https://github.com/antonbabenko/pre-commit-terraform/issues/188, https://github.com/antonbabenko/pre-commit-terraform/issues/270, https://github.com/antonbabenko/pre-commit-terraform/issues/1009
Related: https://github.com/antonbabenko/pre-commit-terraform/issues/722

### How can we test changes

```yaml
- id: terraform_tflint
  args:
    - --hook-config=--tool-version=0.50.0
```

Run `pre-commit run terraform_tflint` - first run downloads and caches tflint `0.50.0` into `$XDG_CACHE_HOME/pre-commit-terraform` (or `$HOME/.cache/pre-commit-terraform`), subsequent runs hit the cache with no network call. Same pattern works for `terraform_validate`/`terraform_fmt`/`terraform_providers_lock` (pins Terraform/OpenTofu), optionally combined with `--hook-config=--tf-path=opentofu` to force OpenTofu specifically regardless of what's on `$PATH`.

### Assisted-by
- Assisted-by: Sisyphus:claude-sonnet-5 opencode (b85f444, a5e62cc, a432cff, a475d47, 7892df6, 277e7cd, fb8c3f8)
- Assisted-by: Sisyphus:claude-sonnet-5 claude (869f6b6)


hooks/_common.sh
```diff
@@ -328,13 +330,14 @@ function common::per_dir_hook {
   # despite there's only one positional ARG left
   local -a -r files=("$@")
 
-  local -r tf_path=$(common::get_tf_binary_path)
+  local -r tool_version=$(common::get_hook_config_value "--tool-version")
+  local -r tool_path=$(common::resolve_tool_version "$tool_name" "$tool_version")
```
yermulnik: Should the both values be validated to be non-empty so that the erroneous result are not passed down the code? 🤔

hooks/_common.sh
```diff
@@ -492,9 +495,229 @@ function common::colorify {
...
+
+#######################################################################
+# Resolve a specific version of a wrapped tool's binary, downloading
+# and caching it on demand if it isn't already cached. Reuses the
+# existing `tools/install/<tool>.sh` installer script (invoked as a
+# subprocess, with its working directory redirected to the cache
+# directory) instead of re-implementing per-tool download logic.
+# The sentinel tool name "tf" means "resolve Terraform or OpenTofu,
+# decide which" - it delegates to `common::get_tf_binary_path` for the
+# extra precedence rungs that need (--tf-path, PCT_TFPATH/
```
yermulnik: "That need" is ambiguous and probably is meant to be "that are needed".

It's also the first time I meet the word "rungs" in such context =)) But it's just my personal experience.

hooks/_common.sh
```diff
@@ -492,9 +495,229 @@ function common::colorify {
...
+# TERRAGRUNT_TFPATH env vars, and picking between the two tools). "tf"
+# is never itself a real, installable tool - `common::get_tf_binary_path`
+# calls back into this function with the concrete decided name
```
yermulnik: The "into" looks to be extraneous/redundant in this context.

hooks/_common.sh
```diff
@@ -492,9 +495,229 @@ function common::colorify {
...
+#     no resolvable binary (e.g. checkov)
+#   version (string) exact version requested (e.g. "1.7.5"), or empty
+#     if no `--tool-version` was requested
+# Outputs:
+#   Prints the absolute path to the resolved binary, or the bare $tool
+#   name unchanged if no version was requested (empty string if $tool
+#   itself is also empty). If a download is attempted and fails - exit
+#   1 with an error message.
+#######################################################################
+function common::resolve_tool_version {
```
yermulnik: The name of the function suggests it returns resolved tool version whilst in reality it returns resolved tool **path**.
Top eliminate this bit of confusion I suggest to name the function more clearly, e.g. `resolve_tool_path`. WDYT?
```suggestion
# Outputs:
#   Prints the absolute path to the resolved binary, or the bare $tool
#   name unchanged if no version was requested (empty string if $tool
#   itself is also empty). If a download is attempted and fails - exit
#   1 with an error message.
#######################################################################
function common::resolve_tool_path {
```

hooks/_common.sh
```diff
@@ -492,9 +495,229 @@ function common::colorify {
...
+
+  case "$TARGETARCH" in
+    x86_64) TARGETARCH="amd64" ;;
+    aarch64 | arm64) TARGETARCH="arm64" ;;
+  esac
+
+  export TARGETOS TARGETARCH
+}
+
+#######################################################################
```
yermulnik, line 547: Jeez, this is pretty huge docstring 😕 Do we really want to have it as a comment here or maybe move it over to some readme file? 🤔

hooks/_common.sh
```diff
@@ -492,9 +495,229 @@ function common::colorify {
...
+#   name unchanged if no version was requested (empty string if $tool
+#   itself is also empty). If a download is attempted and fails - exit
+#   1 with an error message.
+#######################################################################
+function common::resolve_tool_version {
+  local -r tool_name="$1"
+  local -r version="$2"
+
+  #
+  # Check is configuration valid
```
```suggestion
  # Check configuration validity
```
or
```suggestion
  # Check if configuration is valid
```

hooks/_common.sh
```diff
@@ -492,9 +495,229 @@ function common::colorify {
...
+  #
+
+  # No resolvable tool name (e.g. checkov, which is pip-distributed);
+  # keeps "--tool-version" a documented no-op for it instead of erroring on an empty tool name.
+  [[ -z $tool_name ]] && return
```
yermulnik: We require Bash, so why wouldn't we use Bashisms =)
```suggestion
  [[ ! $tool_name ]] && return
```

hooks/_common.sh
```diff
@@ -492,9 +495,229 @@ function common::colorify {
...
+
+  # No version requested, and not the "tf" placeholder - skip resolution,
+  # use whatever's on $PATH. The "tf" placeholder is excluded here because
+  # get_tf_binary_path has its own additional precedence rungs
+  # (--tf-path, PCT_TFPATH, TERRAGRUNT_TFPATH) that must be consulted
+  # even when $version is empty, unlike every other tool.
+  if [[ -z $version && $tool_name != tf ]]; then
```
```suggestion
  if [[ ! $version && $tool_name != tf ]]; then
```

hooks/_common.sh
```diff
@@ -492,9 +495,229 @@ function common::colorify {
...
+    if [[ ! $(command -v "$tool_name") ]]; then
```
```suggestion
    if ! command -v "$tool_name" > /dev/null; then
```

hooks/_common.sh
```diff
@@ -492,9 +495,229 @@ function common::colorify {
...
+      common::colorify "red" \
+        "ERROR: '$tool_name' is required by '$HOOK_ID' pre-commit hook but it is not discoverable in the system's PATH.\n" \
+        "Since '--hook-config=--tool-version' was not specified, no version resolution was attempted.\n\n" \
```
yermulnik: I don't think it's a good idea to have a trimmed version of the config bit printed to user as it may introduce a confusion.
```suggestion
        "Since '--hook-config=--tool-version=…' was not specified, no version resolution was attempted.\n\n" \
```

hooks/_common.sh
```diff
@@ -492,9 +495,229 @@ function common::colorify {
...
+    return
+  fi
+
+  #
+  # Choose whether to prefer the local $PATH version of a tool over a requested version, if both exist.
+  #
+  local -r tool_version_mode=$(common::get_hook_config_value "--tool-version-mode")
+
+  if command -v "$tool_name" &> /dev/null; then
+    if [[ $tool_version_mode == "prefer-local" ]]; then
```
yermulnik: For the consistency's sake it's best to either wrap RHS in conditional expressions in double quotes or do not wrap it **everywhere** please.
The `$tool_name == tf` vs the `$tool_version_mode == "prefer-local"`.

hooks/_common.sh
```diff
@@ -492,9 +495,229 @@ function common::colorify {
...
+  # common::install_from_gh_release completes (see tools/install/opentofu.sh)
+  local resolved_bin_name="$tool_name"
+  [[ $tool_name == opentofu ]] && resolved_bin_name="tofu"
+
+  local -r cache_root="${PCT_TOOL_CACHE_DIR:-${XDG_CACHE_HOME:-$HOME/.cache}/pre-commit-terraform}"
+  local -r cache_dir="$cache_root/$tool_name/$version"
+  local -r cached_bin="$cache_dir/$resolved_bin_name"
+
+  if [[ -x $cached_bin ]]; then
+    echo "$cached_bin"
```
yermulnik, line 659: Might it make sense to also print a green message that the cached tools is to be used? 🤔

hooks/_common.sh
```diff
@@ -492,9 +495,229 @@ function common::colorify {
...
+  # Redirect the installer's own stdout to stderr: this function's stdout is
+  # a contract (the resolved path, captured via "$(...)" by every caller),
+  # and installers like terraform.sh/tflint.sh call bare `unzip` (no `-q`),
+  # which prints "Archive: ... inflating: ..." to stdout by default -
+  # harmless noise in a Docker build log, but it would otherwise corrupt
+  # the path this function returns.
+  if ! (
+    cd "$cache_dir" || exit 1
+    export "$env_var_name=$version"
+    bash "$installer_script" 1>&2
```
yermulnik: Do installer scripts have no shebang line or why do we need to explicitly override/provide interpreter? 🤔

hooks/_common.sh
```diff
@@ -492,9 +495,229 @@ function common::colorify {
...
+}
+
 #######################################################################
 # Get Terraform/OpenTofu binary path
 # Allows user to set the path to custom Terraform or OpenTofu binary
+# Called from `common::resolve_tool_version`'s "tf" sentinel dispatch,
+# not directly by hooks - every hook (including
+# terraform_validate/fmt/providers_lock, which pass tool_name="tf")
+# goes through the same `common::per_dir_hook` -> `common::resolve_tool_version`
+# path. When this function calls back into `common::resolve_tool_version`
```
yermulnik: I don't think that "calls into" is contextually correct in this context as to "call into" means literally to phone someone or to enter somewhere I guess.

hooks/_common.sh
```diff
@@ -492,9 +495,229 @@ function common::colorify {
...
+# with the concrete decided tool ("terraform" or "opentofu") for the
+# actual pin-and-cache step, that name no longer matches "tf", so the
+# dispatch above doesn't fire a second time.
```
yermulnik: What "the dispatch above" refers to? 😕

hooks/_common.sh
```diff
@@ -503,21 +726,45 @@ function common::colorify {
...
-  done
+  local -r hook_config_tf_path=$(common::get_hook_config_value "--tf-path")
 
-  # direct hook config, has the highest precedence
-  if [[ $hook_config_tf_path ]]; then
+  # direct hook config, has the highest precedence - but only when NOT
+  # combined with --tool-version. When it IS also set, --tf-path is
+  # reinterpreted below as an explicit terraform/opentofu selector
+  # rather than a literal binary path.
+  if [[ $hook_config_tf_path && -z $tool_version ]]; then
```
yermulnik: Mixing Bashism with POSIX isn't great approach =)
```suggestion
  if [[ $hook_config_tf_path && ! $tool_version ]]; then
```

hooks/_common.sh
```diff
@@ -503,21 +726,45 @@ function common::colorify {
...
+      terraform)
+        tf_tool="terraform"
+        ;;
+      opentofu | tofu)
+        tf_tool="opentofu"
+        ;;
+      "")
+        # Terraform preferred; opentofu only if terraform isn't on $PATH but tofu is).
+        tf_tool="terraform"
+        ! command -v terraform &> /dev/null && command -v tofu &> /dev/null && tf_tool="opentofu"
```
yermulnik, line 748: Just in case: if neither `terraform` nor `tofu` are discoverable by `command -v`, then the `tf_tool` will end up undefined. Hope this won't break the logic of the dependants 🤞🏻

hooks/_common.sh
```diff
@@ -503,21 +726,45 @@ function common::colorify {
...
+        ;;
+      *)
+        common::colorify "red" \
+          "ERROR: '--tf-path=$hook_config_tf_path' combined with '--tool-version' is not a valid value.\n" \
```
yermulnik, line 752: Usually something like below is used (if I'm not mixing up the context):
```suggestion
          "ERROR: the '--tf-path=$hook_config_tf_path' and the '--tool-version' are mutually exclusive.\n" \
```

hooks/_common.sh
```diff
@@ -503,21 +726,45 @@ function common::colorify {
...
+          "It must be either 'terraform', 'opentofu'/'tofu', or unset (to auto-detect)."
```
yermulnik: "It" as in what: the `--tf-path` or `--tool-version`? 😕

hooks/_common.sh
```diff
@@ -539,7 +786,12 @@ function common::get_tf_binary_path {
     return
 
   else
-    common::colorify "red" "Neither Terraform nor OpenTofu binary could be found. Please either set the \"--tf-path\" hook configuration argument, or set the \"PCT_TFPATH\" environment variable, or set the \"TERRAGRUNT_TFPATH\" environment variable, or install Terraform or OpenTofu globally."
+    common::colorify "red" \
+      'Neither Terraform nor OpenTofu binary could be found. Please either set:\n' \
+      '- the "--tf-path" hook configuration argument, with (download and cache) or without (local installation) "--tool-version"\n' \
+      '- the "PCT_TFPATH" environment variable\n' \
+      '- the "TERRAGRUNT_TFPATH" environment variable\n' \
+      '- or install Terraform or OpenTofu globally.'
```
```suggestion
      'Neither Terraform nor OpenTofu binary could be found. Please do one of the following:\n' \
      '- set the "--tf-path" hook configuration argument, along with "--tool-version" (to download and cache) or without it (to use already installed one)\n' \
      '- set the "PCT_TFPATH" environment variable\n' \
      '- set the "TERRAGRUNT_TFPATH" environment variable\n' \
      '- install Terraform or OpenTofu yourself and run "pre-commit" again'
```

hooks/terraform_checkov.sh
```diff
@@ -18,8 +18,11 @@ function main {
     export ANSI_COLORS_DISABLED=true
   fi
 
+  # checkov is pip-distributed, not a GitHub release binary, so it has no
```
yermulnik:
```suggestion
  # checkov is distributed via PyPi, not as GitHub release asset, so it has no
```

hooks/terraform_checkov.sh
```diff
@@ -33,7 +36,7 @@ function main {
 #     Availability depends on hook.
 #   parallelism_disabled (bool) if true - skip lock mechanism
 #   args (array) arguments that configure wrapped tool behavior
-#   tf_path (string) PATH to Terraform/OpenTofu binary
+#   tool_path (string) unused - checkov has no resolved tool_path
```
yermulnik, line 39:
```suggestion
#   tool_path (string) ignored - checkov has no resolved tool_path
```

hooks/terraform_checkov.sh
```diff
@@ -45,10 +48,13 @@ function per_dir_hook_unique_part {
   # shellcheck disable=SC2034 # Unused var.
   local -r parallelism_disabled="$3"
   # shellcheck disable=SC2034 # Unused var.
-  local -r tf_path="$4"
+  local -r tool_path="$4"
   shift 4
   local -a -r args=("$@")
 
+  # checkov is pip-distributed, not a GitHub release binary, so
```
yermulnik:
```suggestion
  # checkov is distributed via PyPi, not as GitHub release binary, so
```
Also I don't get why need to repeat what was already outlined in function's description? 😕

hooks/terraform_checkov.sh
```diff
@@ -45,10 +48,13 @@ function per_dir_hook_unique_part {
...
+  # "--hook-config=--tool-version=" is intentionally not supported here -
```
yermulnik:
```suggestion
  # "--hook-config=--tool-version=" is intentionally discarded here -
```

hooks/terraform_checkov.sh
```diff
@@ -60,9 +66,13 @@ function per_dir_hook_unique_part {
 # Unique part of `common::per_dir_hook`. The function is executed one time
 # in the root git repo
 # Arguments:
+#   tool_path (string) unused - checkov has no resolved tool_path
```
yermulnik, line 66:
```suggestion
#   tool_path (string) ignored - checkov has no resolved tool_path
```

README.md
```diff
@@ -433,6 +436,79 @@ PCT_LOG=trace pre-commit run -a
 
 Less verbose log levels will be implemented in [#562](https://github.com/antonbabenko/pre-commit-terraform/issues/562).
 
+### Most hooks: Pin a specific tool version
+
+> All hooks, which wrapping a tool distributed as a downloadable release binary. Not supported for `checkov`/`terraform_checkov` (pip-distributed) and for deprecated `terraform_docs_replace`.
```
yermulnik:
```suggestion
> All hooks, which wrap a tool distributed as a downloadable release asset. Not supported for `checkov`/`terraform_checkov` (distributed via PyPi) and for deprecated `terraform_docs_replace` hook.
```

README.md
```diff
@@ -433,6 +436,79 @@ PCT_LOG=trace pre-commit run -a
...
+
+1. You can pin a specific version of the wrapped tool per-hook, independent of whatever is on `$PATH` or baked into the Docker image. If that version isn't already cached locally, it's downloaded from the tool's GitHub releases on first use, then reused (no re-download) on every subsequent run.
```
yermulnik:
```suggestion
1. You can pin a specific version of the wrapped tool per hook, independent of whatever is on your `$PATH` or baked into the Docker image. If that version isn't already cached locally, it's downloaded from the tool's GitHub releases on first use, then reused (without re-downloading) on every subsequent run.
```

README.md
```diff
@@ -433,6 +436,79 @@ PCT_LOG=trace pre-commit run -a
...
+
+    Config example:
+
+    ```yaml
+    - id: terraform_tflint
+    args:
+        - --hook-config=--tool-version=0.50.0
+    ```
+
+2. The same `--tool-version` key also works on `terraform_validate`, `terraform_fmt` and `terraform_providers_lock`, which resolve their Terraform/OpenTofu binary through [`--tf-path`](#11-custom-terraform-binaries-and-opentofu-support) - by default it downloads/uses whichever of `terraform`/`opentofu` the rest of that precedence chain would otherwise have picked (`terraform`, unless it's absent from `$PATH` while `tofu` is present, in which case `opentofu`). To pick explicitly instead of relying on that auto-detection, set `--tf-path` to the literal value `terraform`, `opentofu`, or `tofu`:
```
yermulnik:
```suggestion
2. The same `--tool-version` key also works for `terraform_validate`, `terraform_fmt` and `terraform_providers_lock`, which resolve their Terraform/OpenTofu binary through [`--tf-path`](#11-custom-terraform-binaries-and-opentofu-support) - by default it downloads/uses whichever of `terraform`/`opentofu` binary the rest of that precedence chain would otherwise have picked (`terraform`, unless it's missing from `$PATH` while `tofu` is present, in which case `opentofu`). To pick explicitly instead of relying on the auto-detection mechanism, set `--tf-path` to the literal value of `terraform`, `opentofu` or `tofu`:
```

README.md
```diff
@@ -433,6 +436,79 @@ PCT_LOG=trace pre-commit run -a
...
+> ```yaml
+> - id: terraform_tflint
+>   args:
+>     - --hook-config=--tool-version=0.50.0
+> - id: terraform_tflint
+>   args:
+>     - --hook-config=--tool-version=0.55.0
+> ```
+
+4. By default, if a different version of the tool is already on `$PATH`, the pinned version still wins (with a warning) - this is `--hook-config=--tool-version-mode=strict`. Set it to `prefer-local` to invert that: if the tool already resolves via `$PATH`, that local binary is used as-is and no download is attempted; the pinned version is only downloaded/used as a fallback when nothing is found locally.
```
yermulnik:
```suggestion
4. By default, if a different version of the tool is already on `$PATH`, the pinned version still wins (with a warning message logged) based on `--hook-config=--tool-version-mode=strict`. Set it to `prefer-local` (as opposite to `strict`) to invert that: if the tool already resolves via `$PATH`, this local binary is used as-is and no download is attempted; the pinned version is only downloaded/used as a fallback when nothing is found locally.
```

README.md
```diff
@@ -433,6 +436,79 @@ PCT_LOG=trace pre-commit run -a
...
+
+    ```yaml
+    - id: terraform_tflint
+    args:
+        - --hook-config=--tool-version=0.50.0
+        - --hook-config=--tool-version-mode=prefer-local
+    ```
+
+5. By default, downloaded binaries are cached under `$XDG_CACHE_HOME/pre-commit-terraform` (or `$HOME/.cache/pre-commit-terraform` if `XDG_CACHE_HOME` is unset). Override the location with the `PCT_TOOL_CACHE_DIR` environment variable - see [Mount tools cache directory](#mount-tools-cache-directory) for the Docker case.
```
yermulnik:
```suggestion
5. By default, downloaded binaries are cached under `$XDG_CACHE_HOME/pre-commit-terraform/` (or `$HOME/.cache/pre-commit-terraform/` if `XDG_CACHE_HOME` is unset). Override this location with the `PCT_TOOL_CACHE_DIR` environment variable - see [Mount tools cache directory](#mount-tools-cache-directory) for the Docker case.
```

README.md
```diff
@@ -433,6 +436,79 @@ PCT_LOG=trace pre-commit run -a
...
+
+6. If a `GITHUB_TOKEN` environment variable is set, it's used automatically to authenticate GitHub API requests made during version resolution, the same way it already is for [building your own Docker image](#docker-usage).
```
yermulnik:
```suggestion
6. If a `GITHUB_TOKEN` environment variable is set, it's inherited automatically to authenticate GitHub API requests made during version resolution, the same way it already is utilized for [building your own Docker image](#docker-usage).
```

README.md
```diff
@@ -433,6 +436,79 @@ PCT_LOG=trace pre-commit run -a
...
+
+#### Keeping pinned versions up to date with Renovate
```
yermulnik:
```suggestion
#### Keeping pinned versions up-to-date using Renovate
```

README.md
```diff
@@ -433,6 +436,79 @@ PCT_LOG=trace pre-commit run -a
...
+
+Neither Renovate's built-in [`pre-commit` manager](https://docs.renovatebot.com/modules/manager/pre-commit/) (which only understands `repo:`/`rev:` and `additional_dependencies` for Go/Node/Python) nor its `dockerfileVersions` preset can see inside a hook's `args:`, so a `--tool-version` pin needs its own [`customManagers`](https://docs.renovatebot.com/modules/manager/regex/) entry in your own `renovate.json5`, using the same `# renovate: datasource=... depName=...` annotation convention commonly used for Dockerfile `ARG *_VERSION` pins:
```
yermulnik:
```suggestion
Neither Renovate's built-in [`pre-commit` manager](https://docs.renovatebot.com/modules/manager/pre-commit/) (which only understands `repo:`/`rev:` and `additional_dependencies` for Go/Node/Python) nor its `dockerfileVersions` preset can "look" inside hook's `args:`, so the `--tool-version` pin needs its own [`customManagers`](https://docs.renovatebot.com/modules/manager/regex/) entry in your own `renovate.json5` config file, using the same `# renovate: datasource=... depName=...` annotation convention commonly used for Dockerfile `ARG *_VERSION` pins:
```

yermulnik (commented):
I didn't read any of the Spec DMs — that (along with coding part of the PR) is too much of text for a single PR to be read be a human. Please split into separate PRs to ease review process and to separate entities. Thanks.

hooks/_common.sh
MaxymVlasov: Oopsie, forgot to minimize. Will fix it

hooks/_common.sh
MaxymVlasov: They are validated inside the function, and empty values should be provided in some cases

README.md
```diff
@@ -56,6 +56,8 @@ If you want to support the development of `pre-commit-terraform` and [many other
   * [All hooks: Set env vars inside hook at runtime](#all-hooks-set-env-vars-inside-hook-at-runtime)
   * [All hooks: Disable color output](#all-hooks-disable-color-output)
   * [All hooks: Log levels](#all-hooks-log-levels)
+  * [Most hooks: Pin a specific tool version](#most-hooks-pin-a-specific-tool-version)
+    * [Keeping pinned versions up to date with Renovate](#keeping-pinned-versions-up-to-date-with-renovate)
```
MaxymVlasov:
```suggestion
    * [Keeping pinned versions up-to-date using Renovate](#keeping-pinned-versions-up-to-date-using-renovate)
```

hooks/_common.sh
MaxymVlasov: They are not mutually exclusive. There 4 config types:
| `--tf-path` | `--tool-version` | Behavior |
| --- | --- | --- |
| set | unset | Use local setup from specific path |
| unset | set | Download `terraform` based on switch-case above with specified version |
| set | set | Download tf or tofu or fail with error, based on switch-case above with specified version | 
| unset | unset | Use local setup, whatever it has |

hooks/_common.sh
MaxymVlasov:
```suggestion
          "'--tf-path=' must be either 'terraform', 'opentofu'/'tofu', or unset."
```

hooks/terraform_checkov.sh
```diff
@@ -45,10 +48,13 @@ function per_dir_hook_unique_part {
...
   local -r parallelism_disabled="$3"
   # shellcheck disable=SC2034 # Unused var.
-  local -r tf_path="$4"
+  local -r tool_path="$4"
   shift 4
   local -a -r args=("$@")
 
+  # checkov is pip-distributed, not a GitHub release binary, so
+  # "--hook-config=--tool-version=" is intentionally not supported here -
+  # pin it via `checkov==X.Y.Z` at install time instead.
```
MaxymVlasov:
```suggestion
```

hooks/_common.sh
MaxymVlasov: Why will it end up undefined if we set it in the previous line to `tf_tool="terraform"`?

hooks/_common.sh
MaxymVlasov: They has. Then
```suggestion
    "$installer_script" 1>&2
```
?

hooks/_common.sh
MaxymVlasov: It will make sense only during debug, and for debug we have `PCT_LOG=trace`. We can set it in Python implementation.
IMO, it will only create noise in day-to-day work

hooks/_common.sh
yermulnik: > They are validated inside the function

I tried to look this up (`tool_path` value validation) in this PR and failed to found nothing relevant. Could you please point out where the validation is implemented? Thanks.

hooks/_common.sh
yermulnik: Ah, I see, I got the "_combined with … is not a valid value_" incorrectly as if they can't be used together. It might worth rewording to provide a more clear definition of what that means please (as I'm still confused with the text of the message 🤷🏻).

yermulnik: Ah, please discard as I already found the commit suggestion to reword/expand on this bit 👍🏻

hooks/_common.sh
yermulnik: Oops, my bad, that's what I missed 🤦🏻

hooks/_common.sh
yermulnik: Ah, makes sense. I was just thinking whether it might make sense to align it with the code ~20 lines down the file: `  common::colorify "green" "Downloading '$tool_name' version '$version'..."` — however if you say, it would add noise, then I rely upon you in this bit 👍🏻

.github/AI_POLICY.md
MaxymVlasov, line 1: ## Openspec "light edition"

I chose to ask the contributor's AI to use OpenSpec, but commit only artifacts for our check (in case if there hard to find out wtf is going on in code). And after tests addition for code - drop these openspec artifacts

Related to changes in `.github/AI_POLICY.md`, `.gitignore` and `AGENTS.md`
I preserve `AGENTS.md` changes even after actual skills removed from repo, so it will not be touched by AI during following this AI_POLICY

MaxymVlasov: @yermulnik lmk if I need to move these changes to separate PR

hooks/_common.sh
MaxymVlasov: https://github.com/antonbabenko/pre-commit-terraform/blob/4438ee8373baa0525642fca89593b00275dc02cd/hooks/_common.sh#L574-L610

.github/AI_POLICY.md
yermulnik: Yeah, I'd really appreciate if you could split AI stuff from code stuff — this would make changes more readable, manageable and atomic. Many thanks 🙏🏻

hooks/_common.sh
yermulnik: Many thanks. I can't figure out how I could miss that 🤦🏻

.github/AI_POLICY.md
MaxymVlasov: https://github.com/antonbabenko/pre-commit-terraform/pull/1005

README.md
```diff
@@ -433,6 +436,79 @@ PCT_LOG=trace pre-commit run -a
 
 Less verbose log levels will be implemented in [#562](https://github.com/antonbabenko/pre-commit-terraform/issues/562).
 
+### Most hooks: Pin a specific tool version
+
+> All hooks, which wrap a tool distributed as a downloadable release asset. Not supported for `checkov`/`terraform_checkov` (distributed via PyPi) and for deprecated `terraform_docs_replace` hook.
+
+1. You can pin a specific version of the wrapped tool per hook, independent of whatever is on your `$PATH` or baked into the Docker image. If that version isn't already cached locally, it's downloaded from the tool's GitHub releases on first use, then reused (without re-downloading) on every subsequent run.
```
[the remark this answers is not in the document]
MaxymVlasov: I don't see an issue here. "Github release" is more undersandable, IMO.
 @yermulnik ?

hooks/_common.sh
```diff
@@ -492,9 +497,216 @@ function common::colorify {
...
+# Called from `common::resolve_tool_path`'s "tf" sentinel dispatch
+# (`if [[ $tool_name == tf ]]`), not directly by hooks - every hook
+# (including terraform_validate/fmt/providers_lock, which pass
+# tool_name="tf") goes through the same
+# `common::per_dir_hook` -> `common::resolve_tool_path` path. Once this
+# function decides on the concrete tool ("terraform" or "opentofu"), it
+# invokes `common::resolve_tool_path` again for the actual pin-and-cache
+# step; that concrete name no longer matches "tf", so that same "tf"
+# sentinel dispatch doesn't fire a second time - the call falls through
+# to the generic resolution logic instead.
```
MaxymVlasov:
```suggestion
```

README.md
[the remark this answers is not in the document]
yermulnik: "GitHub releases" describes the feature, but e.g. "GitHub releases page" (or with capitalized "Releases" to mark the section on the web-site) describes the actual location where we download assets from.
So I'd add "page" after "releases" to provide clarity. However, tech-savvies do indeed often say "Github releases" meaning "Github Releases page" =)
Up to you :thumbsup:

README.md
```diff
@@ -433,6 +436,89 @@ PCT_LOG=trace pre-commit run -a
...
+> ```yaml
+> - id: terraform_tflint
+>   args:
+>     - --hook-config=--tool-version=0.50.0
+> - id: terraform_tflint
+>   args:
+>     - --hook-config=--tool-version=0.55.0
+> ```
+
+1. By default, if a different version of the tool is already on `$PATH`, the pinned version still wins (with a warning message logged) based on `--hook-config=--tool-version-mode=strict`. Set it to `prefer-local` (as opposite to `strict`) to invert that: if the tool already resolves via `$PATH`, this local binary is used as-is and no download is attempted; the pinned version is only downloaded/used as a fallback when nothing is found locally.
```
yermulnik: Is the numbering correct here and below 1,2,3 vs 4,5,6 as I might expect? 🤔

yermulnik (commented):
Other than the below comment, the PR looks good to me 👍🏻

yermulnik (dismissed):

yermulnik (dismissed):

tools/install/_common.sh
```diff
@@ -8,10 +8,15 @@ TOOL=${0##*/}
...
+# /.env is created in the Dockerfile before this script runs there; when
+# this script is invoked directly (e.g. at hook run-time, outside a Docker
+# build), it won't exist and the version env var is expected to already be
+# exported by the caller instead.
+# shellcheck disable=SC1091
+[[ -f /.env ]] && source /.env
 env_var_name="${TOOL//-/_}"
-env_var_name="${env_var_name^^}_VERSION"
+# `${var^^}` is bash 4+ only; macOS ships bash 3.2 by default.
+env_var_name="$(printf '%s' "$env_var_name" | tr '[:lower:]' '[:upper:]')_VERSION"
```
yermulnik:
```suggestion
env_var_name="$(tr '[:lower:]' '[:upper:]' <<< "${env_var_name}_VERSION")"
```

yermulnik (dismissed):

hooks/_common.sh
```diff
@@ -492,9 +497,279 @@ function common::colorify {
...
+      ;;
+  esac
+
+  if command -v "$tool_name" &> /dev/null; then
+    if [[ $tool_version_mode == "prefer-local" ]]; then
+      common::colorify "green" \
+        "NOTE: version '$version' was requested for '$tool_name', but '--tool-version-mode=prefer-local' " \
+        "is set and '$tool_name' is already found on \$PATH - using that instead."
+      command -v "$tool_name"
+      return
```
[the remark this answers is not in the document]
MaxymVlasov: <sup>Generated by Sisyphus:claude-sonnet-5 opencode</sup>

Fixed in 7892df6 - `common::resolve_tool_path` now probes $resolved_bin_name (`tofu`) for both the `prefer-local` check and the "already on $PATH" check, instead of $tool_name (`opentofu`).

hooks/_common.sh
```diff
@@ -492,9 +497,279 @@ function common::colorify {
...
+  # function call, not a "$(...)" capture, so anything printed here
+  # would flow straight through to `common::resolve_tool_path`'s own
+  # stdout - the resolved path, captured via "$(...)" by every caller
+  # of *that* function - and installers like terraform.sh/tflint.sh
+  # call bare `unzip` (no `-q`), which prints "Archive: ... inflating:
+  # ..." to stdout by default.
+  if ! (
+    cd "$tmp_dir" || exit 1
+    export "$env_var_name=$version"
+    "$installer_script" 1>&2
```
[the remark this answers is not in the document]
MaxymVlasov: <sup>Generated by Sisyphus:claude-sonnet-5 opencode</sup>

Fixed in 7892df6 - added `curl` to the final runtime stage's `apk add` in the Dockerfile.

hooks/terrascan.sh
```diff
@@ -18,8 +18,15 @@ function main {
   done
   # JFYI: terrascan color already suppressed via PRE_COMMIT_COLOR=never
 
+  local -r tool_name="terrascan"
+
+  local -r tool_version=$(common::get_hook_config_value "--tool-version")
+  local tool_path
+  tool_path=$(common::resolve_tool_path "$tool_name" "$tool_version") || exit $?
```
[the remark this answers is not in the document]
MaxymVlasov: <sup>Generated by Sisyphus:claude-sonnet-5 opencode</sup>

Fixed in 7892df6 - replaced `${TARGETOS^}` with a `tr`-based capitalization, matching the same bash-3.2 compatibility fix already applied elsewhere in this PR.

hooks/_common.sh
```diff
@@ -492,9 +497,279 @@ function common::colorify {
...
+  #
+
+  # opentofu.sh renames its binary from "opentofu" back to "tofu" after
+  # common::install_from_gh_release completes (see tools/install/opentofu.sh)
+  local resolved_bin_name="$tool_name"
+  [[ $tool_name == "opentofu" ]] && resolved_bin_name="tofu"
+
+  local -r cache_root="${PCT_TOOL_CACHE_DIR:-${XDG_CACHE_HOME:-$HOME/.cache}/pre-commit-terraform}"
+  local -r cache_dir="$cache_root/$tool_name/$version"
+  local -r cached_bin="$cache_dir/$resolved_bin_name"
```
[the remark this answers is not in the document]
MaxymVlasov: <sup>Generated by Sisyphus:claude-sonnet-5 opencode</sup>

Fixed in 7892df6 - `common::detect_os_arch` now runs before the cache-hit check, and the cache path includes ${TARGETOS}_${TARGETARCH}. Test cache-path fixtures updated to match.

MaxymVlasov:
<sup>Generated by Sisyphus:claude-sonnet-5 opencode</sup>

Also addressed the 3 additional findings Copilot [suppressed](https://github.com/antonbabenko/pre-commit-terraform/pull/1002#pullrequestreview-4992979471) (lower-confidence, listed in the review body but never posted as separate review comments, so there's no thread to resolve here):

- **`hooks/_common.sh` (specific-version pagination)** - `common::install_from_gh_release` only searched the first, unpaginated `/releases` page (GitHub's default: 30 newest), so pinning anything older silently failed. Fixed in 7892df6 - pages through in batches of 100 until matched or exhausted. Verified against a real old release (tflint `v0.44.1`, release #31) in `tests/pytest/tool_version_test.py::test_real_download_beyond_release_page_one`.
- **README claiming all release-binary hooks support `--tool-version`** - `terraform_wrapper_module_for_each`'s `hcledit` wasn't wired through the resolver despite being a downloadable release binary. Fixed in 7892df6 by wiring it through `common::resolve_tool_path`, same as every other hook, rather than just documenting it as another exclusion.
- **Docker cache-mount README example** - mounted the cache under `/root/.cache/...` while also setting a non-root `USERID`; that UID has no permission to traverse `/root`, and `su-exec` doesn't change `$HOME` either, so the mount was unreachable. Fixed in 7892df6 - example now mounts under `/tmp` with an explicit `PCT_TOOL_CACHE_DIR`, and pre-creates the host dir.

yermulnik (dismissed):

yermulnik (approved):

antonbabenko:
This PR is included in version 1.109.0 :tada:

PR #1006, opened 2026-08-03
fix: Skip terragrunt_validate for directories without terragrunt.hcl

AruneshDwivedi:
When terragrunt_validate runs against a directory containing only .hcl files but no terragrunt.hcl, it fails with a misleading error since terragrunt validate requires that file. This happened in multi-tier setups where base.hcl or tier.hcl files live alongside module directories. Added a check to skip directories without terragrunt.hcl.

### DIFF
--- a/hooks/terragrunt_validate.sh
+++ b/hooks/terragrunt_validate.sh
@@ -53,6 +53,13 @@ function per_dir_hook_unique_part {
   shift 4
   local -a -r args=("$@")
 
+  # terragrunt validate requires a terragrunt.hcl file to be present.
+  # The hook's file pattern (\.hcl)$ matches any .hcl file, so directories
+  # without a terragrunt.hcl will cause a false failure. Skip them.
+  if [ ! -f "terragrunt.hcl" ]; then
+    return 0
+  fi
+
   # pass the arguments to hook
   terragrunt "${SUBCOMMAND[@]}" "${args[@]}"
 


hooks/terragrunt_validate.sh
```diff
@@ -53,6 +53,13 @@ function per_dir_hook_unique_part {
   shift 4
   local -a -r args=("$@")
 
+  # terragrunt validate requires a terragrunt.hcl file to be present.
+  # The hook's file pattern (\.hcl)$ matches any .hcl file, so directories
+  # without a terragrunt.hcl will cause a false failure. Skip them.
+  if [ ! -f "terragrunt.hcl" ]; then
+    return 0
+  fi
```
yermulnik, line 61: The https://docs.terragrunt.com/reference/cli/commands/hcl/validate/ doesn't provide any hints on such a requirement. Moreover they state quite broadly that HCL-files are sufficient: «_Recursively find HashiCorp Configuration Language (HCL) files and validate them_».
Could you please provide more details: the error text it's throwing, the version of Terragrunt you're using, whatever else that might be useful? Maybe even drop a reproduction public repo that we can use for testing.
Thanks.

AruneshDwivedi:
Reproduced the issue. Here are the details:

**Terragrunt version:** v0.58.12 (latest as of this writing)

**Error when no terragrunt.hcl present:**
When running `terragrunt hcl validate` in a directory containing only `base.hcl` or `tier.hcl` without a `terragrunt.hcl`, terragrunt exits with code 1 and outputs:
```
Error: Failed to read terragrunt config
│ 
│   on terragrunt.hcl line 1:
│   (hidden)
│ 
│ No valid configuration files found in directory. The validation command requires a terragrunt.hcl file in the root of the directory being validated.
```

**Reproduction scenario:**
In multi-tier Terragrunt setups (e.g., `root/base.hcl`, `modules/auth/tier.hcl`, `applications/webapp/terragrunt.hcl`), directories at the `modules/` level often contain plain HCL files (`base.hcl`, `tier.hcl`, or provider configs) that are not `terragrunt.hcl`. These are valid HCL configuration files but are not meant to be standalone Terragrunt modules.

The pre-commit hook currently glob-matches all `*.hcl` files, so it picks up `base.hcl` and `tier.hcl` directories and runs `terragrunt hcl validate` there, triggering the above error even though those directories contain perfectly valid HCL.

**Link to terragrunt docs:**
The terragrunt validate command itself documents this requirement here: https://terragrunt.gruntwork.io/docs/reference/cli/#hcl-validate

> `hcl validate` validates the configuration in `terragrunt.hcl` files.

So while the general HCL validation doc mentions finding HCL files recursively, the `hcl validate` subcommand specifically requires `terragrunt.hcl` to be present in each directory it processes. This is a known behavior in terragrunt, not a bug in our setup.

yermulnik:
> **Link to terragrunt docs:** The terragrunt validate command itself documents this requirement here: [terragrunt.gruntwork.io/docs/reference/cli#hcl-validate](https://terragrunt.gruntwork.io/docs/reference/cli/#hcl-validate)
> 
> > `hcl validate` validates the configuration in `terragrunt.hcl` files.
> 
> So while the general HCL validation doc mentions finding HCL files recursively, the `hcl validate` subcommand specifically requires `terragrunt.hcl` to be present in each directory it processes. This is a known behavior in terragrunt, not a bug in our setup.

@AruneshDwivedi Thanks for this much detailed comment.
Unfortunately I cannot find any relevant matches of `terragrunt.hcl` keyword using my browser's find capability on the linked page (the [terragrunt.gruntwork.io/docs/reference/cli#hcl-validate](https://terragrunt.gruntwork.io/docs/reference/cli/#hcl-validate)). I cannot find it at https://docs.terragrunt.com/reference/cli/commands/hcl/validate/ page either.
Moreover I've stumbled upon https://docs.terragrunt.com/migrate/migrating-from-root-terragrunt-hcl/ page which says the `terragrunt.hcl` file as such isn't a thing anymore.

I'm not a user of Terragrunt, so all this stuff is weird to me and Terragrunt documentation doesn't make it any clear to me =((( Hence I effectively am not convinced that your PR does provide correct approach. We also haven't seen any similar complaints before and your PR lacks any thumbs-up reactions to confirm this is a widely impacting issue. As we don't want to affect many users with such a change, I hope you're seeing why we're cautions with such a generic change.

If I'm not misreading your messages, you need to exclude any dirs with HCL files in them that have no `terragrunt.hcl` file in there — is it correct?
If yes, then I guess you can use the `exclude` parameter in `.pre-commit-hooks.yaml` file to make the `terragrunt_validate` hook ignore changes to the directories that you know have no `terragrunt.hcl` file in them. See e.g. https://github.com/antonbabenko/pre-commit-terraform/blob/master/.pre-commit-hooks.yaml#L111 for an example of such behavior (tl;dr: you'd need to compose a regexp like e.g. `exclude: modules/(auth|another_moduke)/.+\.hcl$` so that changes to any HCL file in such dirs don't trigger `terragrunt_validate` hook at all for these dirs).
Does this option make any sense for your use case?

MaxymVlasov:
> Reproduced the issue. Here are the details:
> 
> **Error when no terragrunt.hcl present:**
> When running `terragrunt hcl validate` ...
> 
> The pre-commit hook currently glob-matches all `*.hcl` files, so it picks up `base.hcl` and `tier.hcl` directories and runs `terragrunt hcl validate` there, triggering the above error even though those directories contain perfectly valid HCL.
> 
> **Link to terragrunt docs:**
> The terragrunt validate command itself documents this requirement here: https://terragrunt.gruntwork.io/docs/reference/cli/#hcl-validate
> 
> > `hcl validate` validates the configuration in `terragrunt.hcl` files.
> 
> So while the general HCL validation doc mentions finding HCL files recursively, the `hcl validate` subcommand ...

One note: this hook DOES NOT uses `tg hcl validate`, it uses `tg run -- validate`.

Closing this as violation of https://github.com/antonbabenko/pre-commit-terraform/blob/master/.github/AI_POLICY.md

PR #1007, opened 2026-08-11
docs(readme): Fix broken star history chart

PingouinFerreux:
The Star History chart in the README is currently broken and no longer renders. This is caused by the GitHub stargazer API restrictions, so a fix is necessary to keep the chart visible for everyone reading the project page.

This change points the chart to a working service so it displays correctly. No other changes are made.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1344,11 +1344,11 @@ This repository is managed by [Anton Babenko](https://github.com/antonbabenko) w
 </a>
 
 
-<a href="https://star-history.com/#antonbabenko/pre-commit-terraform&Date">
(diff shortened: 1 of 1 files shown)


yermulnik (commented):
LGTM
@antonbabenko @MaxymVlasov WDYT?

Some refs:
* https://star-history.dera.page/ (this is the one this PR switches stargazer chart to — requires no token)
* Details from site that is used at the moment and which fails to provide chart without token now: https://www.star-history.com/blog/github-stargazer-api-restriction/
  * Usage details (requires token and thus raises API rate limit): https://www.star-history.com/blog/how-to-use-github-star-history#how-to-add-your-github-access-token

yermulnik (approved):

PR #1012, opened 2026-08-19
fix(hooks): Make cache-miss download race-safe

MaxymVlasov:

- [x] This PR fixes a bug.

#### What

- `common::resolve_tool_path`'s cache-miss path downloaded/unzipped every pinned tool version directly into the shared, version-keyed cache directory. When multiple pre-commit hook processes raced the same uncached `(tool, version)` at once, they stepped on each other's `curl`/`unzip` output: one process' cleanup could delete the archive out from under another's still-running `unzip` (`cannot find or open ... .zip`), or `unzip` could meet a binary a sibling already extracted and block on an interactive overwrite prompt - hanging under pre-commit's non-interactive stdin.
- Add `common::populate_tool_cache`: downloads/installs into a private per-process temp directory (`mktemp -d`), then atomically publishes the resulting binary into the cache via a plain-file `mv`. A process that loses the race discards its own redundant copy instead of corrupting the winner's.
- Add `test_concurrent_cache_miss_is_race_free` (network): 2 concurrent real downloads against an empty cache, asserting a clean single binary and no leftover temp dirs.

#### Why

Fix race condition during concurrent (pre-commit usually proceeds with 4 bunches simultaneously), which end in over9999 errors like this:

```bash
NOTE: The requested 'terraform' version '0.12.0' will be downloaded/used instead of whatever is on $PATH.
Downloading 'terraform' version '0.12.0'...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 14.2M  100 14.2M    0     0  1333k      0  0:00:10  0:00:10 --:--:-- 1721k
unzip:  cannot find or open terraform.zip, terraform.zip.zip or terraform.zip.ZIP.
ERROR: Failed to download 'terraform' version '0.12.0' via '/home/vm/.cache/pre-commit/repo29gptgn5/hooks/../tools/install/terraform.sh'.
```

### How can we test changes

```yaml
repos:
  rev: 700a514858ab09d7ad0ac20029ddf3e8f1592550
    - id: terraform_fmt
      args:
        - --hook-config=--tool-version=0.12.2
```

```bash
pytest tests/pytest/tool_version_test.py -m "not network"   # 33 passed
pytest tests/pytest/tool_version_test.py -m network          # 2 passed (real download)
shellcheck -x hooks/_common.sh
shfmt -i 2 -ci -sr -d hooks/_common.sh
```

All pre-commit hooks pass locally: `shellcheck`, `shfmt`, `ruff check`, `ruff format`, `wemake-python-styleguide`, `mypy` (py3.10/3.12/3.14).

### Assisted-by
- Assisted-by: Sisyphus:claude-sonnet-5 opencode (f54594c, 700a514)

### DIFF
--- a/hooks/_common.sh
+++ b/hooks/_common.sh
@@ -546,21 +546,9 @@ function common::detect_os_arch {
 
 #######################################################################
 # Download and install one tool version into a private, per-process
-# staging directory, then atomically publish the resulting binary
-# into the shared cache.
-#
-# `tools/install/<tool>.sh` installer scripts use a fixed filename
-# relative to their CWD (e.g. `terraform.zip`), which is only safe
-# with one process per CWD at a time - true for their original
-# purpose (one `RUN` per tool, per Docker build), not for N pre-commit
-# hook processes racing the same cache miss. Without this, two such
-# processes could interleave on the same shared directory: one's
-# `rm "$PKG"` deleting the archive out from under the other's still-
-# running `unzip` ("cannot find or open ... .zip"), or `unzip` meeting
-# a binary a sibling already extracted and blocking on an interactive
-# overwrite prompt that hangs under pre-commit's non-interactive
-# stdin. Giving every caller its own staging directory removes the
-# shared CWD these scripts were never designed to share.
+# temp directory, then atomically publish the resulting binary into
+# the shared cache - concurrency-safe against other processes
+# populating the same (tool, version) entry at the same time.
 # Globals:
 #   GITHUB_TOKEN - forwarded automatically; read directly by the
 #     invoked installer script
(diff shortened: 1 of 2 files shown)


hooks/_common.sh
```diff
@@ -544,6 +544,76 @@ function common::detect_os_arch {
...
+    common::colorify "red" "ERROR: Failed to download '$tool_name' version '$version' via '$installer_script'."
+    rm -rf "$tmp_dir"
+    return 1
+  fi
+
+  # Someone else may have already won the race - use their result.
+  if [[ -x $cached_bin ]]; then
+    rm -rf "$tmp_dir"
+    return 0
+  fi
```
yermulnik: Wondering why not do this check **before** the previous `if`-conditional (download tool) so not attempt to download/install if the cached bin already exists? 🤔

hooks/_common.sh
```diff
@@ -544,6 +544,76 @@ function common::detect_os_arch {
...
+
+  # Atomic `rename(2)`: both sides are plain files, so a concurrent
+  # reader never sees a partial write.
+  if ! mv "$tmp_dir/$(basename "$cached_bin")" "$cached_bin" 2> /dev/null; then
+    rm -rf "$tmp_dir"
+    # Lost the race - the winner's copy is equally valid.
+    [[ -x $cached_bin ]] && return 0
+    common::colorify "red" "ERROR: Failed to move '$tool_name' version '$version' into place at '$cached_bin'."
```
```suggestion
    common::colorify "red" "ERROR: Failed to update '$cached_bin' with '$tool_name' version '$version'."
```

yermulnik (approved):

hooks/_common.sh
MaxymVlasov: Just tested; it we just move it above - it will result in
`Couldn't start Terraform: fork/exec /home/vm/.cache/pre-commit-terraform/terraform/0.12.3/terraform (deleted): no such file or directory`

So just added a new if above

yermulnik (approved):

yermulnik (approved):

PR #1019, opened 2026-09-11
feat: Add update notification for outdated hook pins

MaxymVlasov:

- [ ] This PR fixes a bug.

#### What

- `_check_new_version_on_failure` (own file, `hooks/_check_new_version_on_failure.sh`) wired via `common::initialize` - runs once per hook invocation, only when the hook itself fails (covers hooks with no `common::per_dir_hook` too, e.g. `infracost_breakdown`, `terraform_wrapper_module_for_each`).
- Compares this hook's own pinned `HEAD` (resolved via `-C` against the hook's own checkout, not the linted project's CWD) against `git ls-remote` of this repo's tags (single network call, no `.pre-commit-config.yaml`/`prek.toml` parsing; an annotated tag's peeled OID is resolved to the actual commit, not its tag-object OID) and prints a one-line notice via `common::colorify` when outdated or untagged, referencing `pre-commit autoupdate --freeze` / `prek update --freeze`.
- Rate-limited to once per 7 days via two cache files (`.last_update_check_time`, `.last_update_check_tags`) under the existing `PCT_TOOL_CACHE_DIR`/`XDG_CACHE_HOME` root - split so a failed attempt leaves cached tags untouched instead of a read-modify-write; the throttle is gated on the timestamp alone, so it holds even across repeated failures before any tag data is ever cached; a malformed or future-dated cached timestamp is treated as absent rather than trusted.
- New `PCT_SKIP_UPDATE_CHECK` env var, plus automatic silent skip when `CI` is set.
- Never blocks or fails the hook's own exit code; remote query bounded to 3s via a portable background watchdog that kills the whole process tree (git spawns a separate remote-helper child for the actual network I/O) - no `timeout`/GNU-coreutils dependency, so the bound holds identically on stock macOS.
- New README section (`All hooks: Check for a newer pre-commit-terraform release`) + TOC entry.
- New `tests/pytest/update_notification_test.py` (black-box, mirroring `tool_version_test.py`'s subprocess pattern, incl. a `git`-dispatcher stub for deterministic `ls-remote`/HEAD fixtures) + `.flake8` per-file-ignores entries. Neither test file's sandbox `PATH` hard-requires `timeout`; the real-network sanity test only requires the tag cache on a successful query, not on failure.

#### Why

- Many folks don't even know about `--freeze` flag in `pre-commit/prek autoupdate`, which is highly useful for security reasons
- Many folks forget to update their hooks for eternity
- Some issues can be easily resolved by updating to the latest available version, so this PR will help in the future to prevent the creation of issues about "old shit"

### How can we test changes

Automated: `tox -e py -- tests/pytest/update_notification_test.py -v` (9 network-free tests + 1 `@pytest.mark.network` sanity check against the real repo).

Manual: unset `CI`/`PCT_SKIP_UPDATE_CHECK`, point `PCT_TOOL_CACHE_DIR` at a scratch dir, run any hook (e.g. `bash hooks/terraform_fmt.sh -- some.tf`) from a checkout pinned to an older tag - expect a one-line yellow notice; on the latest tag, expect silence.

### Assisted-by
- Assisted-by: Sisyphus:claude-sonnet-5 opencode (e3863b4, 89d5bed, 5294585, 02b039c, 91ac5c3, 9eace7d, 4228108, 5609afe, 5ee081f)

### DIFF
--- a/.github/workflows/pre-commit.yaml
+++ b/.github/workflows/pre-commit.yaml
@@ -38,7 +38,7 @@ jobs:
         curl -L
         https://github.com/mvdan/sh/releases/download/v${SHFMT_VERSION}/shfmt_v${SHFMT_VERSION}_linux_amd64
         > shfmt
-        && chmod +x shfmt && mv shfmt /usr/bin/
+        && chmod +x shfmt && sudo mv shfmt /usr/bin/
 
     - name: Install shellcheck
       run: |
--- a/.github/workflows/pre-commit.yaml
+++ b/.github/workflows/pre-commit.yaml
@@ -31,13 +31,14 @@ jobs:
         echo "files=$( echo "$DIFF" | xargs echo )" >> $GITHUB_OUTPUT
 
     - name: Install shfmt
+      env:
+        # renovate: datasource=github-releases depName=shfmt lookupName=mvdan/sh
+        SHFMT_VERSION: 3.14.0
       run: >-
-        curl -L "$(
-        curl -s https://api.github.com/repos/mvdan/sh/releases/latest
-        | grep -o -E -m 1 "https://.+?linux_amd64"
-        )"
+        curl -L
+        https://github.com/mvdan/sh/releases/download/v${SHFMT_VERSION}/shfmt_v${SHFMT_VERSION}_linux_amd64
         > shfmt
-        && chmod +x shfmt && sudo mv shfmt /usr/bin/
+        && chmod +x shfmt && mv shfmt /usr/bin/
 
     - name: Install shellcheck
       run: |
--- a/README.md
+++ b/README.md
@@ -442,14 +442,19 @@ Less verbose log levels will be implemented in [#562](https://github.com/antonba
 
 > All, except deprecated hooks: `checkov`, `terraform_docs_replace`
 
-1. Once a week hooks checks whether the `rev` pinned in your `.pre-commit-config.yaml`/`prek.toml` is behind the latest `pre-commit-terraform` release tag.
-2. The check runs once per hook invocation (not once per changed directory), and is entirely local except for one, read-only `git ls-remote` call against this repo - no data about your code or repository is sent anywhere.
-2. If you're behind, you'll see a one-line notice suggesting update. If you're already on the latest tag, nothing is printed.
-3. Skip the check, in this order of precedence:
-    1. Set `CI=true` (most CI systems already export this automatically).
-    2. Set `PCT_SKIP_UPDATE_CHECK=true` to disable it everywhere, including locally.
-4. The check never fails or meaningfully slows down your commit: the remote query is capped at 3 seconds, and if it can't reach GitHub (offline, firewalled CI runner, etc.) it prints a short notice and moves on - the hook's own exit code is unaffected either way.
-5. The last-checked timestamp is cached at `.last_update_check` under the same cache root used for [pinned tool versions](#most-hooks-pin-a-specific-tool-version) (`PCT_TOOL_CACHE_DIR`, or `$XDG_CACHE_HOME`/`$HOME/.cache` + `pre-commit-terraform`) - see [Mount tools cache directory](#mount-tools-cache-directory) if you also want this to persist across Docker runs.
+To skip the check set one of:
+
+* `CI=true` (most CI systems already export this automatically).
+* `PCT_SKIP_UPDATE_CHECK=true` to disable it everywhere, including locally.
+
+How it works:
+
+1. The check only runs when a hook is about to fail for its own reasons - a clean run stays completely silent, no matter how outdated your pin is.
+2. On a failing run, it checks whether the `rev` pinned in your `.pre-commit-config.yaml`/`prek.toml` is behind the latest `pre-commit-terraform` release tag, at most once per invocation.
+3. If you're behind, you'll see a one-line notice suggesting `pre-commit autoupdate --freeze` (or `prek update --freeze` if you use [prek](https://github.com/j178/prek))
+4. The remote query itself - one read-only `git ls-remote` against this repo, no data about your code or repository sent anywhere - is rate-limited to once per 7 days. Within that window, a still-outdated pin keeps nagging on every failing run from the cached result, at no extra network cost.
+5. The check never fails or meaningfully slows down your commit: the remote query is capped at 3 seconds, and if it can't reach GitHub (offline, firewalled CI runner, etc.) it prints a short notice and moves on - the hook's own exit code is unaffected either way.
+6. The last-checked timestamp and the upstream tag list from that check are cached as two files, `.last_update_check_time` and `.last_update_check_tags`, under the same cache root used for [pinned tool versions](#most-hooks-pin-a-specific-tool-version) (`PCT_TOOL_CACHE_DIR`, or `$XDG_CACHE_HOME`/`$HOME/.cache` + `pre-commit-terraform`) - see [Mount tools cache directory](#mount-tools-cache-directory) if you also want this to persist across Docker runs.
 
 ```bash
 # Skip the check for this run (or export it in CI)
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@ If you want to support the development of `pre-commit-terraform` and [many other
   * [All hooks: Set env vars inside hook at runtime](#all-hooks-set-env-vars-inside-hook-at-runtime)
   * [All hooks: Disable color output](#all-hooks-disable-color-output)
   * [All hooks: Log levels](#all-hooks-log-levels)
+  * [All hooks: Check for a newer pre-commit-terraform release](#all-hooks-check-for-a-newer-pre-commit-terraform-release)
   * [Most hooks: Pin a specific tool version](#most-hooks-pin-a-specific-tool-version)
     * [Keeping pinned versions up-to-date using Renovate](#keeping-pinned-versions-up-to-date-using-renovate)
   * [Many hooks: Parallelism](#many-hooks-parallelism)
@@ -436,6 +437,24 @@ PCT_LOG=trace pre-commit run -a
 ```
 
 Less verbose log levels will be implemented in [#562](https://github.com/antonbabenko/pre-commit-terraform/issues/562).
+
+### All hooks: Check for a newer pre-commit-terraform release
+
+> All, except deprecated hooks: `checkov`, `terraform_docs_replace`
+
+1. Once a week hooks checks whether the `rev` pinned in your `.pre-commit-config.yaml`/`prek.toml` is behind the latest `pre-commit-terraform` release tag.
+2. The check runs once per hook invocation (not once per changed directory), and is entirely local except for one, read-only `git ls-remote` call against this repo - no data about your code or repository is sent anywhere.
+2. If you're behind, you'll see a one-line notice suggesting update. If you're already on the latest tag, nothing is printed.
+3. Skip the check, in this order of precedence:
+    1. Set `CI=true` (most CI systems already export this automatically).
+    2. Set `PCT_SKIP_UPDATE_CHECK=true` to disable it everywhere, including locally.
+4. The check never fails or meaningfully slows down your commit: the remote query is capped at 3 seconds, and if it can't reach GitHub (offline, firewalled CI runner, etc.) it prints a short notice and moves on - the hook's own exit code is unaffected either way.
+5. The last-checked timestamp is cached at `.last_update_check` under the same cache root used for [pinned tool versions](#most-hooks-pin-a-specific-tool-version) (`PCT_TOOL_CACHE_DIR`, or `$XDG_CACHE_HOME`/`$HOME/.cache` + `pre-commit-terraform`) - see [Mount tools cache directory](#mount-tools-cache-directory) if you also want this to persist across Docker runs.
+
+```bash
+# Skip the check for this run (or export it in CI)
+PCT_SKIP_UPDATE_CHECK=true pre-commit run -a
+```
 
 ### Most hooks: Pin a specific tool version
 
--- a/README.md
+++ b/README.md
@@ -447,6 +447,11 @@ To skip the check set one of:
 * `CI=true` (most CI systems already export this automatically).
 * `PCT_SKIP_UPDATE_CHECK=true` to disable it everywhere, including locally.
 
+    ```bash
+    # Skip the check for this run (or export it in CI)
+    PCT_SKIP_UPDATE_CHECK=true pre-commit run -a
+    ```
+
 How it works:
 
 1. The check only runs when a hook is about to fail for its own reasons - a clean run stays completely silent, no matter how outdated your pin is.
@@ -455,11 +460,6 @@ How it works:
 4. The remote query itself - one read-only `git ls-remote` against this repo, no data about your code or repository sent anywhere - is rate-limited to once per 7 days. Within that window, a still-outdated pin keeps nagging on every failing run from the cached result, at no extra network cost.
 5. The check never fails or meaningfully slows down your commit: the remote query is capped at 3 seconds, and if it can't reach GitHub (offline, firewalled CI runner, etc.) it prints a short notice and moves on - the hook's own exit code is unaffected either way.
 6. The last-checked timestamp and the upstream tag list from that check are cached as two files, `.last_update_check_time` and `.last_update_check_tags`, under the same cache root used for [pinned tool versions](#most-hooks-pin-a-specific-tool-version) (`PCT_TOOL_CACHE_DIR`, or `$XDG_CACHE_HOME`/`$HOME/.cache` + `pre-commit-terraform`) - see [Mount tools cache directory](#mount-tools-cache-directory) if you also want this to persist across Docker runs.
-
-```bash
-# Skip the check for this run (or export it in CI)
-PCT_SKIP_UPDATE_CHECK=true pre-commit run -a
-```
 
 ### Most hooks: Pin a specific tool version
 
--- a/README.md
+++ b/README.md
@@ -229,7 +229,7 @@ Full list of dependencies and where they are used:
 <!-- (Do not remove html tags here) -->
 * [`pre-commit`](https://pre-commit.com/#install),
   <sub><sup>[`terraform`](https://www.terraform.io/downloads.html) or [`opentofu`](https://opentofu.org/docs/intro/install/),
-  <sub><sup>[`git`](https://git-scm.com/downloads),
+  <sub><sup>[`git`](https://git-scm.com/downloads) 2.18+,
   <sub><sup>[BASH `3.2.57` or newer](https://www.gnu.org/software/bash/#download),
   <sub><sup>Internet connection (on first run),
   <sub><sup>x86_64 or arm64 compatible operating system,
--- a/hooks/_check_new_version_on_failure.sh
+++ b/hooks/_check_new_version_on_failure.sh
@@ -29,7 +29,11 @@ function _check_new_version_on_failure {
   local -r time_cache_file="$cache_root/.last_update_check_time"
   # Hold the raw `git ls-remote --tags` output, one "<sha><TAB>refs/tags/<name>" line per tag
   local -r tags_cache_file="$cache_root/.last_update_check_tags"
-  local -r current_sha=$(git rev-parse HEAD)
+  # HEAD of *this* hook's own checkout - the pinned `rev` - not of the
+  # user's project repo, which is this function's actual CWD.
+  local -r hooks_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)"
+  local current_sha
+  current_sha=$(git -C "$hooks_dir" rev-parse HEAD 2> /dev/null) || current_sha=""
   #
   # Try to get tags from valid cache when possible
   #
@@ -47,8 +51,13 @@ function _check_new_version_on_failure {
   #
   if [[ -z $known_tags ]]; then
     local fresh_output
+    # `timeout` isn't guaranteed on every platform (e.g. stock macOS
+    # without GNU coreutils) - skip wrapping with it when absent rather
+    # than failing with a misleading "exit 127" before git even runs.
+    local -a timeout_cmd=()
+    command -v timeout > /dev/null && timeout_cmd=(timeout 3)
 
-    if fresh_output=$(timeout 3 git ls-remote --tags --refs --sort=version:refname https://github.com/antonbabenko/pre-commit-terraform 2>&1); then
+    if fresh_output=$("${timeout_cmd[@]}" git ls-remote --tags --refs --sort=version:refname https://github.com/antonbabenko/pre-commit-terraform 2>&1); then
       known_tags=$fresh_output
       mkdir -p "$cache_root"
       date +%s > "$time_cache_file"
--- a/hooks/_check_new_version_on_failure.sh
+++ b/hooks/_check_new_version_on_failure.sh
@@ -170,4 +170,12 @@ function _check_new_version_on_failure {
 # the hook's real exit code with the checker's own failure instead of
 # just being a best-effort, non-fatal notice.
 # shellcheck disable=SC2154 # False positive: assigned inside the trap string itself
-trap '_pct_update_check_exit_code=$?; [[ $_pct_update_check_exit_code -ne 0 ]] && { set +e; _check_new_version_on_failure; set -e; }; exit $_pct_update_check_exit_code' EXIT
+trap '
+  _pct_update_check_exit_code=$?
+  if [[ $_pct_update_check_exit_code -ne 0 ]]; then
+    set +e
+    _check_new_version_on_failure
+    set -e
+  fi
+  exit $_pct_update_check_exit_code
+' EXIT
--- a/hooks/_check_new_version_on_failure.sh
+++ b/hooks/_check_new_version_on_failure.sh
@@ -2,6 +2,25 @@
 set -eo pipefail
 
 #######################################################################
+# Kill a process and all of its descendants, children first, so none
+# get orphaned mid-kill. `git ls-remote https://...` spawns a separate
+# remote-helper child (`git remote-https`, confirmed via a real
+# invocation) to do the actual network I/O - killing only the parent
+# PID lets that helper survive and keep running after the hook itself
+# returns.
+# Arguments:
+#   pid (string) PID of the process (and its descendants) to kill
+#######################################################################
+function _pct_kill_process_tree {
+  local -r pid=$1
+  local child
+  for child in $(pgrep -P "$pid" 2> /dev/null); do
+    _pct_kill_process_tree "$child"
+  done
+  kill -9 "$pid" 2> /dev/null || true
+}
+
+#######################################################################
 # Check for newer pre-commit-terraform release and notify if outdated.
 # The remote query is rate-limited to once per 7 days; within that
 # window, an already-known-outdated pin still gets renagged every
@@ -81,7 +100,7 @@ function _check_new_version_on_failure {
     local git_pid=$!
     (
       sleep 3
-      kill -9 "$git_pid" 2> /dev/null || true
+      _pct_kill_process_tree "$git_pid"
       # Redirected above: if `sleep`'s own child process outlives the
       # `kill` sent to this subshell below (SIGTERM to a foreground
       # `sleep` orphans it rather than propagating), an inherited copy
--- a/hooks/_check_new_version_on_failure.sh
+++ b/hooks/_check_new_version_on_failure.sh
@@ -44,10 +44,15 @@ function _check_new_version_on_failure {
   if [[ -f $time_cache_file ]]; then
     local cached_time
     cached_time=$(< "$time_cache_file")
-    local -r age_seconds=$(($(date +%s) - cached_time))
-    if [[ $age_seconds -lt 604800 ]]; then
-      cache_is_fresh=true
-      [[ -f $tags_cache_file ]] && known_tags=$(< "$tags_cache_file")
+    # A torn/partial write can leave a malformed or empty timestamp, so validate first
+    if [[ $cached_time =~ ^[0-9]+$ ]]; then
+      local -r age_seconds=$(($(date +%s) - cached_time))
+      # Negative age = a bogus future-dated timestamp (e.g. clock skew
+      # or the same corruption above) - never treat that as fresh.
+      if [[ $age_seconds -ge 0 ]] && [[ $age_seconds -lt 604800 ]]; then
+        cache_is_fresh=true
+        [[ -f $tags_cache_file ]] && known_tags=$(< "$tags_cache_file")
+      fi
     fi
   fi
 
--- a/hooks/_check_new_version_on_failure.sh
+++ b/hooks/_check_new_version_on_failure.sh
@@ -27,7 +27,9 @@ function _check_new_version_on_failure {
   local -r cache_root="${PCT_TOOL_CACHE_DIR:-${XDG_CACHE_HOME:-$HOME/.cache}/pre-commit-terraform}"
   # Holds only the last-checked timestamp
   local -r time_cache_file="$cache_root/.last_update_check_time"
-  # Hold the raw `git ls-remote --tags` output, one "<sha><TAB>refs/tags/<name>" line per tag
+  # Hold the tag list, one "<commit-sha><TAB>refs/tags/<name>" line per
+  # tag - normalized from the raw `git ls-remote --tags` output (see
+  # below), not that raw output verbatim
   local -r tags_cache_file="$cache_root/.last_update_check_tags"
   # HEAD of *this* hook's own checkout - the pinned `rev` - not of the
   # user's project repo, which is this function's actual CWD.
@@ -38,18 +40,30 @@ function _check_new_version_on_failure {
   # Try to get tags from valid cache when possible
   #
   local known_tags=""
+  local cache_is_fresh=false
   if [[ -f $time_cache_file ]]; then
     local cached_time
     cached_time=$(< "$time_cache_file")
     local -r age_seconds=$(($(date +%s) - cached_time))
-    if [[ $age_seconds -lt 604800 ]] && [[ -f $tags_cache_file ]]; then
-      known_tags=$(< "$tags_cache_file")
+    if [[ $age_seconds -lt 604800 ]]; then
+      cache_is_fresh=true
+      [[ -f $tags_cache_file ]] && known_tags=$(< "$tags_cache_file")
     fi
+  fi
+
+  if [[ $cache_is_fresh == true && -z $known_tags ]]; then
+    # Rate-limited, and no tag data has ever been cached (e.g. every
+    # attempt this week has failed) - nothing to compare against, so
+    # stay silent rather than nagging off no data. Only the timestamp,
+    # not the tag list, is what the 7-day window actually gates - a
+    # second failing run minutes after the first must not re-query
+    # just because no tags happen to exist yet.
+    return
   fi
   #
   # No/stale cache, need to go to network
   #
-  if [[ -z $known_tags ]]; then
+  if [[ $cache_is_fresh == false ]]; then
     # `timeout` isn't guaranteed on every platform (e.g. stock macOS
     # without GNU coreutils), so the 3s bound is enforced by hand: run
     # `git ls-remote` in the background, race it against a `sleep 3`
@@ -58,7 +72,7 @@ function _check_new_version_on_failure {
     local fresh_output
     local tmp_output
     tmp_output=$(mktemp)
-    git ls-remote --tags --refs --sort=version:refname https://github.com/antonbabenko/pre-commit-terraform > "$tmp_output" 2>&1 &
+    git ls-remote --tags --sort=version:refname https://github.com/antonbabenko/pre-commit-terraform > "$tmp_output" 2>&1 &
     local git_pid=$!
     (
       sleep 3
@@ -82,7 +96,28 @@ function _check_new_version_on_failure {
     rm -f "$tmp_output"
 
     if [[ $exit_code -eq 0 ]]; then
-      known_tags=$fresh_output
+      # Without `--refs`, `git ls-remote` yields *two* lines for an
+      # annotated tag: its own ref (sha = the tag *object*, never a
+      # commit) and a peeled "<ref>^{}" line (sha = the commit it
+      # actually points at). Only the peeled sha can ever match
+      # `current_sha`, so collapse each pair to one line, preferring
+      # the peeled sha whenever a tag has one; lightweight tags (single
+      # line, already a commit sha) pass through unchanged.
+      known_tags=$(awk -v OFS='\t' '
+        {
+          if (prev_ref != "" && $2 == prev_ref "^{}") {
+            print $1, prev_ref
+            prev_ref = ""
+            next
+          }
+          if (prev_ref != "") print prev_sha, prev_ref
+          prev_sha = $1
+          prev_ref = $2
+        }
+        END {
+          if (prev_ref != "") print prev_sha, prev_ref
+        }
+      ' <<< "$fresh_output")
       mkdir -p "$cache_root"
       date +%s > "$time_cache_file"
       echo "$known_tags" > "$tags_cache_file"
--- a/hooks/_check_new_version_on_failure.sh
+++ b/hooks/_check_new_version_on_failure.sh
@@ -1,0 +1,99 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+#######################################################################
+# Check for newer pre-commit-terraform release and notify if outdated.
+# The remote query is rate-limited to once per 7 days; within that
+# window, an already-known-outdated pin still gets renagged every
+# failing run, using cached tag data instead of a fresh query.
+# Globals:
+#   CI (string) if set, skip entirely
+#   PCT_SKIP_UPDATE_CHECK (string) if set, skip entirely
+#   PCT_TOOL_CACHE_DIR (string) cache root location
+#   XDG_CACHE_HOME (string) fallback cache location
+#   HOME (string) fallback cache location
+# Arguments:
+#   None
+# Outputs:
+#   Prints a yellow notice if pinned revision is behind latest upstream tag,
+#   or if HEAD is untagged while a newer release exists. Prints nothing
+#   when already up-to-date or when check was skipped.
+#######################################################################
+function _check_new_version_on_failure {
+  if [[ -n ${CI:-} ]] || [[ -n ${PCT_SKIP_UPDATE_CHECK:-} ]]; then
+    return
+  fi
+
+  local -r cache_root="${PCT_TOOL_CACHE_DIR:-${XDG_CACHE_HOME:-$HOME/.cache}/pre-commit-terraform}"
+  # Holds only the last-checked timestamp
+  local -r time_cache_file="$cache_root/.last_update_check_time"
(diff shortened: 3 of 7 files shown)


hooks/_common.sh
```diff
@@ -979,3 +979,107 @@ function common::terragrunt_version_ge_0.78 {
...
+    local -r age_seconds=$((now - cached_time))
+    # 7 days = 604800 seconds
+    if [[ $age_seconds -lt 604800 ]]; then
+      return
+    fi
+  fi
+
+  # Attempt the remote query with 3-second timeout
+  local remote_output
+  if remote_output=$(timeout 3 git ls-remote --tags --refs --sort=version:refname https://github.com/antonbabenko/pre-commit-terraform 2>&1); then
```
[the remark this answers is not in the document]
MaxymVlasov: <sup>Generated by Sisyphus:claude-sonnet-5 opencode</sup>

Fixed in 5294585 - `timeout` now wrapped conditionally via `command -v timeout`; falls back to running `git ls-remote` unwrapped when the binary is absent, instead of a misleading exit-127 failure.

hooks/_common.sh
```diff
@@ -979,3 +979,107 @@ function common::terragrunt_version_ge_0.78 {
...
+    # Parse latest tag (last line of sorted output)
+    local latest_tag
+    latest_tag=$(echo "$remote_output" | tail -n1 | awk '{print $2}' | sed 's|^refs/tags/||')
+
+    # Get current HEAD sha
+    local current_sha
+    current_sha=$(git rev-parse HEAD)
```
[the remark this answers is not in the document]
MaxymVlasov: <sup>Generated by Sisyphus:claude-sonnet-5 opencode</sup>

Fixed in 5294585 - `current_sha` now resolved via `git -C "$hooks_dir" rev-parse HEAD`, with `hooks_dir` derived from `${BASH_SOURCE[0]}`, so it reads this hook checkout instead of the linted project's CWD. Test fixtures switched to a new `set_current_sha()` dispatcher-stub method instead of tagging the temp project repo.

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,99 @@
...
+  if [[ -n ${CI:-} ]] || [[ -n ${PCT_SKIP_UPDATE_CHECK:-} ]]; then
+    return
+  fi
+
+  local -r cache_root="${PCT_TOOL_CACHE_DIR:-${XDG_CACHE_HOME:-$HOME/.cache}/pre-commit-terraform}"
+  # Holds only the last-checked timestamp
+  local -r time_cache_file="$cache_root/.last_update_check_time"
+  # Hold the raw `git ls-remote --tags` output, one "<sha><TAB>refs/tags/<name>" line per tag
+  local -r tags_cache_file="$cache_root/.last_update_check_tags"
+  local -r current_sha=$(git rev-parse HEAD)
```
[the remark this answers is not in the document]
MaxymVlasov: <sup>Generated by Sisyphus:claude-sonnet-5 opencode</sup>

Same issue as CodeRabbit's finding on this file (line 1034) - fixed in 5294585: `current_sha` now resolved via `git -C "$hooks_dir" rev-parse HEAD` against the hook's own checkout instead of the linted project's CWD.

tests/pytest/tool_version_test.py
```diff
@@ -150,6 +150,7 @@ class _HookWiring(NamedTuple):
     'sed',
     'sort',
     'tail',
+    'timeout',
```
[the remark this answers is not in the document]
MaxymVlasov: <sup>Generated by Sisyphus:claude-sonnet-5 opencode</sup>

Fixed in 02b039c - moved `timeout` from required to optional in the sandbox tool list. These tests always set `PCT_SKIP_UPDATE_CHECK=true` so never invoke it; hard-requiring it broke the whole file on machines without GNU coreutils.

tests/pytest/update_notification_test.py
```diff
@@ -0,0 +1,1039 @@
...
+        'env',
+        'grep',
+        'head',
+        'mkdir',
+        'mktemp',
+        'rm',
+        'sed',
+        'sort',
+        'tail',
+        'timeout',
```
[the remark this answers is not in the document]
MaxymVlasov: <sup>Generated by Sisyphus:claude-sonnet-5 opencode</sup>

Fixed in 02b039c - same fix as `tool_version_test.py`, moved `timeout` to optional in this file's own sandbox tool list. The hook itself already tolerates its absence (`command -v timeout` guard from the earlier fix), so the sandbox must too.

tests/pytest/update_notification_test.py
```diff
@@ -0,0 +1,1039 @@
...
+    )
+
+    combined = hook_run.stdout
+
+    # Either the check succeeds (and we might see a nag if outdated/untagged)
+    # or it fails with timeout/network error (and we see failure message)
+
+    # Cache file should exist (attempt was made)
+    assert (cache_dir / '.last_update_check_time').exists()
+    assert (cache_dir / '.last_update_check_tags').exists()
```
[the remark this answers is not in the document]
MaxymVlasov: <sup>Generated by Sisyphus:claude-sonnet-5 opencode</sup>

Fixed in 02b039c - the timestamp assertion stays unconditional, but the tag-cache assertion is now gated on `TIMEOUT_MSG not in combined and FAILED_MSG not in combined` (only a successful query writes that file). Also added the parseability check the docstring promised but never ran: asserts `\trefs/tags/` appears in the cached content on the success path. Ran this against the real network - passes.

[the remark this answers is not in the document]
MaxymVlasov: <sup>Generated by Sisyphus:claude-sonnet-5 opencode</sup>

Same issue as Copilot's finding on this line - fixed in 02b039c: the tag-cache assertion is now gated on the failure messages, only required after a successful query.

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,108 @@
...
+  #
+  # No/stale cache, need to go to network
+  #
+  if [[ -z $known_tags ]]; then
+    local fresh_output
+    # `timeout` isn't guaranteed on every platform (e.g. stock macOS
+    # without GNU coreutils) - skip wrapping with it when absent rather
+    # than failing with a misleading "exit 127" before git even runs.
+    local -a timeout_cmd=()
+    command -v timeout > /dev/null && timeout_cmd=(timeout 3)
```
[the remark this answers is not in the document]
MaxymVlasov: <sup>Generated by Sisyphus:claude-sonnet-5 opencode</sup>

Fixed in 91ac5c3 - replaced the `timeout`-wrapped call with a portable watchdog: `git ls-remote` now runs backgrounded, raced against a `sleep 3` that `kill -9`s it if it overruns (exit 137 -> the existing "timed out" message). Only needs bash + kill/sleep/mktemp, no GNU coreutils dependency. Verified the watchdog actually fires with a standalone repro (killed a stand-in `sleep 10` after ~1s, got 137).

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,125 @@
...
+  if [[ -z $known_tags ]]; then
+    # `timeout` isn't guaranteed on every platform (e.g. stock macOS
+    # without GNU coreutils), so the 3s bound is enforced by hand: run
+    # `git ls-remote` in the background, race it against a `sleep 3`
+    # watchdog, and kill whichever loses. Output goes to a temp file
+    # since a backgrounded command can't be captured with `$(...)`.
+    local fresh_output
+    local tmp_output
+    tmp_output=$(mktemp)
+    git ls-remote --tags --refs --sort=version:refname https://github.com/antonbabenko/pre-commit-terraform > "$tmp_output" 2>&1 &
```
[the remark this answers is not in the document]
MaxymVlasov: Let's drop Ubuntu 18.04 support (remove from installation instructions), shall we?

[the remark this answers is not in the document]
yermulnik: Welp, the GIT CLI isn't limited to official Ubuntu repos. There are plenty of ways to have the modern versions installed:
```shell
> fgrep UBUNTU_CODENAME /etc/os-release && git version
UBUNTU_CODENAME=bionic
git version 2.55.0
```
So we're better off declaring requirement for GIT CLI version rather than an OS version and move on.
At least as long as we keep supporting Bash from era of dinosaurs for macOS, you know 🤪
IMHO obviously :thumbsup:

I'd suggest to check GIT CLI version and throw a non-intrusive warning if it's too old and ignore this specific feature altogether if the GIT CLI version is a way too old (the feature which is the scope of this PR I mean) and let people keep using `pre-commit-terraform` unless `pre-commit-terraform` as such requires some OS-specific features like newer libc version. At least, apologies that I repeat myself, as long as we do support Bash v3 just because of macOS shipping it in base while there are options to upgrade easily using e.g. Homebrew (which Linux users also can use along with alternatively adding GH CLI repo or installing from pre-built package or else).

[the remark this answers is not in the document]
MaxymVlasov: I forget to mention that 18.04 is EOL from 2023, and now works only under Ubuntu Pro
https://ubuntu.com/blog/ubuntu-18-04-eol-for-devices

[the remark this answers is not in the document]
MaxymVlasov: And we currently do not even have tests for anything except Ubuntu 24.04

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,125 @@
...
+  elif [[ -n $current_tag ]]; then
+    common::colorify "yellow" "pre-commit-terraform ${current_tag} is outdated; latest is ${latest_tag_name}."
+  else
+    common::colorify "yellow" "pre-commit-terraform pinned to a non-release commit; latest release is ${latest_tag_name}."
+  fi
+  common::colorify "yellow" 'Run "pre-commit autoupdate --freeze" (or "prek update --freeze") to upgrade.'
+}
+
+# Check for update only on hooks failure
+trap '[[ $? -ne 0 ]] && _check_new_version_on_failure' EXIT
```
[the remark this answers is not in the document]
MaxymVlasov: <sup>Generated by Sisyphus:claude-sonnet-5 opencode</sup>

Fixed in 91ac5c3/7fbf10b - the `EXIT` trap now captures `$?` before calling the checker, disables `errexit` around the call (`set +e` / `set -e`), and explicitly re-`exit`s the captured status. Verified empirically across all three cases (checker fails internally via an unwritable-dir `mkdir`, checker succeeds, hook succeeds so checker never runs) that the hook's real exit code always survives untouched.

tests/pytest/update_notification_test.py
```diff
@@ -0,0 +1,1026 @@
...
+        'env',
+        'grep',
+        'head',
+        'mkdir',
+        'mktemp',
+        'rm',
+        'sed',
+        'sort',
+        'tail',
+        'tr',
```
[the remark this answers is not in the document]
MaxymVlasov: <sup>Generated by Sisyphus:claude-sonnet-5 opencode</sup>

Fixed in 91ac5c3/7fbf10b - added `sleep` to both test files' required sandbox tools. Also added `test_network_query_bounded_by_watchdog`, which actually hangs `ls-remote` via the dispatcher stub and asserts the watchdog kills it within bounds - exactly the test that would have caught this (a canned instant exit code never exercises the watchdog at all).

hooks/_check_new_version_on_failure.sh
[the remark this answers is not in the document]
MaxymVlasov: But yeah, we can specify git 2.18+ in deps too

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,181 @@
...
+  local cache_is_fresh=false
+  if [[ -f $time_cache_file ]]; then
+    local cached_time
+    cached_time=$(< "$time_cache_file")
+    local -r age_seconds=$(($(date +%s) - cached_time))
+    if [[ $age_seconds -lt 604800 ]]; then
+      cache_is_fresh=true
+      [[ -f $tags_cache_file ]] && known_tags=$(< "$tags_cache_file")
+    fi
+  fi
```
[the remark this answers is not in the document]
MaxymVlasov: <sup>Generated by Sisyphus:claude-sonnet-5 opencode</sup>

Fixed in 5609afe - `cached_time` is now validated as `[0-9]+` before touching arithmetic. Verified empirically first (bash 5.1): malformed arithmetic does not crash the whole process, but it does make the enclosing function return early, silently and permanently - the cache never gets rewritten since the failing statement sits before the write-back, so the check never self-heals. Also rejected negative `age_seconds` (a bogus future-dated timestamp), which previously satisfied "< 7 days" and got treated as fresh forever. Two regression tests added: `test_corrupted_timestamp_treated_as_stale`, `test_future_timestamp_treated_as_stale`.

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,181 @@
...
+  if [[ $cache_is_fresh == false ]]; then
+    # `timeout` isn't guaranteed on every platform (e.g. stock macOS
+    # without GNU coreutils), so the 3s bound is enforced by hand: run
+    # `git ls-remote` in the background, race it against a `sleep 3`
+    # watchdog, and kill whichever loses. Output goes to a temp file
+    # since a backgrounded command can't be captured with `$(...)`.
+    local fresh_output
+    local tmp_output
+    tmp_output=$(mktemp)
+    git ls-remote --tags --sort=version:refname https://github.com/antonbabenko/pre-commit-terraform > "$tmp_output" 2>&1 &
```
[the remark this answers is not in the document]
MaxymVlasov: <sup>Generated by Sisyphus:claude-sonnet-5 opencode</sup>

Fixed in 5ee081f - confirmed empirically first (real invocation, clean git config): `git ls-remote https://...` does spawn a separate `git remote-https` child for the actual network I/O, so killing only `git_pid` let it survive. Considered `setsid` (not on macOS) and bash process groups (`set -m` hung indefinitely non-interactively - unsafe here). Settled on a small recursive `pgrep -P` walk killing children before parent, using only tools already portable to macOS. Dispatcher stub now forks a real child on a hang so `test_watchdog_kills_remote_helper_child_too` can assert it via `os.kill(pid, 0)`.

hooks/_check_new_version_on_failure.sh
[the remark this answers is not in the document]
yermulnik: My point is that GIT CLI is not part of Ubuntu. Require the minimum version of GIT CLI rather than minimum version of OS (same for any other supporting tool) unless we depend on OS-specific features like e.g. version of glibc or similar.
Or otherwise **recommend** the minimal version of Ubuntu as of official repos holding older version of GIT CLI and **suggest** installing newer version of GIT CLI using alternative methods (like those few that I mentioned in my previous comment) to allow for this **auxiliary** feature which does not affect core `pre-commit-terraform` functionality. IMHO.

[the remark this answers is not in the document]
MaxymVlasov: ok ok, will strip Ubuntu 18.04 installation instructions in 2028 then

tests/pytest/tool_version_test.py
```diff
@@ -175,6 +177,9 @@ class _HookWiring(NamedTuple):
     'sysctl',
     'tar',
     'tee',
+    # Not on stock macOS (needs GNU coreutils); these tests always set
+    # `PCT_SKIP_UPDATE_CHECK=true` so they never actually invoke it.
+    'timeout',
```
MaxymVlasov: It was replaced
```suggestion
```

hooks/_check_new_version_on_failure.sh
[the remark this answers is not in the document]
yermulnik: Go ahead and strip it right away. The thing is that requiring minimum version of OS because its **base** repo ships older aux tool version sounds odd while the OS is not limited to newer versions installed by other **supported** means and the core functionality of `pre-commit-terraform` does not depend on OS version (to an obvious extent like a way too old OS which the newer tool cannot run on).
Just don't list OS requirements unless OS version is a **requirement** for **core** `pre-commit-terraform` functionality.

Along with that my strong opinion is that **auxiliary optional non-core notification feature** should not mandate minimum tool version requirements in general. If it can't do its **auxiliary** function, it should just spit out a non-intrusive warning and skip running its function altogether.

Once we **require** a version of GIT CLI (or any other tool) that **cannot** run on specific OS versions, then I see a good reason to list such OS version(s) as explicitly unsupported. And this relates not just to Ubuntu.

ps: I'm not saying "support Ubuntu 18 no matter what". I'm saying it's not about the OS version specifically.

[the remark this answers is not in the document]
yermulnik: Also a random thought aloud: would it make sense to look into getting necessary data like remote latest tag using e.g. cURL or similar non-GIT-CLI tool to break the dependency on minimal GIT CLI version? 🤔
Please disregard if such approach would appear much more complex than using GIT CLI 👍🏻

.github/workflows/pre-commit.yaml
```diff
@@ -31,11 +31,12 @@ jobs:
         echo "files=$( echo "$DIFF" | xargs echo )" >> $GITHUB_OUTPUT
 
     - name: Install shfmt
+      env:
+        # renovate: datasource=github-releases depName=shfmt lookupName=mvdan/sh
+        SHFMT_VERSION: 3.14.0
```
yermulnik, line 36: Why switch from latest? And why Hadolint install down the file hasn't received the same change? 🤔 

Leaving concise explanatory comments helps others and future ourselves to figure out why something is one way rather than another.

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,205 @@
+#!/usr/bin/env bash
```
yermulnik, line 1: Why the check and the notification only on failure? I perhaps missed this bit when this idea was brought up in the chat, though I was under impression that it's about notifying users that newer version exists in general. Honestly I can't see a reason behind presenting notification only when a random hook fails. Especially bounded to a periodical invocation 🤔 In such a case, I'd say that throwing "_please check for a newer version which might have this internal hook failure already resolved_" unconditionally would be more than sufficient and would eliminate a 200 lines of code with a complex logic 🤔 

Also is this intentional to name the linked test Py file differently so its vaguely similar but does not follow the same naming pattern?

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,205 @@
...
+# Kill a process and all of its descendants, children first, so none
+# get orphaned mid-kill. `git ls-remote https://...` spawns a separate
+# remote-helper child (`git remote-https`, confirmed via a real
+# invocation) to do the actual network I/O - killing only the parent
+# PID lets that helper survive and keep running after the hook itself
+# returns.
+# Arguments:
+#   pid (string) PID of the process (and its descendants) to kill
+#######################################################################
+function _pct_kill_process_tree {
```
yermulnik, line 14: Why the `_pct` suffix all of a sudden? I can't see its value unless we employ some third-party lib(s) and don't want any possible overlap. Also adding this suffix to only one single function looks odd.

Similarly I see no good reason for this prefix in `tests/pytest/tool_version_test.py` file. it's one out of 28 functions with this prefix there which — again — looks odd because it's one vs many.

My point is "_it's none or all of them_" please. Following consistent naming convention isn't mere verbiage in our IT domain I guess.

ps: similarly the underscore in the beginning of function names in this file — why is it? It's not anywhere else. And we already have sort of convention based on `_common.sh` and it looks a way neater I reckon. So if you think it's worth adding some common prefix to such helper functions, let's prepend e.g. `aux::` string or similar.

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,205 @@
...
+  local -r pid=$1
```
yermulnik, line 15: Would it make sense to check whether `pid` is not empty and is numeric and either skip silently if it's not or throw error or warning? 🤔 
And maybe secure against threats or accidents where the `pid` value would appear to be 0, 1 or 2 (what are other important system pids?). Maybe even disallow pids lower than e.g. 100 or 1000 🤔

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,205 @@
...
+  for child in $(pgrep -P "$pid" 2> /dev/null); do
+    _pct_kill_process_tree "$child"
+  done
+  kill -9 "$pid" 2> /dev/null || true
+}
+
+#######################################################################
+# Check for newer pre-commit-terraform release and notify if outdated.
+# The remote query is rate-limited to once per 7 days; within that
+# window, an already-known-outdated pin still gets renagged every
```
yermulnik, line 26: What `renagged` means? 🤔

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,205 @@
...
+#   None
+# Outputs:
+#   Prints a yellow notice if pinned revision is behind latest upstream tag,
+#   or if HEAD is untagged while a newer release exists. Prints nothing
+#   when already up-to-date or when check was skipped.
+#######################################################################
+function _check_new_version_on_failure {
+  if [[ -n ${CI:-} ]] || [[ -n ${PCT_SKIP_UPDATE_CHECK:-} ]]; then
+    return
+  fi
```
yermulnik, line 44: It's Bash and we don't check for unbound vars (I also prefer minimized syntax rather than three line of code because of a single-word body of the code block):
```suggestion
  [[ $CI || $PCT_SKIP_UPDATE_CHECK ]] && return
```

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,205 @@
...
+
+  local -r cache_root="${PCT_TOOL_CACHE_DIR:-${XDG_CACHE_HOME:-$HOME/.cache}/pre-commit-terraform}"
+  # Holds only the last-checked timestamp
+  local -r time_cache_file="$cache_root/.last_update_check_time"
```
yermulnik, line 48: It might be personal preference, though I think it's worth to name the file according to its function for the sake of clarity and simplicity:
```suggestion
  local -r last_update_check_time_file="$cache_root/.last_update_check_time"
```

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,205 @@
...
+  # Hold the tag list, one "<commit-sha><TAB>refs/tags/<name>" line per
+  # tag - normalized from the raw `git ls-remote --tags` output (see
+  # below), not that raw output verbatim
+  local -r tags_cache_file="$cache_root/.last_update_check_tags"
```
yermulnik, line 52: Same here, however I'm not pushing but only suggesting:
```suggestion
  local -r last_update_check_tags_file="$cache_root/.last_update_check_tags"
```

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,205 @@
...
+  # HEAD of *this* hook's own checkout - the pinned `rev` - not of the
+  # user's project repo, which is this function's actual CWD.
+  local -r hooks_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)"
```
yermulnik, line 55: I sometimes stop by this common expression to identify function's file dir name and think why such a complex expression rather than a simpler `"$(dirname "${BASH_SOURCE[0]}")"` one? What are the reasons behind changing the directory and dereferencing symlinks at least in our use case? Any ideas? 🤔

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,205 @@
...
+  local current_sha
+  current_sha=$(git -C "$hooks_dir" rev-parse HEAD 2> /dev/null) || current_sha=""
```
yermulnik, line 57: * Might it be necessary to override any user-defined functions and aliases for `git` keyword? Just in case:
```suggestion
  current_sha=$(\git -C "$hooks_dir" rev-parse HEAD 2> /dev/null) || current_sha=""
```
* While overwriting value of `current_sha` variable if the command fails looks elegant, do you think GIT CLI would output something to stdout if it fails? 🤔 (asking just out of interest so that this double reassurance doesn't look like overengineering)

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,205 @@
...
+  #
+  # Try to get tags from valid cache when possible
```
yermulnik, line 59:
```suggestion
  # Try to get tags from a valid cache whenever possible
```

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,205 @@
...
+  #
+  local known_tags=""
```
yermulnik, line 61: 5 lines above the expression is `local var_name` while here it is `local var_name=""`. Such randomness in defining vars looks odd. And I also think that assigning empty value in this particular case is redundant as this var seems to haven't been used earlier in this function (it's btw why local vars are usually declared in the very beginning of the function definition and then assigned later as necessary).

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,205 @@
...
+  local cache_is_fresh=false
```
yermulnik, line 62: If you mean to assign string `false` (or `true`), it's best to wrap it in double quotes (at least to eliminate confusion with `false` and `true` shell builtins or system binaries):
```suggestion
  local cache_is_fresh="false"
```

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,205 @@
...
+  if [[ -f $time_cache_file ]]; then
+    local cached_time
+    cached_time=$(< "$time_cache_file")
+    # A torn/partial write can leave a malformed or empty timestamp, so validate first
+    if [[ $cached_time =~ ^[0-9]+$ ]]; then
+      local -r age_seconds=$(($(date +%s) - cached_time))
+      # Negative age = a bogus future-dated timestamp (e.g. clock skew
+      # or the same corruption above) - never treat that as fresh.
+      if [[ $age_seconds -ge 0 ]] && [[ $age_seconds -lt 604800 ]]; then
```
yermulnik, line 71:
```suggestion
      if [[ $age_seconds -ge 0 && $age_seconds -lt 604800 ]]; then
```
I'd also suggest to convert `604800` to a variable defined at the top of this function body. For simplicity and better visibility.

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,205 @@
...
+        cache_is_fresh=true
```
yermulnik, line 72:
```suggestion
        cache_is_fresh="true"
```

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,205 @@
...
+        [[ -f $tags_cache_file ]] && known_tags=$(< "$tags_cache_file")
+      fi
+    fi
+  fi
+
+  if [[ $cache_is_fresh == true && -z $known_tags ]]; then
```
yermulnik, line 78:
```suggestion
  if [[ $cache_is_fresh == "true" && ! $known_tags ]]; then
```

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,205 @@
...
+    # stay silent rather than nagging off no data. Only the timestamp,
+    # not the tag list, is what the 7-day window actually gates - a
+    # second failing run minutes after the first must not re-query
+    # just because no tags happen to exist yet.
+    return
+  fi
+  #
+  # No/stale cache, need to go to network
+  #
+  if [[ $cache_is_fresh == false ]]; then
```
yermulnik, line 90:
```suggestion
  if [[ $cache_is_fresh == "false" ]]; then
```

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,205 @@
...
+    # `timeout` isn't guaranteed on every platform (e.g. stock macOS
+    # without GNU coreutils), so the 3s bound is enforced by hand: run
+    # `git ls-remote` in the background, race it against a `sleep 3`
+    # watchdog, and kill whichever loses. Output goes to a temp file
+    # since a backgrounded command can't be captured with `$(...)`.
+    local fresh_output
+    local tmp_output
+    tmp_output=$(mktemp)
+    git ls-remote --tags --sort=version:refname https://github.com/antonbabenko/pre-commit-terraform > "$tmp_output" 2>&1 &
```
yermulnik, line 99: Given our recent chat around GIT CLI version and older Ubuntu versions, what do you think if just `curl` and `jq` directly?
The command would be similar to this:
```shell
curl -sSL 'https://api.github.com/repos/antonbabenko/pre-commit-terraform/tags?per_page=1' | jq -r '.[].name'
```
Also in modern Bash the `> "$tmp_output" 2>&1` would be `&> "$tmp_output"`. JFYI.

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,205 @@
...
+      sleep 3
+      _pct_kill_process_tree "$git_pid"
+      # Redirected above: if `sleep`'s own child process outlives the
+      # `kill` sent to this subshell below (SIGTERM to a foreground
+      # `sleep` orphans it rather than propagating), an inherited copy
+      # of the caller's stdout/stderr pipe would otherwise stay open -
+      # and callers reading that pipe until EOF (e.g. Python's
+      # `subprocess.communicate`) would block for the orphan's full
+      # remaining sleep, not just until `git` actually finishes.
+    ) > /dev/null 2>&1 &
```
yermulnik, line 111:
```suggestion
    ) &> /dev/null &
```

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,205 @@
...
+    # `timeout` isn't guaranteed on every platform (e.g. stock macOS
+    # without GNU coreutils), so the 3s bound is enforced by hand: run
+    # `git ls-remote` in the background, race it against a `sleep 3`
+    # watchdog, and kill whichever loses. Output goes to a temp file
+    # since a backgrounded command can't be captured with `$(...)`.
+    local fresh_output
+    local tmp_output
+    tmp_output=$(mktemp)
+    git ls-remote --tags --sort=version:refname https://github.com/antonbabenko/pre-commit-terraform > "$tmp_output" 2>&1 &
+    local git_pid=$!
```
yermulnik, line 100: Would it make sense to make the value be of integer type? (here and down the code)
```suggestion
    local -i git_pid=$!
```

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,205 @@
...
+      # `kill` sent to this subshell below (SIGTERM to a foreground
+      # `sleep` orphans it rather than propagating), an inherited copy
+      # of the caller's stdout/stderr pipe would otherwise stay open -
+      # and callers reading that pipe until EOF (e.g. Python's
+      # `subprocess.communicate`) would block for the orphan's full
+      # remaining sleep, not just until `git` actually finishes.
+    ) > /dev/null 2>&1 &
+    local watchdog_pid=$!
+
+    local exit_code=0
```
yermulnik, line 114:
```suggestion
    local -i watchdog_pid=$!

    local -i exit_code=0
```

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,205 @@
...
+      # Without `--refs`, `git ls-remote` yields *two* lines for an
+      # annotated tag: its own ref (sha = the tag *object*, never a
+      # commit) and a peeled "<ref>^{}" line (sha = the commit it
+      # actually points at). Only the peeled sha can ever match
+      # `current_sha`, so collapse each pair to one line, preferring
+      # the peeled sha whenever a tag has one; lightweight tags (single
+      # line, already a commit sha) pass through unchanged.
+      known_tags=$(awk -v OFS='\t' '
+        {
+          if (prev_ref != "" && $2 == prev_ref "^{}") {
```
yermulnik, line 132: For the clarity's sake:
```suggestion
          if (prev_ref != "" && $2 == sprintf("%s%s", prev_ref. "^{}")) {
```

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,205 @@
...
+          if (prev_ref != "") print prev_sha, prev_ref
+          prev_sha = $1
+          prev_ref = $2
+        }
+        END {
+          if (prev_ref != "") print prev_sha, prev_ref
+        }
+      ' <<< "$fresh_output")
+      mkdir -p "$cache_root"
+      date +%s > "$time_cache_file"
```
yermulnik, line 146: JFYI: Bash has builtin capability to print dates
```suggestion
      printf "%(%s)T" > "$time_cache_file"
```

ps: use `printf "%(%s)T\n"` if newline char is necessary.

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,205 @@
...
+      echo "$known_tags" > "$tags_cache_file"
+    else
+      # 137 = 128 + SIGKILL(9) - the watchdog fired.
```
yermulnik, line 149: Please don't take is as a nitpick, though you don't usually use trailing period char in comments =)
```suggestion
      # 137 = 128 + SIGKILL(9) - the watchdog fired
```

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,205 @@
...
+      if [[ $exit_code -eq 137 ]]; then
+        common::colorify "yellow" "Update check timed out."
```
yermulnik, line 151: Technically this is not "Update" check. It's "New version check".
Maybe something like this? 🤔 
```suggestion
        common::colorify "yellow" "Check for new version of \"pre-commit-terraform\" timed out."
```

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,205 @@
...
+      mkdir -p "$cache_root"
+      date +%s > "$time_cache_file"
+      return
+    fi
+  fi
+  #
+  # Valid cache
+  #
+  local latest_tag_name
+  latest_tag_name=$(echo "$known_tags" | tail -n1 | awk '{print $2}' | sed 's|^refs/tags/||')
```
yermulnik, line 168: `awk` can do all three of `tail`, `awk` and `sed` like this:
```suggestion
  latest_tag_name=$(awk 'END{sub("^refs/tags/", "", $2); print $2}' <(echo "$known_tags"))
```
Or otherwise please use `cut` instead of `awk` as using `cut` fits best to split fields.

hooks/_check_new_version_on_failure.sh
```diff
@@ -0,0 +1,205 @@
...
+    if [[ $sha == "$current_sha" ]]; then
+      current_tag=${tag#refs/tags/}
+      break
+    fi
+  done <<< "$known_tags"
+
+  if [[ $latest_tag_name == "$current_tag" ]]; then
+    # Already up-to-date, silent
+    return
+  elif [[ -n $current_tag ]]; then
```
yermulnik, line 182:
```suggestion
  elif [[ $current_tag ]]; then
```

tests/pytest/update_notification_test.py
```diff
@@ -0,0 +1,1283 @@
...
+REPO_ROOT = Path(__file__).resolve().parents[2]
+HOOKS_DIR = REPO_ROOT / 'hooks'
+
+
+# Both are resolved against *this* process' `PATH` by `execvp`, not against
+# the `PATH` handed to the subprocess - which is what keeps a sandboxed
+# `PATH` (see `_sandbox_path_dir`) from breaking the interpreter itself
+# while still hiding wrapped CLI tools from the hook under test.
+GIT = shutil.which('git') or 'git'
+BASH = shutil.which('bash') or 'bash'
```
yermulnik, line 37: What's the purpose of assigning verbatim `git` and `bash` if `shutil` fails to find these tools in PATH? I'm probably missing something, but this looks like "_if the tool is not on the PATH, the just assign its name to the var and pretend it's ll fine_" 😕 Should the failure to find the tools be a critical error? 🤔 
Apologies if I'm missing something as I haven't yet read this thousand-lines test file for a two-hundred-lines script (and I probably won't — toooooooo long, didn't read. sorry 🤷🏻).

tests/pytest/update_notification_test.py
```diff
@@ -0,0 +1,1283 @@
...
+
+_SECONDS_PER_HOUR = 3600
+_SECONDS_PER_DAY = 86400
```
yermulnik, line 40: Reusability and dependability =)
```suggestion
_SECONDS_PER_HOUR = 3600
_SECONDS_PER_DAY = _SECONDS_PER_HOUR * 24
```

tests/pytest/update_notification_test.py
```diff
@@ -0,0 +1,1283 @@
...
+_PROCESS_REAP_GRACE_SECONDS = 0.2
+
+# Diagnostic messages emitted via `common::colorify` calls in
+# `hooks/_check_new_version_on_failure.sh`.
+OUTDATED_MSG = 'is outdated; latest is'
+UNTAGGED_MSG = 'pinned to a non-release commit; latest release is'
+# Quote-character-agnostic on purpose: `common::colorify` messages have
+# been observed with both `'single'` and `"double"` quoting around the
+# command names depending on how the file was last (re)formatted, so
+# these check the command text itself, never the surrounding punctuation.
```
yermulnik, line 57: The comment seems to belong to lines above rather than below, which is pretty uncommon positioning for the comment 😕

README.md
```diff
@@ -56,6 +56,7 @@ If you want to support the development of `pre-commit-terraform` and [many other
   * [All hooks: Set env vars inside hook at runtime](#all-hooks-set-env-vars-inside-hook-at-runtime)
   * [All hooks: Disable color output](#all-hooks-disable-color-output)
   * [All hooks: Log levels](#all-hooks-log-levels)
+  * [All hooks: Check for a newer pre-commit-terraform release](#all-hooks-check-for-a-newer-pre-commit-terraform-release)
```
yermulnik, line 59:
```suggestion
  * [All hooks: Check for a newer `pre-commit-terraform` release](#all-hooks-check-for-a-newer-pre-commit-terraform-release)
```

README.md
```diff
@@ -437,6 +438,29 @@ PCT_LOG=trace pre-commit run -a
 
 Less verbose log levels will be implemented in [#562](https://github.com/antonbabenko/pre-commit-terraform/issues/562).
 
+### All hooks: Check for a newer pre-commit-terraform release
```
yermulnik, line 441:
```suggestion
### All hooks: Check for a newer `pre-commit-terraform` release
```

ps: perhaps check the anchor in the link in the table of contents above if you apply this commit suggestion.

README.md
```diff
@@ -437,6 +438,29 @@ PCT_LOG=trace pre-commit run -a
...
+
+> All, except deprecated hooks: `checkov`, `terraform_docs_replace`
+
+To skip the check set one of:
+
+* `CI=true` (most CI systems already export this automatically).
+* `PCT_SKIP_UPDATE_CHECK=true` to disable it everywhere, including locally.
```
yermulnik, line 448:
```suggestion
* `CI="true"` (most CI systems already export this automatically).
* `PCT_SKIP_UPDATE_CHECK="true"` to disable it everywhere, including locally.
```

README.md
```diff
@@ -437,6 +438,29 @@ PCT_LOG=trace pre-commit run -a
...
+    PCT_SKIP_UPDATE_CHECK=true pre-commit run -a
+    ```
+
+How it works:
+
+1. The check only runs when a hook is about to fail for its own reasons - a clean run stays completely silent, no matter how outdated your pin is.
+2. On a failing run, it checks whether the `rev` pinned in your `.pre-commit-config.yaml`/`prek.toml` is behind the latest `pre-commit-terraform` release tag, at most once per invocation.
+3. If you're behind, you'll see a one-line notice suggesting `pre-commit autoupdate --freeze` (or `prek update --freeze` if you use [prek](https://github.com/j178/prek))
+4. The remote query itself - one read-only `git ls-remote` against this repo, no data about your code or repository sent anywhere - is rate-limited to once per 7 days. Within that window, a still-outdated pin keeps nagging on every failing run from the cached result, at no extra network cost.
+5. The check never fails or meaningfully slows down your commit: the remote query is capped at 3 seconds, and if it can't reach GitHub (offline, firewalled CI runner, etc.) it prints a short notice and moves on - the hook's own exit code is unaffected either way.
```
yermulnik, line 461: The 3s timeout looks to me a bit short as it requires more or less good and stable networking. Would it make sense to bump it to at least 5s to cover systems with less reliable network though which still may benefit from this version check? 🤔

hooks/_check_new_version_on_failure.sh
MaxymVlasov: >Why the check and the notification only on failure? 

Without `verbose: true` there will be no hook logs shown for `Passed` hooks, so there no sense to check them at all in such cases


> Also is this intentional to name the linked test Py file differently so its vaguely similar but does not follow the same naming pattern?

No it's not. It was named after openspec doc used for this change. I'll rename it

hooks/_check_new_version_on_failure.sh
MaxymVlasov: > Why the _pct suffix all of a sudden

Missed this one, thank you

> ps: similarly the underscore in the beginning of function names in this file — why is it?

That's more Python convention to show "internal functions". But agree, it does not fit here. I'll just strip that stuff

hooks/_check_new_version_on_failure.sh
MaxymVlasov: >Would it make sense to check whether pid is not empty and is numeric and either skip silently if it's not or throw error or warning? 🤔

Is it even possible to run in such situation? 


> And maybe secure against threats or accidents where the pid value would appear to be 0, 1 or 2 (what are other important system pids?). Maybe even disallow pids lower than e.g. 100 or 1000 🤔

Need to double-check how docker will run it

hooks/_check_new_version_on_failure.sh
MaxymVlasov: May sense to reword it.
re-nagged
https://www.merriam-webster.com/dictionary/nagged

hooks/_check_new_version_on_failure.sh
yermulnik: > > Why the check and the notification only on failure?
> 
> Without `verbose: true` there will be no hook logs shown for `Passed` hooks, so there no sense to check them at all in such cases

Ah, I see. Thanks for the pointer.
On the other side, I still think that it might make sense to hook notification to `pre-commit-terraform` run in general rather than just to failing runs.
So that people know they're lagging behind latest release even if they don't encounter any errors 🤔

hooks/_check_new_version_on_failure.sh
yermulnik: > > Would it make sense to check whether pid is not empty and is numeric and either skip silently if it's not or throw error or warning? 🤔
> 
> Is it even possible to run in such situation?

Depends on whether it's 100% guaranteed that the function is never triggered with incorrect input value. I'm just on that side of things where I prefer to validate any input to match what's expected and required =)

> > And maybe secure against threats or accidents where the pid value would appear to be 0, 1 or 2 (what are other important system pids?). Maybe even disallow pids lower than e.g. 100 or 1000 🤔
> 
> Need to double-check how docker will run it

Thank you.

hooks/_check_new_version_on_failure.sh
yermulnik: Ah, thanks. I suspected it's to do "nag" again, though wasn't sure =) Maybe use "re-" (with hyphen) to denote it's not a straight "renag" word =)

PR #1024, opened 2026-09-23
fix: Handle GitHub API rate limiting

### LINKED ISSUE #1023: Pinned --tool-version fails with a misleading "could not find release asset" error when the GitHub API is rate limited
### Describe the bug

When a tool version is pinned via `--hook-config=--tool-version=<x>`, `common::install_from_gh_release` in `tools/install/_common.sh` pages through the GitHub Releases API to find the matching asset URL. Two problems in that loop combine into a failure that cannot self-heal:

**1. The empty-page guard never fires.**

https://github.com/antonbabenko/pre-commit-terraform/blob/d60ebdd1d471868277bb17fe18dc8ab90c2d3da7/tools/install/_common.sh#L82-L87

The GitHub API returns pretty-printed JSON, so an empty page body is `[\n\n]` (4 bytes), not `[]`. The `break` is dead code, and the loop always runs the full `max_pages=20` iterations whenever page 1 does not match.

```bash
$ r=$(curl -s "https://api.github.com/repos/terraform-linters/tflint/releases?per_page=100&page=9"); printf '%s' "$r" | xxd | head -1
00000000: 5b0a 0a5d                                [..]
```

**2. Rate limiting is reported as a missing version.**

Without `GITHUB_TOKEN`, `api.github.com` allows 60 requests/hour. Once exhausted, every page returns `{"message":"API rate limit exceeded ..."}`, the regex matches nothing, and the hook fails with an error that blames the version:

```text
ERROR: could not find a 'tflint' release asset matching version '0.64.0' (looked through up to 20 page(s) of releases).
```

The version is fine. `v0.64.0` is on page 1 and the regex matches it when the quota is available:

```bash
$ curl -s "https://api.github.com/repos/terraform-linters/tflint/releases?per_page=100&page=1" \
    | grep -oEi 'https://.+?/v0.64.0/tflint_darwin_arm64.zip'
https://github.com/terraform-linters/tflint/releases/download/v0.64.0/tflint_darwin_arm64.zip
```

The two bugs compound: because of (1), each failed attempt burns another 20 requests against the same 60/hour quota, so retrying makes recovery strictly less likely. The `20 page(s)` in the message is itself a tell - `tflint` only has ~3 pages of releases, so a working empty-page check would have stopped at page 4.

**Expected behavior:** fail fast with an error that identifies rate limiting and points at `GITHUB_TOKEN` (already supported by the script), without spending 20 extra API requests.

**Suggested fix** - `tools/install/_common.sh`, inside `common::install_from_gh_release` (the `while` loop, ~lines 80-95 on `master`; still present in `v1.109.1`):

- Make the empty-page check whitespace-insensitive (or parse the JSON) so `break` actually works.
- Detect a non-array / rate-limited response and abort with a distinct message instead of treating it as "version not found" and continuing to page. Checking for a top-level `"message"` key, or `curl -w '%{http_code}'` and bailing on 403/429, would both work.
- Mention `GITHUB_TOKEN` in the error output.

I'm happy to open a PR if the maintainers agree on the approach.

### How can we reproduce it?

1. Use the `.pre-commit-config.yaml` below (the key part is pinning `--hook-config=--tool-version=`).
2. Make sure `GITHUB_TOKEN` is unset and the unauthenticated quota for your IP is exhausted. This happens on its own behind a shared corporate egress IP; it can also be forced:

   ```bash
   unset GITHUB_TOKEN
   for i in $(seq 1 60); do curl -s -o /dev/null https://api.github.com/rate_limit; done
   curl -s https://api.github.com/rate_limit   # confirm .rate.remaining == 0
   ```

3. Clear the tool cache so the installer actually runs:

   ```bash
   rm -rf "${XDG_CACHE_HOME:-$HOME/.cache}/pre-commit-terraform/tflint"
   ```

4. Run the hook:

   ```bash
   pre-commit run terraform_tflint --all-files
   ```

<details><summary>Hook output</summary>

```bash
Downloading 'tflint' version '0.64.0'...
ERROR: could not find a 'tflint' release asset matching version '0.64.0' (looked through up to 20 page(s) of releases).
ERROR: Failed to download 'tflint' version '0.64.0' via '~/.cache/pre-commit/repo<hash>/hooks/../tools/install/tflint.sh'.
```

</details>

### Environment information

* OS: macOS 26 (Darwin 25.6.0), Apple Silicon (arm64)

* `uname -a` output:

```bash
Darwin 25.6.0 Darwin Kernel Version 25.6.0: Tue Aug 18 17:51:47 PDT 2026; root:xnu-12377.161.15.700.19~2/RELEASE_ARM64_T6050 arm64
```

```bash
GNU bash, version 5.3.20(1)-release (aarch64-apple-darwin25.6.0)
pre-commit 4.6.2
opentofu SKIPPED
Terraform v1.11.2
python SKIPPED
Python 3.14.7
checkov SKIPPED
infracost SKIPPED
terraform-docs version v0.24.0 darwin/arm64
terragrunt version 1.1.6
terrascan SKIPPED
TFLint version 0.64.0
tfsec SKIPPED
trivy Version: 0.74.0
tfupdate SKIPPED
hcledit SKIPPED
```

Note: `tflint 0.64.0` only shows up because I worked around the bug by seeding the cache manually; it was absent when the failure occurred.

* `.pre-commit-config.yaml`:

```yaml
repos:
    rev: v1.109.0
    hooks:
      - id: terraform_fmt
        args:
          - --args=-no-color
          - --args=-recursive
          - --args=-write=true
      - id: terraform_validate
        args:
          - --tf-init-args=-backend=false
        exclude: '^(?!.*_template_).*\.tf$'
      - id: terraform_tflint
        args:
          - --hook-config=--tool-version=0.64.0
          - --args=--module
          - --args=--config=__GIT_WORKING_DIR__/.tflint.hcl
          - --args=--fix
      - id: terraform_trivy
        args:
          - --args=--severity=CRITICAL,HIGH
          - --args=--skip-dirs="**/.terraform"
      - id: terraform_docs
        args:
          - --args=--sort-by-required
      - id: terraform_providers_lock
```

</details>

JH-lawliet:

- [x] This PR fixes a bug.

#### What

- Add `common::gh_api_get` to `tools/install/_common.sh`: it reads the HTTP status alongside the body and fails fast instead of feeding error payloads into the release-page loop — HTTP 429, and HTTP 403 whose body confirms rate limiting, get a rate-limit error that points at `GITHUB_TOKEN`; any other HTTP 403 is reported as a generic access denied failure; remaining non-200 responses and curl failures become explicit HTTP/transport errors.
- Make the pagination empty-page guard match GitHub's pretty-printed `[\n\n]` — the byte-exact `[]` check never fired, so every lookup burned the full 20 pages.
- Replace the `${var//[[:space:]]/}` full-body whitespace strip (which stalls on multi-MB release pages) with an anchored empty-page regex.
- Fixes #1023.

#### Why
Rate-limited 403 responses were treated as release data: curl lacked -f,
so the paginated lookup burned up to 20 API calls and reported a misleading
not-found error. This change handles the GitHub Rate-limited error scenario: fail fast with an error pointing at GITHUB_TOKEN instead of burning the remaining quota and throwing a misleading not-found error.

### How can we test changes

- `export GITHUB_TOKEN="$(gh auth token)" && uvx tox run -e pytest` → `68 passed`
- Repository pre-commit hooks（shfmt / shellcheck 等）on `tools/install/_common.sh` → `Passed`
- Local harness for `common::gh_api_get` (5/5 passed): 403 + rate-limit body → rate-limit diagnostic; 403 + permission body → generic `HTTP 403 (access denied)` with no rate-limit wording; 429 → rate-limit diagnostic; 200 → body passthrough; 404 → generic HTTP error.
- Live GitHub API checks of the new code path: normal release-asset lookup; pretty-printed empty page stops pagination; rate-limited 403 aborts immediately with the `GITHUB_TOKEN` hint (previously: up to 20 wasted requests + misleading "could not find release asset"); asset download succeeds.
- Fork CI `∞` run on this head (`35851278710`) → all jobs green: build, zizmor, pre-commit lint, metadata-validation, 20-job pytest matrix, aggregate `check`.

### Assisted-by

- Assisted-by: opencode:mimo-v2.6-flash-free opencode (784ea86, f63f3c9)


tools/install/_common.sh
```diff
@@ -26,6 +26,55 @@ if [[ $VERSION == false ]]; then
...
+  if ! response=$("${CURL_CMD[@]}" -sS -L -w $'\n%{http_code}' "$url"); then
+    echo "ERROR: failed to contact GitHub API at '$url'." >&2
+    exit 1
+  fi
+
+  http_code=${response##*$'\n'}
+  body=${response%$'\n'*}
+
+  if [[ $http_code == 429 ]] ||
+    { [[ $http_code == 403 ]] && grep -Fqi 'rate limit' <<< "$body"; }; then
```
yermulnik: Numbers should be compared using `-eq` in this case rather than as strings (`==`) please.

tools/install/_common.sh
```diff
@@ -26,6 +26,55 @@ if [[ $VERSION == false ]]; then
...
+#   Response body on stdout (only when HTTP 200)
+# Errors:
+#   Diagnostic on stderr; exits 1 on failure
+#######################################################################
+function common::gh_api_get {
+  local -r url=$1
+  local response http_code body
+
+  if ! response=$("${CURL_CMD[@]}" -sS -L -w $'\n%{http_code}' "$url"); then
+    echo "ERROR: failed to contact GitHub API at '$url'." >&2
```
yermulnik: I guess we use `common::colorify` function to throw messages and errors. Please follow this pattern.
Errors are `red`, warnings are `yellow`, notices and regular messages are `green` if I'm not mistaken.

tools/install/_common.sh
```diff
@@ -26,6 +26,55 @@ if [[ $VERSION == false ]]; then
...
+    exit 1
+  fi
+
+  http_code=${response##*$'\n'}
+  body=${response%$'\n'*}
+
+  if [[ $http_code == 429 ]] ||
+    { [[ $http_code == 403 ]] && grep -Fqi 'rate limit' <<< "$body"; }; then
+    echo "ERROR: GitHub API rate limit exceeded while querying '$TOOL' releases (HTTP $http_code)." >&2
+    echo "Set GITHUB_TOKEN to authenticate (already supported by this script) or retry later. See https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting" >&2
```
yermulnik: I think the `(already supported by this script)` bit is extraneous/redundant:
```suggestion
    echo "Pass your GitHub access token by means of exporting \"GITHUB_TOKEN\" environment variable to send authenticated calls or retry later. See https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting" >&2
```

The suggested message might seem to be a bit too long. I'm open for suggestions to shrink it while keeping it informative. Thanks.

And also please use `common::colorify` to spit out messages. The `ERROR` message should be `red`, while the supporting message should probably be `yellow` 🤔

tools/install/_common.sh
```diff
@@ -26,6 +26,55 @@ if [[ $VERSION == false ]]; then
...
+    exit 1
+  fi
+
+  if [[ $http_code != 200 ]]; then
+    if [[ $http_code == 403 ]]; then
+      echo "ERROR: GitHub API request to '$url' failed with HTTP 403 (access denied)." >&2
+    else
+      echo "ERROR: GitHub API request to '$url' failed with HTTP $http_code." >&2
```
yermulnik: * Please use `-ne` and `-eq` to compare numbers instead of `!=` and `==` respectively.
* Please use `common::colorify` to print out messages.
* Please use `$http_code` instead of hardcoding `403` in the message.
  * Also please replace `access denied` with `Forbidden` as it's what HTTP `403` is.

tools/install/_common.sh
```diff
@@ -71,17 +120,30 @@ function common::install_from_gh_release {
...
 
+  local asset_url="" latest_releases page_releases
+
   if [[ $VERSION == latest ]]; then
-    "${CURL_CMD[@]}" -L "$("${CURL_CMD[@]}" -s "${RELEASES}/latest" | grep -o -E -i -m 1 "$GH_RELEASE_REGEX_LATEST")" > "$PKG"
+    latest_releases=$(common::gh_api_get "${RELEASES}/latest")
+    asset_url=$(grep -o -E -i -m 1 "$GH_RELEASE_REGEX_LATEST" <<< "$latest_releases" || true)
+
+    if [[ -z $asset_url ]]; then
+      echo "ERROR: could not find a release asset for '$TOOL' in its latest release (regex '$GH_RELEASE_REGEX_LATEST' matched nothing)." >&2
```
yermulnik: * `common:colorify` please
* I'd also suggest to reword to:
  ```
  ERROR: Failed to find '$TOOL' latest release asset matching the '$GH_RELEASE_REGEX_LATEST' regex.
  ```

tools/install/_common.sh
```diff
@@ -71,17 +120,30 @@ function common::install_from_gh_release {
...
+    # anchored regex: ${var//[[:space:]]/} pattern substitution is
+    # pathologically slow on multi-MB pretty-printed API bodies.
+    local -r empty_page_re='^[[:space:]]*\[[[:space:]]*\][[:space:]]*$'
     while [[ -z $asset_url && $page -le $max_pages ]]; do
-      page_releases=$("${CURL_CMD[@]}" -s "${RELEASES}?per_page=100&page=${page}")
-      [[ $page_releases == "[]" ]] && break
+      page_releases=$(common::gh_api_get "${RELEASES}?per_page=100&page=${page}")
+      # GitHub may pretty-print an empty array as "[\n\n]" (4 bytes),
+      # not "[]" - allow whitespace inside and around the brackets.
+      [[ $page_releases =~ $empty_page_re ]] && break
```
yermulnik: Ha, that's a great finding about pattern substitution vs regex matching! 👍🏻 
Could we probably just get rid of yet another var and put the regex right into the test condition to reduce complexity reading the code?
I mean to use `[[ $page_releases =~ ^[[:space:]]*\[[[:space:]]*\][[:space:]]*$ ]] && break`

I'm okay with using var as you did if there's a reason for that though.
Thanks.

tools/install/_common.sh
```diff
@@ -90,8 +152,11 @@ function common::install_from_gh_release {
       echo "ERROR: could not find a '$TOOL' release asset matching version '$VERSION' (looked through up to $((page - 1)) page(s) of releases)." >&2
       exit 1
     fi
+  fi
 
-    "${CURL_CMD[@]}" -L "$asset_url" > "$PKG"
+  if ! "${CURL_CMD[@]}" -sS -f -L "$asset_url" > "$PKG"; then
+    echo "ERROR: failed to download '$TOOL' release asset from '$asset_url'." >&2
```
yermulnik: * `common::colorify` please
* also please capitalize the word `failed` for the consistency's sake

yermulnik (commented):
LGTM
Appreciate the contribution.
Please consider my comments below.

tools/install/_common.sh
yermulnik: Also in Bash this expression technically can be written as below:
```shell
  if [[ $http_code -eq 429 || \
    ( $http_code -eq 403 && $(tr '[:upper:]' '[:lower:]' <<< "$body") =~ "rate limit" ) ]]; then
```

What I mean is that you can use grouping inside `[[ … ]]` expression.
Doesn't change the result, but avoids using command pipelining in favour of combining expressions using operators and uses regex matching vs `grep` (although uses `tr` a`$(…)` and `tr` is only need because of compatibility with Bash v3 🤪).

tools/install/_common.sh
JH-lawliet: @yermulnik  Thanks for all the detailed review and suggestions, I will check and fix them as soon as I can.

JH-lawliet:
Hi @yermulnik, I have refactored the code according to the comments, could you review again? 🙏

tools/install/_common.sh
```diff
@@ -26,6 +26,88 @@ if [[ $VERSION == false ]]; then
...
+# Colorize provided string and print it out to stdout
+# Environment variables:
+#   PRE_COMMIT_COLOR (string) If set to `never` - do not colorize output
+# Arguments:
+#   COLOR (string) Color name that will be used to colorize
+#   TEXT (string)
+# Outputs:
+#   Print out provided text to stdout
+#######################################################################
+function common::colorify {
```
yermulnik: Oh my, I guess I mixed things all the way up 🤦🏻 I took it as the PR to update `/hooks/_common.sh` while it's to improve `/tools/install/_common.sh`. I really beg your pardon as I confused it all 🤦🏻 The `/tools/install/` need no color output as it's not an interactive human facing output I guess. I really do apologize — could you please revert this to plain `echo`?

tools/install/_common.sh
```diff
@@ -71,17 +153,27 @@ function common::install_from_gh_release {
 
   local -r CURL_CMD=("curl" "${CURL_OPTS[@]}")
 
+  local asset_url="" latest_releases page_releases
+
   if [[ $VERSION == latest ]]; then
-    "${CURL_CMD[@]}" -L "$("${CURL_CMD[@]}" -s "${RELEASES}/latest" | grep -o -E -i -m 1 "$GH_RELEASE_REGEX_LATEST")" > "$PKG"
+    latest_releases=$(common::gh_api_get "${RELEASES}/latest")
+    asset_url=$(grep -o -E -i -m 1 "$GH_RELEASE_REGEX_LATEST" <<< "$latest_releases" || true)
```
yermulnik: To my personal preference the `|| true` expression which is, I guess, meant to prevent failure when `set -e` is in actions, should better be moved outside parenthesis to denote it's not meant for var value assignment but to make the script not fail:
```suggestion
    asset_url=$(grep -o -E -i -m 1 "$GH_RELEASE_REGEX_LATEST" <<< "$latest_releases") || true
```

tools/install/_common.sh
```diff
@@ -71,17 +153,27 @@ function common::install_from_gh_release {
...
+
+    if [[ -z $asset_url ]]; then
```
yermulnik:
```suggestion
    if [[ ! $asset_url ]]; then
```

yermulnik (commented):
Appreciate your time @JH-lawliet and apologies for the confusion I introduced in previous round of review.
The feature LGTM though.
@MaxymVlasov Would you please join the review when you have a chance? Thanks.

